PR #214, opened 2013-05-12
Fixed py.test coverage.

hynekcer:
Import refactored so that "import jedi.setting" and "import test.conftest"
imports currently only a small number of short files. Previously it imported
almost all files in the package. Some names use functional proxy for lazy import.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,9 @@
 [run]
 omit =
+    jedi/__init__.py
     jedi/_compatibility.py
+    jedi/errors.py
+    jedi/settings.py
 
 [report]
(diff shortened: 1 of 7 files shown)


vlcinsky:
I had very similar problem, in my case it did not import anything for analysing the coverage despite the fact, py.test run the tests using the imported code.

Quick fix was:

    $ export PYTHONPATH="."

This often helped me in the past, when pytest had problem to import something.

After this change, my coverage jumped from 0% to 90%.

davidhalter:
What I usually do to solve all issues regarding installed packages is: `pip install -e .`

Issue #295, opened 2013-08-18
REPL completion just shows dots


asmeurer:
See https://github.com/davidhalter/jedi/pull/268#issuecomment-22817522 for more information. The REPL completer shows completions like

``` py
>>> ImportError if foo else Byte
...sWarning  ...array     ...s
>>> ImportError if foo else Byte
```

This is because the default would be to show

``` py
>>> ImportError if foo else Byte
ImportError if foo else BytesWarning  ImportError if foo else bytearray     ImportError if foo else bytes
>>> ImportError if foo else Byte
```

But what would be best would be

``` py
>>> ImportError if foo else Byte
BytesWarning  bytearray     bytes
>>> ImportError if foo else Byte
```

I'll take a shot at fixing this. You just need to write a hook for use by `readline.set_completion_display_matches_hook`.  Unfortunately, I don't see a lot of reusable logic from the readline module, so I may have to write it from scratch.

PR #296, opened 2013-08-19
Work on making the REPL completion show names instead of dots

asmeurer:
This is NOT ready to be merged. I am just opening a pull request so it is easier to see the code.  See https://github.com/davidhalter/jedi/issues/295.

So far:
- [ ] This will need to use the `rl` library. I believe that provides sufficient functions to make this really easy.
- [ ] I am using a regular expression to get the "completed" part out of the input, but surely Jedi provides an API for this. 

I think I know how to do the first point (if I'm not mistaken, http://pythonhosted.org/rl/readline.html#rl.readline.get_completion_display_matches_hook should help), but I'm not sure about the second. Any suggestions there would help.

### DIFF
--- a/jedi/utils.py
+++ b/jedi/utils.py
@@ -128,22 +128,19 @@ def setup_readline(namespace_module=__main__):
         # re.UNICODE is technically not correct in Python 2, but it shouldn't hurt
         identifier = re.compile(r"[^\d\W]\w*$", re.UNICODE)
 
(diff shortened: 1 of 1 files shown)


davidhalter:
Since no work has been done in the rl library I'm closing this one. I don't think this is going anywhere. If anyone has improvements I'm happy to get new pull requests.

PR #321, opened 2013-09-28
Fix the Jedi REPL completion to work with IPython

asmeurer:
Previously, it wasn't really working (it just looked like it did because it
set the completer to use `...`). The reason is that IPython resets the completer
to its own completer after every prompt. The solution is to detect when we are
in IPython and disable its readline support.

### DIFF
--- a/jedi/utils.py
+++ b/jedi/utils.py
@@ -40,7 +40,7 @@ class JediRL():
         except IndexError:
             return None
 
-def setup_readline(namespace_module=__main__):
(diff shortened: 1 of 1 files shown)


davidhalter:
Wait long enough and people solve your problems. Jedi is now part of IPython (the big new feature of IPython 6): http://ipython.readthedocs.io/en/stable/whatsnew/version6.html#new-completion-api-and-interface

The only thing that might need updating now is the Jedi documentation itself.

dbrgn:
Wow, congrats :) :sparkles:

PR #484, opened 2014-09-14
fix tokenization of code containing unicode strings

hatamov:
Code containing unicode strings (strarting with: u""", u''' and u") tokenizes incorrect.


davidhalter:
ping?

hatamov:
pong

davidhalter:
Thanks so much for the quick reaction :-)

PR #505, opened 2014-11-11
remove compatibility for hasattr && autopep8

dongweiming:
`hasattr` and `next` has in > python2.6.


davidhalter:
I have done it by my own hands now, because I don't agree with the autopep8 stuff (also there's a merge conflict now. Thanks anyway for bringing this up!

Issue #526, opened 2015-01-15
Issue when filtering completion candidates with East-Asian characters


KenetJervet:
I thought it was a company-mode bug so I created the issue there:
company-mode/company-mode#277

I played a bit with Jedi API, and got the following result:

This is correct:
![image](https://cloud.githubusercontent.com/assets/1560189/5757560/853449e6-9cf7-11e4-804f-a5c084ce7731.png)

This is wrong:
![image](https://cloud.githubusercontent.com/assets/1560189/5757573/a5463190-9cf7-11e4-88db-7b36f9a1315d.png)

The issue reference to company-mode has more detailed description.

Thanks!

davidhalter:
Interesting! Can you post the code as pure text?

KenetJervet:
@davidhalter: Will do it in a minute.
At the same time, I just looked into the code, and it seems the issue stems from `namechars` defined in tokenize.py
it assumes a word that starts with anything in string.ascii_letters are names, but apparently it does not include any Unicode characters.

KenetJervet:
Here's the code:
```Python
import jedi
code = '''
def func_我():
  pass
  
def func_你():
  pass
  
func_我
'''
script = jedi.api.Script(code, line=8, column=6)
script.completions()
```

davidhalter:
Hmmm since the tokenizer is a fork of the stdlib tokenizer, is that a bug as well in the standard library? (in `tokenize.py`)

KenetJervet:
hi. Unicode chars are only considered valid in Python 3. I briefly compared two versions of tokenize.py from stdlib, and I suppose you forked the one in Python 2.
Python 3 has changed the implementation a bit. So instead of
checking `initial in namechars`, it checks `initial.isidentifier()`.

I'd love to make a pull request but since I've never written any code that's 2/3 compatible I hesitate to do so.

KenetJervet:
Hi @davidhalter , I made a pull request. Now in my end it works well with company-mode and elpy. You may check if there's something that I missed.

davidhalter:
Fixed by #527.

PR #528, opened 2015-01-16
Fixed issue #526.

KenetJervet:
Hi. I rewrote the test for the new parser branch. I'm a bit confused about how to write the test for Python 2. tokenize.py from the standard library will parse the Unicode string into an ERRORTOKEN but your version returns an OP. It seems not a big deal for now.


davidhalter:
Looks good, thanks a lot!

> tokenize.py from the standard library will parse the Unicode string into an ERRORTOKEN but your version returns an OP. It seems not a big deal for now.

Yeah they are quite different, but most of the differences can be ignored, because of the huge test coverage.

PR #530, opened 2015-01-16
Updated Arch Linux installation

felipeacsi:
Hello. Thanks for your work on jedi, it's really helpful.

The Arch Linux packages are now in the [official repository](https://www.archlinux.org/packages/?sort=&q=jedi) and not in AUR anymore, so I've just updated the docs.


davidhalter:
Great, thanks!

PR #532, opened 2015-01-19
Add any .egg-link paths from VIRTUAL_ENV to sys.path

blueyed:
I've noticed that Jedi does not appear to handle packages that are installed using `pip install -e`, where an `.egg-link` is being used in the virtualenv, e.g.:

    $VIRTUAL_ENV/lib/python2.7/site-packages/Django.egg-link

This PR fixes this by adding the contents of all `.egg-link` files to `sys.path`.

There might be a better method, e.g. some helper from setuptools or something similar though?!

### DIFF
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -1,3 +1,4 @@
+import glob
 import os
 import sys
 
@@ -23,6 +24,10 @@ def get_sys_path():
                              'site-packages')
         if p not in sys_path:
             sys_path.insert(0, p)
+
+        # Add all egg-links from the virtualenv.
+        for egg_link in glob.glob(os.path.join(p, '*.egg-link')):
+            sys_path.insert(0, open(egg_link).readline().rstrip())
 
     check_virtual_env(sys.path)
     return [p for p in sys.path if p != ""]


davidhalter:
Nice!

Two things:

1. Would it be possible to test this? (In an easy way, you don't need to actually run `pip install...`, just modify the sys path)
2. I'm currently developing against the "parser" branch, could you PR it against that branch? Quite a few things have changed (in the sys path logic as well).
 
If you're not doing it, I will do it sometime in the future :)

blueyed:
1. Yes, tests would be nice of course. How about adding a `test_check_virtual_env` method in `test/test_evaluate/test_sys_path.py`, and then mocking `glob` and `open` from there? Should I use the `mock` library? I could not find anything about "mock" in the current source.

2. Sure. I've cherry-picked the commit into a new branch, and will push it as a new PR, because you cannot edit an existing PR with regard to where it should get merged into, can you?

davidhalter:
Yes. You need to create a new PR.

I would not mock anything if I were you. Just create a folder like `test/test_evaluate/egg-link` and put all the files that would be needed for a test into it and change the `sys.path`. You will probably also need to add an exception in `pytest.ini`.

blueyed:
Submitted a new PR: #534.

Issue #535, opened 2015-01-24
API to adjust jedi's internal sys.path


blueyed:
I am using jedi through YouCompleteMe in Vim, and would like to adjust / change its `sys.path` during runtime.

The specific use case is (de)activating a virtualenv from within Vim (I am using https://github.com/jmcantrell/vim-virtualenv).

davidhalter:
Fixed AFAIK by #636.

Issue #385, opened 2014-03-18
Discussion: Jedi wrapper module using server-client architecture to deal with multiprocessing, async interface, multiple Python version, cache lock, etc.


tkf:
From https://github.com/davidhalter/jedi/issues/384#issuecomment-37860688 (discussion about multithreading / multiprocessing) and https://github.com/davidhalter/jedi/issues/340#issuecomment-37878924 (discussion about asynchronous interface), I got this idea of creating a "wrapper" module of Jedi with server-client architecture which launches possibly multiple Jedi processes behind the scene, to solve problems such as:
- Multiple Python versions.  Jedi parses Python stdlib and import `*.so` so it is not possible to complete Python 3 modules while running Jedi in Python 2.  However, getting completions for both Python 2 and 3 can be useful when you are writing Python 2/3 compatible module.
- Async interface.  Due to the GIL, and intense computation required by Jedi, it is not feasible to have async interface within Jedi.
- Cache locking/sync:  Jedi does not lock cache files.  Also, when cache is updated in two processes, later one might override the former one.
- Cache update triggered by file modification:  When some modules are updated, Jedi has to wait until next access to update the cache.  If we can trigger cache update by file modification then Jedi can be faster.

If we have a centralized server, a pool of workers and clients to connect to the workers then we can solve the above problems nicely.  I think it is also consistent with a philosophy of keeping Jedi itself minimal and free of external modules.

Client part of this module can expose (blocking) API which is exactly the same as Jedi itself.  Then, every editor can just import this wrapper module instead of Jedi to use it, without much change to the editor plugin.  If the editor has an async support, then it can use async API of this client module.

Maybe my suggestion over complicates the problem, but I think it is worth discussing it anyway, since some problems are better _not_ solved in Jedi.  Delegating these problems to other module can be a good option to consider.

fjfnaranjo:
I will like to add a little information here about jedi-vim so it can be taken into account when anyone starts working on this.

I have been using this https://github.com/nakamuray/jedi-rpc.vim to fix all my problems related to Python+virtualenv version compatibility problems and package discovery in jedi-vim. All credit goes to @nakamuray , of course.

[...]

PR #539, opened 2015-01-26
HACK: append sys.path from other major python via subprocess

blueyed:
This makes working with Python 2 and 3 in parallel less painfull.

My use case is Jedi being used in Vim via YouCompleteMe, which only
supports Python 2, but I'd like to use the completion also when editing
Python 3 files.

What do you think? Is this a approach in the right direction?
See also #535 (API to adjust sys.path).

### DIFF
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -24,6 +24,18 @@ def get_sys_path():
         if p not in sys_path:
             sys_path.insert(0, p)
 
+    # HACK: add sys.path from python3 and vice versa.
+    import subprocess
+    other_python = 'python%d' % (22 if sys.version_info.major == 3 else 3,)
+    try:
+        other_sys_path = subprocess.check_output(
+            [other_python, '-c', 'import sys; print(sys.path)']).decode('utf-8')
+    except OSError:
+        pass
+    else:
+        other_sys_path = other_sys_path[2:-3].split("', '")
+        sys.path += other_sys_path
+
     check_virtual_env(sys.path)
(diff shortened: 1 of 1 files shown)


davidhalter:
Interesting, but this is really risky, because it can lead to segfaults. Jedi tries to import libraries like `PyQt4` and so on. Those binaries must never be executed with the wrong Python version.

blueyed:
> Those binaries must never be executed with the wrong Python version.

Is it possible to detect those somehow?
Are those in 'lib/python3.4/lib-dynload' usually - then that path could be skipped.

davidhalter:
I mean you could just create "ignore paths" for binaries. That would fix the issue.

But at the same time, I don't like the approach in general too much. I would rather have an implementation of https://github.com/davidhalter/jedi/issues/385, to make it easy to switch Python versions.

blueyed:
@davidhalter 
Has this been closed automatically because the "parser" branch was deleted, or intentionally (given your last comment(s))?

davidhalter:
automatic close. cannot even reopen it haha. Is it ok to keep it closed? IMHO this is not the way to go, because it's really just a workaround for a problem that I would like to solve different.

A patch I would accept would include something more like a fallback to **just** third party libraries, it would need a way to exclude all binary files. (or a different loading mechanism for binary files via multiprocessing).

This is definitely something that would take a lot of time to implement. If you're interested, go on. Otherwise you can just open a new issue if you're really interested.

blueyed:
> Is it ok to keep it closed? 

Yes.  Thanks for your additional information and feedback!

Issue #545, opened 2015-02-09
max recursion limit with new parser code and buildout project


mfussenegger:
Since the new parser code has been merged I wanted to give it a try. But it seems like the buildout support got broken in the re-write.

In ``sys_path.py`` the ``_check_module`` function is missing its second argument (in line 158) but after adding that it runs into the max recursion limit as seen in the following gist.

https://gist.github.com/mfussenegger/1a036789c29eeb620d32.

davidhalter:
Thanks! Well, probably buildout is not well tested then.

PR #546, opened 2015-02-10
run buildout detection only once per buildout script

mfussenegger:
in order to avoid running into the max recursion limit.

This solves the issue reported here https://github.com/davidhalter/jedi/issues/545

The test doesn't really trigger the recursion limit problem but should at least make sure that the buildout related methods in `sys_path_with_modifications` are called correctly.


davidhalter:
Thanks!

I don't like the new global variables. Whenever I did that I had to undo it later. Do you really need that to avoid recursions? If so, could you please post the recursion traceback (or at least the part that's repeating and the beginning).

mfussenegger:
I fully agree that the globals are not nice. 
Would probably also cause problems if the same instance is used for multiple problems.

I'll try to figure out a better solution once I can set aside some time.

mfussenegger:
Tried to dig a little bit deeper.

It seems that some buildout scripts cause an recursion loop. I tested it in the
[crate-python](https://github.com/crate/crate-python) project and there the bin
folder looks like this:

    ls -1 bin

    buildout
    coverage
    createcoverage
    py
    rst2html.py
    rst2latex.py
    rst2man.py
    rst2odt_prepstyles.py
    rst2odt.py
    rst2pseudoxml.py
    rst2s5.py
    rst2xetex.py
    rst2xml.py
    rstpep2html.py
    sphinx
    test
    tox
    wheel

What I've discovered is that every time the ``_check_module`` is called on the
``sphinx`` file it will somehow run into ``sys_path_with_modifications`` again
and start all over.

This is the [sphinx file](https://gist.github.com/mfussenegger/84b1f51c7874a583e144)

and [here is some debug
output](https://gist.github.com/mfussenegger/2353b3b69b1f58efd6f2) with the
stack trace printed every time ``_paths_from_assignment`` is called.

Any ideas?

davidhalter:
Hmm, it seems like `sys_path_with_modifications` is somehow callable as a recursion. I'm still not exactly sure where it comes from, but it probably makes sense.

I remember changing the default of `sys_path_with_modifications` memoization to a list. Could you retry now and see if it still fails.

mfussenegger:
Just tried it. The process ended with 100% cpu utilization - I let it run for a while to see if it would return eventually but it ended up eating away all memory (10G+) so I had to kill it.

Would it be okay to use the python ast module instead to parse the buildout scripts? Their syntax must be correct anyway.

davidhalter:
Hmm what why would that help?

davidhalter:
All right. I've had another look on this:

      @memoize_default(evaluator_is_first_arg=True, default=[])
      def sys_path_with_modifications(evaluator, module):

`sys_path_with_modifications` is memoized. Therefore if called with the same module, it should not be called multiple time (the second time it returns the default `[]`. Since the evaluator doesn't change, the changing part would be `module`. The question is: Why does `module` change. Modules are cached in Jedi and should not be loaded multiple times. So what I would need is either a way to reproduce the issue (source code) or a printout of the module values passed to `sys_path_with_modifications`.

mfussenegger:
Sorry for the late response.

You're right, using the stock python AST wouldn't really help as the recursion is triggered somewhere in the code that analyzes the result of the parser.

Since you're planning to make a 0.9 release soonish I'll try to take a further look at it today in the evening.

Otherwise if you wan't to reproduce the error yourself you should be able to if you e.g. checkout https://github.com/crate/crate-python and run the buildout (see https://github.com/crate/crate-python/blob/master/DEVELOP.rst) and then try to use jedi's goto definition or something like that somewhere in the sources.

davidhalter:
:+1:

mfussenegger:
What exactly do you mean by

> The question is: Why does module change. Modules are cached in Jedi and should not be loaded multiple times

in ``sys_path_with_modifications``
there is:

    p = Parser(evaluator.grammar,
                common.source_to_unicode(source),
                buildout_script)

so the buildout scripts are parsed multiple times.


I've looked at what is fed into memoize_default and detected the following:

    function: <function sys_path_with_modifications at 0x7f3728e481b8>
    obj: <jedi.evaluate.Evaluator object at 0x7f3728d26f10>
    memo: 
    { (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([])): [ ],
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/src/crate/crash/command.py@25-516>,), frozenset([])): [ ]}

    key NOT in memo: (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([]))

and after the next call it looks like this:

    function: <function sys_path_with_modifications at 0x7f3728e481b8>
    obj: <jedi.evaluate.Evaluator object at 0x7f3728d26f10>
    memo: 
    { (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([])): [ ],
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([])): [ ],
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/src/crate/crash/command.py@25-516>,), frozenset([])): [ ]}

    key NOT in memo: (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([]))

So it seems that the evaluator/module stuff ends up in the cache but for some reason the module's don't match:

    ipdb> pp key
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>,
    (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,),
    frozenset([]))

    ipdb> pp memo
    {(<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([])): [],
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/bin/sphinx@3-28>,), frozenset([])): [],
    (<jedi.evaluate.Evaluator object at 0x7f3728d26f10>, (<Module: /workspace/code/ct/crash/src/crate/crash/command.py@25-516>,), frozenset([])): []}

    ipdb> pp k2
    <Module: /workspace/code/ct/crash/bin/sphinx@3-28>

    ipdb> pp m2
    <Module: /workspace/code/ct/crash/bin/sphinx@3-28>

    ipdb> pp k2 == m2
    False

So before I continue to debug this: If the same source is parsed two times,
should the resulting ``p1.module`` be equal to ``p2.module`` or does this rely
on identity equality and the same buildout scripts should never be parsed two
times?

davidhalter:
The same buildout scripts should never be parsed two times. In general, no module should exist as a duplicate within Jedi.

mfussenegger:
Just added a fixup where I use memoize_default to cache the extra sys-paths that are retrieved from a buildout script. Seems to work.

Should I change it to use ``cache.load_parser(path, name)`` anyway? 

I've also rebased against dev so this PR looks now a bit odd because it was created against master.

davidhalter:
Hmm well, could you open a new PR just against dev?

Issue #466, opened 2014-08-17
SyntaxError: EOL while scanning string literal


jorgenschaefer:
Jedi seems to have problems with this piece of code:

```Python
lit = u"""\
# -*- coding: utf-8 -*-
"""
```

The difference to #465 is that the string now is a unicode string, not a byte string. The error is a different one.

```Python
>>> source = u'lit = u"""\\\n# -*- coding: utf-8 -*-\n"""\n'
>>> script = jedi.Script(column=3, source=source, line=2, encoding='utf-8')
>>> script.call_signatures()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/api/__init__.py", line 558, in call_signatures
    call, index = search_call_signatures(user_stmt, self._pos)
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/evaluate/helpers.py", line 123, in search_call_signatures
    arr, index = call_signature_array_for_pos(user_stmt, position)
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/evaluate/helpers.py", line 103, in call_signature_array_for_pos
    for command in stmt.expression_list():
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/cache.py", line 139, in wrapper
    result = func(self)
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/parser/representation.py", line 1183, in expression_list
    call = cls(self._sub_module, tok_str, tok.start_pos, tok.end_pos, self)
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/parser/representation.py", line 1339, in __init__
    self.value = literal_eval(literal)
  File "/home/forcer/.virtualenvs/elpy/local/lib/python2.7/site-packages/jedi/_compatibility.py", line 193, in literal_eval
    return ast.literal_eval(string)
  File "/usr/lib/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/usr/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 2
    # -*- coding: utf-8 -*-
                          ^
SyntaxError: EOL while scanning string literal
[...]

PR #552, opened 2015-03-06
Fix tokenization of code containing unicode strings.

hatamov:
Code containing unicode strings (strarting with: u""", u''' and u") tokenizes incorrect. 
This is remake of PR https://github.com/davidhalter/jedi/pull/484


davidhalter:
Exactly the right timing. I was just trying to fix #466 and then this PR comes along, awesome! :-)

However, you tested against a much older Jedi version (dev is quite a bit "newer"). But I'm going to fix that.

PR #556, opened 2015-03-18
Dynamic check for VIRTUAL_ENV python version

walkert:
Previously, an assumption was made that the virtualenv was created using
a python version that matched sys.version_info[:2]. If this was not the
case, an invalid path is added to sys.path and we don't get any
completion.

This commit checks the 'pythonX.X' file within $VIRTUAL_ENV/lib and sets
the path accordingly. The test_get_sys_path() test has been updated to
properly test this by mocking os.listdir().

### DIFF
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -33,8 +33,15 @@ def _get_venv_sitepackages(venv):
     if os.name == 'nt':
         p = os.path.join(venv, 'lib', 'site-packages')
     else:
-        p = os.path.join(venv, 'lib', 'python%d.%d' % sys.version_info[:2],
-                         'site-packages')
+        # Account for the VIRTUAL_ENV using an alternate python verison
+        venv_lib = os.path.join(venv, 'lib')
+        try:
+            pyver = [p for p in os.listdir(venv_lib) if
+                     p.startswith('python')][0]
+            p = os.path.join(venv_lib, pyver, 'site-packages')
+        except IndexError:
+            # There should always be a 'pythonX.X' file, but just in case
+            p = ''
     return p
 
 
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -1,5 +1,6 @@
 import glob
 import os
+import re
 import sys
 
 from jedi._compatibility import exec_function, unicode
@@ -30,6 +31,7 @@ def get_sys_path():
     return [p for p in sys.path if p != ""]
 
 def _get_venv_sitepackages(venv):
+    pyre = re.compile('^python\d\.\d$')
     if os.name == 'nt':
         p = os.path.join(venv, 'lib', 'site-packages')
     else:
@@ -37,7 +39,7 @@ def _get_venv_sitepackages(venv):
         venv_lib = os.path.join(venv, 'lib')
         try:
             pyver = [p for p in os.listdir(venv_lib) if
-                     p.startswith('python')][0]
+                     pyre.search(p)][0]
             p = os.path.join(venv_lib, pyver, 'site-packages')
         except IndexError:
             # There should always be a 'pythonX.X' file, but just in case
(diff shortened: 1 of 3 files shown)


walkert:
Alas the py27 tests were already failing before I submitted this..

davidhalter:
@walkert Thanks so much

@blueyed Didn't you add this virtualenv support? Can you confirm that you were actually wrong (or are there two different ways?)

walkert:
@davidhalter @blueyed Daniel's last commit adds support for egg-link resolution which is still valid. The issue is when you've created a virtualenv like this:

virtualenv -p /path/to/non-standard/python <some name>

If you install a pip install from here and try using jedi - it will assume the python version matches that of vim's built-in python.

jedi/evaluate/sys_path.py
```diff
@@ -33,8 +33,15 @@ def _get_venv_sitepackages(venv):
     if os.name == 'nt':
         p = os.path.join(venv, 'lib', 'site-packages')
     else:
-        p = os.path.join(venv, 'lib', 'python%d.%d' % sys.version_info[:2],
-                         'site-packages')
+        # Account for the VIRTUAL_ENV using an alternate python verison
+        venv_lib = os.path.join(venv, 'lib')
+        try:
+            pyver = [p for p in os.listdir(venv_lib) if
```
blueyed: This might (theoretically) match multiple paths.

walkert: Theoretically yes. I believe however that having multiple pythonX.X files in the lib directory would be a non-standard setup (certainly when calling virtualenv with or without '-p' you should get a single-entry list returned.

If you think this is a valid corner case to be accounted for then check_virtual_env() could be modified to deal with multiple results from _get_venv_sitepackages().

blueyed: I've meant to say that it should make sure to only return a single entry, probably not by using `startswith` (below, IIRC) / matching more strictly.

jedi/evaluate/sys_path.py
```diff
@@ -30,11 +31,19 @@ def check_virtual_env(sys_path):
     return [p for p in sys.path if p != ""]
 
 def _get_venv_sitepackages(venv):
+    pyre = re.compile('^python\d\.\d$')
```
walkert: Fair enough. I'd originally avoided an re for speed but since this should be memoized there shouldn't be much of a hit.

davidhalter:
Hmm, do I understand correctly that this patch will allow loading any virtualenv? Regardless of Python version? I wasn't sure what the idea was before.

walkert:
Yes indeed. The aim is to make it version agnostic.

davidhalter:
@walkert That is something that I cannot accept in the current state of Jedi. If you're doing this without proper checks, you're actually making it easy for segfaults. Compiled (binary) files in a virtualenv can easily cause one - even with normal packages like `PyQt4`.

The same reason caused me to reject #539. It's not like I wouldn't want a version agnostic Jedi, but helping segfaults to work is definitely not what I want either.

I guess we have to move the whole import/compiled logic of Jedi into a separate `multiprocessing` process. That would solve our issues and we could easily just restart it if it segfaults. By doing that, we could maybe also include @blueyed's plan to include other library versions. This would also make Jedi more secure in general. There wouldn't be a reason anymore for Jedi to stop working (which it could do now with home made invalid compiled files that cause segfaults).

walkert:
@davidhalter I'm a little unclear on how this makes the situation worse though that is likely my understanding of what Jedi is doing.

Assuming I modified my patch to check that the pythonX.X value returned is a valid directory, I don't see how appending $VIRTUAL_ENV/<python_version>/lib/site-packages to sys.path makes it any more possible for compiled files to segfault than if a user had them in the standard path that you currently expect?

davidhalter:
Well it has nothing to do with virtualenvs. It's just about not importing files from other Python versions. Jedi doesn't do that. Jedi on Python 2.7 scans only for 2.7 directories, because if you scan for 3.x as well, you end up importing PyQt4 (compiled for 3.x), which would cause the Python compiler to segfault (the last time I've tried).

walkert:
Fair enough. To be honest I care little for python3.X compatibility. My situation involves the standard install being 2.6 and all of my venvs are on 2.7. I'd happily restrict it further but if safety dictates you can only consider sticking with the version compiled into vim then so be it.

davidhalter:
> My situation involves the standard install being 2.6 and all of my venvs are on 2.7.

AFAIK that can still cause segfaults, because some C headers are different. But I'm not sure. I will definitely keep an eye on this PR, because once we have a multiprocessing solution of the import logic, we can just use this PR to allow venvs with other python versions.

davidhalter:
Moved the discussion into #962.

PR #562, opened 2015-04-05
Improve virtualenv support & egg-link resolution

### LINKED ISSUE #384: Suggestion: keep sys.path per Evaluator instance
Currently Jedi uses sys.path directly and so editors have to modify sys.path in order to tell where modules locate.  This is not ideal.  For example, users may want to add Python 3 modules to sys.path while running Jedi in Python 2 (e.g., to write Python 2/3 compatible module).  To do that, I suggest to keep additional paths in a place other than sys.path.  Not only that, we can do more.  If we keep sys.path per Evaluator instance, then we can ask Jedi "what kind of modules I'd have if I have this set of `sys.path`?" without running Jedi in another Python process.  This may be useful, for example if you want to have different set of completions when you switch project.

(BTW, I am using partially 8 months-old knowledge of Jedi so, sorry if I miss some new features.)

### LINKED ISSUE #443: .pth status
* pth in user or global site-packages seem to work OK. Therefore, as I can see, they are not covered by tests (and not obvious that they are actually required).
* pth inside virtualenv's site-packages is broken. And tests are required.
* #345 solution seems to be broken in two points. First, only pth files from site-packages dirs should be included, but in that solution pth files from all sys.path entries are included. Second, in original site.py code there is probably (didn't check, `makepath` func) a workaround for windows case-insensitive paths (?).
* I have some doubt about how it exactly should be implemented.
  * add pth parsing directly in sys.path? Looks hacky, but as soon as venv's site-packages is dealed in sam way, it should be ok.
  * I didn't investigate what is the correct place for these paths. Probably, after venv's site-packages dir.

Will fix it if have a time. Otherwise, hope my comments will help to fix it in future.

immerrr:
AFAIU, it is generally considered ok to mix jedi's own sys.path and virtualenv paths, but recently I have run into a conflict between setuptools/distribute found in:
- jedi's virtualenv
- user site directory (`$HOME/.local/lib/pythonX.X/site-packages`)
- virtualenv I was actually using

This conflict caused problems in resolving something as simple as `setuptools.setup` and made me rethink module resolution towards "purity". 

This PR adds the following:
- add sys_path= kwarg to Script & Evaluator constructors
- store sys_path for each evaluator instance (closes #384)
- add get_venv_path(venv) function to determine virtualenv's path:
  - by default, dump path from live python interpreter
  - use old path extension variant for fallback
  - in old variant, use addsitedir to load .pth extension files (closes #443)
- look for egg-link files in all directories in path

Several discussion questions here:
- starting python interpreter from the virtualenv is the most precise way to know its sys.path, but are there drawbacks besides process startup overhead? e.g. should we clean `PYTHONSTARTUP` env var?
- if running python interpreter is ok, how do we test it? 
- as path resolution grows more and more heavy (.pth, egg-links, now subprocesses), do we need to cache it somehow? and for a long-running jedi process, like ycmd, how do we flush the cache to reflect package installation/removal.
- is there a use case for removing `site.USER_SITE` from lookups, too?

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -34,6 +34,7 @@ from jedi.evaluate.cache import memoize_default
 from jedi.evaluate.helpers import FakeName, get_module_names
 from jedi.evaluate.finder import global_names_dict_generator, filter_definition_names
 from jedi.evaluate import analysis
+from jedi.evaluate.sys_path import get_venv_path
 
 # Jedi uses lots and lots of recursion. By setting this a little bit higher, we
 # can remove some "maximum recursion depth" errors.
@@ -75,7 +76,8 @@ class Script(object):
     :type encoding: str
     """
     def __init__(self, source=None, line=None, column=None, path=None,
-                 encoding='utf-8', source_path=None, source_encoding=None):
+                 encoding='utf-8', source_path=None, source_encoding=None,
+                 sys_path=None):
         if source_path is not None:
             warnings.warn("Use path instead of source_path.", DeprecationWarning)
             path = source_path
@@ -109,7 +111,11 @@ class Script(object):
         self._parser = UserContextParser(self._grammar, self.source, path,
                                          self._pos, self._user_context,
                                          self._parsed_callback)
-        self._evaluator = Evaluator(self._grammar)
+        if sys_path is None:
+            venv = os.getenv('VIRTUAL_ENV')
+            if venv:
+                sys_path = list(get_venv_path(venv)) + sys.path
+        self._evaluator = Evaluator(self._grammar, sys_path=sys_path)
         debug.speed('init')
 
     def _parsed_callback(self, parser):
--- a/jedi/evaluate/compiled/__init__.py
+++ b/jedi/evaluate/compiled/__init__.py
@@ -10,7 +10,6 @@ from functools import partial
 from jedi._compatibility import builtins as _builtins, unicode
 from jedi import debug
 from jedi.cache import underscore_memoization, memoize_method
-from jedi.evaluate.sys_path import get_sys_path
 from jedi.parser.tree import Param, Base, Operator, zero_position_modifier
 from jedi.evaluate.helpers import FakeName
 from . import fake
@@ -309,15 +308,12 @@ class CompiledName(FakeName):
         pass  # Just ignore this, FakeName tries to overwrite the parent attribute.
 
 
-def dotted_from_fs_path(fs_path, sys_path=None):
+def dotted_from_fs_path(fs_path, sys_path):
     """
     Changes `/usr/lib/python3.4/email/utils.py` to `email.utils`.  I.e.
     compares the path with sys.path and then returns the dotted_path. If the
     path is not in the sys.path, just returns None.
     """
-    if sys_path is None:
-        sys_path = get_sys_path()
-
     if os.path.basename(fs_path).startswith('__init__.'):
         # We are calculating the path. __init__ files are not interesting.
         fs_path = os.path.dirname(fs_path)
@@ -341,13 +337,13 @@ def dotted_from_fs_path(fs_path, sys_path=None):
     return _path_re.sub('', fs_path[len(path):].lstrip(os.path.sep)).replace(os.path.sep, '.')
 
 
-def load_module(path=None, name=None):
+def load_module(evaluator, path=None, name=None):
+    sys_path = evaluator.sys_path
     if path is not None:
-        dotted_path = dotted_from_fs_path(path)
+        dotted_path = dotted_from_fs_path(path, sys_path=sys_path)
     else:
         dotted_path = name
 
-    sys_path = get_sys_path()
     if dotted_path is None:
         p, _, dotted_path = path.partition(os.path.sep)
         sys_path.insert(0, p)
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -71,7 +71,7 @@ def _get_venv_path_online(venv):
             break
     else:
         raise RuntimeError("Cannot find python executable in venv: %s" % venv)
-    command = [python_path, '-c', 'import sys; print(sys.path)']
+    command = [python_path, '-E', '-c', 'import sys; print(sys.path)']
     return literal_eval(check_output(command))
 
 
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -1,6 +1,9 @@
 import glob
 import os
 import sys
+from subprocess import check_output
+from ast import literal_eval
+from site import addsitedir
 
 from jedi._compatibility import exec_function, unicode
 from jedi.parser import tree
@@ -11,24 +14,66 @@ from jedi import common
 from jedi import cache
 
 
-def get_sys_path():
-    def check_virtual_env(sys_path):
-        """ Add virtualenv's site-packages to the `sys.path`."""
-        venv = os.getenv('VIRTUAL_ENV')
-        if not venv:
-            return
-        venv = os.path.abspath(venv)
-        p = _get_venv_sitepackages(venv)
-        if p not in sys_path:
-            sys_path.insert(0, p)
+def get_venv_path(venv):
+    """Get sys.path for specified virtual environment."""
+    try:
+        sys_path = _get_venv_path_online(venv)
+    except Exception as e:
+        debug.warning("Error when getting venv path: %s" % e)
+        sys_path = _get_venv_path_offline(venv)
+    with common.ignored(ValueError):
+        sys_path.remove('')
+    return _get_sys_path_with_egglinks(sys_path)
 
-        # Add all egg-links from the virtualenv.
+
+def _get_sys_path_with_egglinks(sys_path):
+    """Find all paths including those referenced by egg-links.
+
+    Egg-link-referenced directories are inserted into path immediately after
+    the directory on which their links were found.  Such directories are not
+    taken into consideration by normal import mechanism, but they are traversed
+    when doing pkg_resources.require.
+    """
+    result = []
+    for p in sys_path:
+        result.append(p)
         for egg_link in glob.glob(os.path.join(p, '*.egg-link')):
             with open(egg_link) as fd:
-                sys_path.insert(0, fd.readline().rstrip())
+                for line in fd:
+                    line = line.strip()
+                    if line:
+                        result.append(os.path.join(p, line))
+                        # pkg_resources package only interprets the first
+                        # non-empty line in egg-link files.
+                        break
+    return result
 
-    check_virtual_env(sys.path)
-    return [p for p in sys.path if p != ""]
+
+def _get_venv_path_offline(venv):
+    """Get sys.path for venv without starting up the interpreter."""
+    venv = os.path.abspath(venv)
+    sitedir = _get_venv_sitepackages(venv)
+    sys.path, old_sys_path = [], sys.path
+    try:
+        addsitedir(sitedir)
+        return sys.path
+    finally:
+        sys.path = old_sys_path
+
+
+def _get_venv_path_online(venv):
+    """Get sys.path for venv by running its python interpreter."""
+    venv = os.path.abspath(os.path.expanduser(venv))
+    for python_binary in ('python', 'python3', 'python.exe',
+                          'python3.exe'):
+        python_path = os.path.join(venv, 'bin', python_binary)
+        if os.path.isfile(python_path):
+            break
+    else:
+        raise RuntimeError("Cannot find python executable in venv: %s" % venv)
+    command = [python_path, '-c', 'import sys; print(sys.path)']
+    return literal_eval(check_output(command))
+
 
 
 def _get_venv_sitepackages(venv):
@@ -109,7 +154,6 @@ def _paths_from_list_modifications(module_path, trailer1, trailer2):
     name = trailer1.children[1].value
     if name not in ['insert', 'append']:
         return []
-
     arg = trailer2.children[1]
     if name == 'insert' and len(arg.children) in (3, 4):  # Possible trailing comma.
         arg = arg.children[2]
@@ -117,6 +161,9 @@ def _paths_from_list_modifications(module_path, trailer1, trailer2):
 
 
 def _check_module(evaluator, module):
+    """
+    Detect sys.path modifications within module.
+    """
     def get_sys_path_powers(names):
         for name in names:
             power = name.parent.parent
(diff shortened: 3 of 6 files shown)


immerrr:
I have tried to use this with emacs and it turns out that storing sys_path per Script (or Evaluator) instances is not very useful as long as Script objects are not persistent. Otherwise one might as well just calculate the sys_path to store it in their IDE of choice and send it with the regular completion/goto-definition requests.

Another alternative would be to make Script objects stateful, enable changing source piecewise (e.g. `script.set_source(begin, end, text)` and run completion as `completions(line, column)`. But something tells me that avoiding state was one of the initial design decisions.

immerrr:
I have encountered #539 which uses the same "live interpreter" approach and see it is turned down because of non-cross-version-safe imports happening in jedi. Ping me if you need this venv path detection scheme removed.

davidhalter:
Wow! That's quite a lot! Thanks for all the work that went into this!

I really like the general direction your pushing this. Please give me some time to review it.

jedi/api/__init__.py
```diff
@@ -108,7 +110,11 @@ def __init__(self, source=None, line=None, column=None, path=None,
         self._user_context = UserContext(self.source, self._pos)
         self._parser = UserContextParser(self._grammar, self.source, path,
                                          self._pos, self._user_context)
-        self._evaluator = Evaluator(self._grammar)
+        if sys_path is None:
+            venv = os.getenv('VIRTUAL_ENV')
+            if venv:
+                sys_path = list(get_venv_path(venv)) + sys.path
```
davidhalter: So if VIRTUAL_ENV is set, why is it not enough to just take the current `sys.path`? I have never really understood this.

jedi/evaluate/compiled/__init__.py
```diff
@@ -332,13 +328,13 @@ def dotted_from_fs_path(fs_path, sys_path=None):
     return _path_re.sub('', fs_path[len(path):].lstrip(os.path.sep)).replace(os.path.sep, '.')
 
 
-def load_module(path, name):
+def load_module(path, name, evaluator):
+    sys_path = evaluator.sys_path
```
davidhalter: I'm actually worried now that this line might cause some issues. It should probably be a `sys_path_with_modifications` call, but that's not your problem (or at least not a problem of this PR).

I just realized this, but even if it causes issues, there's not a big implication for all normal use cases, so don't care about it too much. It's much more that I should remember this.

jedi/api/__init__.py
immerrr: The idea is to have jedi installed somewhere once per system (or per user) and to have it look into various virtual environments without installing jedi in all of these environments. In Emacs one of the easiest ways to accomplish this (without pulling virtualenv names through all the APIs) is to launch jedi process with this extra environment variable whenever I change the venv I'm working on.

davidhalter: But if you change the virtualenv variable, wouldn't the `sys.path` also change automatically?

immerrr: I don't think so. AFAIR, `sys.path` is generated once at the startup and the virtualenv magic really relies on looking up candidate directories relative to the python interpreter location. Here's an article that probably can explain it better than I can (especially, within the scope of a comment): http://mikeboers.com/blog/2014/05/23/where-does-the-sys-path-start

davidhalter:
I don't see how this solves #443. Can you explain?

> starting python interpreter from the virtualenv is the most precise way to know its sys.path, but are there drawbacks besides process startup overhead? e.g. should we clean PYTHONSTARTUP env var?

Good question, I think it's ok as it is. The overhead shouldn't be that crazy - let's at least try. If the overhead is too big, we could work with caches that are deleted over time and take `VIRTUAL_ENV` as a key. I would actually delete the `PYTHONSTARTUP` var, I think that would make sense.

> if running python interpreter is ok, how do we test it?

Well, good question. Virtualenv stuff hasn't been really tested anyways. I guess we should start testing it. Those tests might be a bit "slow", but as long as there's an easy switch (py.test -k or --ignore), that's not a big issue. You could setup a virtualenv with code or just fake it (with a few symlinks). Whatever you prefer.

> as path resolution grows more and more heavy (.pth, egg-links, now subprocesses), do we need to cache it somehow? and for a long-running jedi process, like ycmd, how do we flush the cache to reflect package installation/removal.

As written above, Jedi supports time caches. However, I stopped caching stuff without actually profiling it first. So if you think that this really takes quite a lot of time, try to profile it first. Get a good use case and then we can actually work on improving the speed there.

My intuition is that actually checking all of that is not as time consuming as you might think it is. The big parts are probably still parsing and evaluation of the tree, but I might be wrong.

> is there a use case for removing site.USER_SITE from lookups, too?

I don't see a reason for that.

jedi/api/__init__.py
davidhalter: Hmm I've read it, very good article! I think I finally understand why `site.py` exists and what it does.

If you start a new process with Jedi in an activated virtualenv, wouldn't we see the correct `sys.path`? Are you talking about a changing just the  `VIRTUAL_ENV` variable without restarting the process?

In general, have you read the idea of #385, to make virtualenv support easier?

immerrr:
> I don't see how this solves #443. Can you explain?

When using live interpreter, the interpreter's startup procedure will resolve .pth links as it happens for the regular Python scripts. For offline detection, [this section](https://github.com/davidhalter/jedi/pull/562/files#diff-69648a5e66596e92bc556ea82fe84748R57) is an approximation of what happens during startup.

> Well, good question. Virtualenv stuff hasn't been really tested anyways. I guess we should start testing it.

I'm not sure I understand jedi/travis testing quite well, is it OK if I ask you to address the testing infrastructure?

*re:drop PYTHONSTARTUP* 

ok, will do

*re:caching*

The problem with caching this stuff is that the cache will never know when I install or remove packages into the environment, so there has to be an API to reset it.

I was looking at this from emacs-jedi perspective and couldn't find a better solution than just calculating `sys.path` once and then pass it for every jedi RPC command. Basically, RPC-wise it would be nice to be able to create a named piece of configuration via API (e.g. `configs["virtualenv=MYVENV"] = get_venv_path('MYVENV')` to live with the server and refer to it in future API requests, however it feels more like a problem for Jedi's RPC servers, not for Jedi itself. I wonder if we could do this in ycmd API...

jedi/api/__init__.py
immerrr: Yes, if Jedi is started in a virtualenv, it will get a proper `sys.path`.

As of now, I'm restarting the RPC server process, but in general it would be nice to be able to use one server for completions in different projects/virtualenvs.

FWIW, I think it is the right decision to start up some sort of a wrapper in the target environment, for stability and flexibility, but I don't think it's necessary to over-engineer it. All it needs to do is to provide some bare-minimum-RPC without dependencies, so that jedi can simply execute the script without forcing the user to install anything. It also doesn't need to be asynchronous, but being able to cancel an operation would be nice. Maybe even something as simple as json-over-stdin/stdout with a couple of threads (to be able to send and receive cancellation signals) would work.

immerrr: Or one can even substitute json with pickle to be able to call standard library functions without any extra code.

davidhalter:
> I'm not sure I understand jedi/travis testing quite well, is it OK if I ask you to address the testing infrastructure (UPD: and I will do the tests themselves) ?

What do you mean by that? I'm certainly willing to help, I just don't know what you mean :)

>  I wonder if we could do this in ycmd API...

That would certainly be an option! I think at the same time it would also be nice to have something lightweight, pure Python.

jedi/evaluate/sys_path.py
```diff
@@ -10,24 +13,78 @@
...
-        p = _get_venv_sitepackages(venv)
-        if p not in sys_path:
-            sys_path.insert(0, p)
-
-        # Add all egg-links from the virtualenv.
+def get_venv_path(venv):
+    """Get sys.path for specified virtual environment."""
+    try:
+        sys_path = _get_venv_path_online(venv)
+    except Exception as e:
```
davidhalter: I don't like this. Exceptions should not be catched like that. Only catch what you really know can go wrong.

jedi/api/__init__.py
davidhalter: However that `VIRTUAL_ENV` variable will not change, just because you're activating a virtualenv. So this logic effectively doesn't do anything else than loading the same `sys.path`?!

Or are you thinking about changing the environment variable before each Jedi Script invocation? In that case I would rather add another parameter to Script.

cjrh:
I've made [vim-conda](https://github.com/cjrh/vim-conda) which modifies `$PATH` (for running code) and `sys.path` (for jedi-vim completion) depending on the selected conda environment.

I would much prefer to pass in a particular `sys.path` for completion as per this PR, rather than what is currently happening.  I just wanted to mention that hard-coded `'VIRTUAL_ENV'` statements make me a little nervous. Conda uses `CONDA_DEFAULT_ENV` (not a complete path) and `CONDA_ENV_PATH` (a complete path, but not of the virtual environment) as part of its internal mechanism for switching between envs.  I guess I just wanted offer a suggestion that it might be worthwhile to treat the handling of custom `sys.path` args and the use of venv/conda as separate concerns.  I have not looked at this PR in enough detail of offer any specific advice however.

davidhalter:
@cjrh I agree with you: It would be nice if we had a `sys.path` like in this PR. I would also be ok with a separate PR for that (if you or anyone else wants to do it). There has been a long time in which @immerrr has not replied and the code base has shifted anyway (so merge conflicts will be the case anyway). Just create a new PR if you want.

jedi/evaluate/sys_path.py
immerrr: The idea behind this catch-all expression is that whatever prevents this sys.path detection heuristic from working should not prevent normal jedi functioning and only report the error as a warning (which is done one line below that) as "An indication that something unexpected happened... The software is still working as expected."

immerrr:
@davidhalter i'm sorry for being rather inactive, my schedule is a bit tight.

I was under impression that you rather wanted something client-server like with a portable jedi server to be run by a python in the target environment (portable in a sense that it doesn't need to be installed in that environment).

davidhalter:
@immerrr No problem, I don't expect you to be active, thanks for the PR anyway :) Hmm you're probably right. I wasn't actually reading the old conversation anymore. I just realized that this solution was an improvment, but has still certain short comings.

I was waiting for the questions to be answered. There are some things in this PR that are really nice (if the virtualenv stuff gets merged or not, like egg-link resolution and `sys.path` api support.

jedi/api/__init__.py
immerrr: I gave this some thought at the time and I couldn't find a way around the client being able to supply the requested `sys.path` to Jedi in public API. Now, to my liking, full `sys_path` is too long for a proper interface parameter and most of the time it will refer to one or more virtualenvs, so probably specifying virtualenv paths will look better. For example, we could add `venvs=[path1, path2]` to use those venvs and `extra_sys_path=[path1, path2]` just in case someone wants, say, to additionally use some source code checkouts without installing them into the said venvs.

If this API addition goes through, I'm not sure if it is worth keeping the `VIRTUAL_ENV` logic: 
it boils down to having both `VIRTUAL_ENV`'s and jedi's `sys.path` in one interpreter and this is better accomplished the other way around. I mean, that logic starts with jedi on `sys.path` and then tries to add venv path afterwards, but as long as jedi has no external dependencies, we can just start up `PATH/TO/VENV/bin/python PATH/TO/JEDI/portable.py` with `portable.py` adding its location to `sys.path` and then running jedi as usual. That way it will have more binary compatibility with code in the venv.

immerrr:
hey @davidhalter is there anything else you want me to do here before merging?

jedi/evaluate/sys_path.py
davidhalter: Problem is: Whenever you do something like this, it's really hard to tell what the issue is from the outside. That's why I never do this.

jedi/api/__init__.py
davidhalter: > most of the time it will refer to one or more virtualenvs, so probably specifying virtualenv paths will look better. 

Why would it refer to multiple virtualenvs? That's not something that happens in Python itself?

> If this API addition goes through, I'm not sure if it is worth keeping the VIRTUAL_ENV logic: it boils down to having both VIRTUAL_ENV's and jedi's sys.path in one interpreter and this is better accomplished the other way around. I mean, that logic starts with jedi on sys.path and then tries to add venv path afterwards,

Hmm I know what you mean. I personally think that adding `sys_path` as an API is really valuable, because some people really want to play with it. Most of the time this is not the case and we should focus on what happens when it is `None`. This still includes virtual envs. However, I agree totally with the `portable.py` idea. I think Jedi should be started with just the virtualenv as its path. This is what I'm talking about when I talk about Jedi's future async server/client structure. You name a certain virtualenv and Jedi starts a new process that includes that virtualenv. This way, we don't have segfaults and all other crazy stuff...

What do you think? This would however mean that we don't include this pull request, we would rather just add the sys_path logic. And BTW this is not to say that your work is bad. I'm really thankful for it. It taught me a lot! :)

~ Dave

immerrr: > Why would it refer to multiple virtualenvs? That's not something that happens in Python itself?

Yes, that was a mistake. Some time ago I tried conserving precious SSD space using a virtualenv for project dependencies and another virtualenv for common tools, like ipython, and pudb, and profilers, and whatnot. But then I realized that with conda you don't need that as it hard-links packages where possible.

> What do you think? This would however mean that we don't include this pull request, we would rather just add the sys_path logic. And BTW this is not to say that your work is bad. I'm really thankful for it. It taught me a lot! :)

I am not completely against ditching this PR altogether, although I do think that this PR offers better support of the current Jedi workflow. It also doesn't seem to me that the new, arguably more correct, workflow involving `portable.py` and dropping all that `sys_path` magic is coming soon. But if it is and there's no time window during which this PR can be useful then feel free to close it.

davidhalter: > It also doesn't seem to me that the new, arguably more correct, workflow involving portable.py and dropping all that sys_path magic is coming soon.

Unfortunately and probably true :( Can you quote yourself from earlier how this PR improves virtualenvs? I still don't see it :)

immerrr: > Can you quote yourself from earlier how this PR improves virtualenvs?

- fetching `sys_path` from a live interpreter in the specified virtualenv, if possible
- if online fetching fails, offline detector now additionally follows `.pth` links (cannot seem to find the `addsitedir` invocation in the old code)
- `egg-link` files are consulted in all new path directories, not just the `site-packages` one

immerrr:
Ok, I'm probably going to close this, because the offline sys path detection after #636 is good enough for me. More accurate detection of paths that involves running a subprocess will invariably pick up some incompatible binary libraries which might crash the interpreter.

davidhalter:
Ok. Thanks still. I think some of this code might even be useful when we're trying to improve virtualenv stuff, but for now I agree that this is an issue.

PR #567, opened 2015-04-08
run buildout detection only once per buildout script

mfussenegger:
same as https://github.com/davidhalter/jedi/pull/546 but against the dev branch

### DIFF
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -8,6 +8,7 @@ from jedi.parser import Parser
 from jedi.evaluate.cache import memoize_default
 from jedi import debug
 from jedi import common
+from jedi import cache
 
 
 def get_sys_path():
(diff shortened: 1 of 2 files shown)


davidhalter:
Hmm why aren't you using the parser caches? I would probably prefer that to a memoize call, since that will cache the buildout scripts for future usages as well.

mfussenegger:
Saw your updated comment only after I've already changed it to use the memoize call. 
If you want I can try to use the parser caches instead.

davidhalter:
Yeah I guess I would prefer that, since it's not really clear why we need the memoize call there from just reading the code, it doesn't see like a very "slow" thing from the outside.

But nice fix anyways. It looks good.

mfussenegger:
Added another fixup which now uses the parser cache. Let me know if it's okay, I'll then squash the commits.

davidhalter:
Nice! I like it! I just realized that I screwed up the whole test suite a while ago. But I'll fix that and you can squash.

Thanks a lot.

mfussenegger:
Rebased and squashed it.

davidhalter:
Thanks so much for the effort :)

PR #581, opened 2015-05-06
Use repr() for the Whitespace repr

asmeurer:
This makes whitespace appear as `<Whitespace: '\n'>` instead of 
```
<Whitespace:
>
```


davidhalter:
Nice, thanks!

Issue #580, opened 2015-05-05
Semantic keyword completion


davidhalter:
It would be nice if Jedi supported semantic keyword completion. Currently Jedi only lists keywords and adds them to the completion list. This doesn't make sense.

For example:

- `as` should never be able to be completed at the start of a line. (except after a `\`)
- keywords like `continue`, `try`, etc. should only be added after whitespace.

This is the easy part. It's getting more complicated if you want to check if it's possible to write `if`, for example it cannot be written after a bracket `(`, but after a name `x`: `x if foo else 3`.
To do that, we can either write a lot of code by hand or maybe abuse the parser somehow (inluding Parser.error_statements).

squidarth:
@davidhalter I'm going to start taking a stab at this today

PR #586, opened 2015-05-11
Added support for correct continue & break autocompletion.

squidarth:
@davidhalter Resolves https://github.com/davidhalter/jedi/issues/580

### Approach I chose to take

I chose to add a `completion_names` methods to the keywords module, used for figuring out what keywords to add (now that we will no longer just be adding all available keywords to the list of completions).

`break` and `continue` statements don't require their own lines (`a +b;continue` is valid), so the rules I decided on was checking to see if the current `user_stmt` is a `Leaf` type.

I also separated out the `keywords` method in to `all_keywords` and `keyword`, to simplify the method, let me know what you think about this change.

### Issues

Had a few issues resolving this issue--first of all, I'm not totally sure what the behavior for the case of `b + continue` ought to be.  The `user_stmt` that gets returned in this case is of type `Whitespace`, I assume because of how the parser parses out the keywords `continue` and `break`.  Do you know if the parser handles these keywords differently, and if so, should I work on changing it to parse them as `Name` if they are part of a more complex expression? 

Let me know what you think!

### TODO

- [x] Figure out how to resolve `b + continue` issue

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -157,7 +157,7 @@ class Script(object):
                     if unfinished_dotted:
                         return completion_names
                     else:
-                        return keywords.keyword_names('import')
+                        return set([keywords.keyword('import').name])
 
             if isinstance(user_stmt, tree.Import):
                 module = self._parser.module()
@@ -168,7 +168,11 @@ class Script(object):
             if names is None and not isinstance(user_stmt, tree.Import):
                 if not path and not dot:
                     # add keywords
-                    completion_names += keywords.keyword_names(all=True)
+                    completion_names += keywords.completion_names(
+                        self._evaluator,
+                        user_stmt,
+                        self._pos,
+                        module)
                     # TODO delete? We should search for valid parser
                     # transformations.
                 completion_names += self._simple_complete(path, dot, like)
--- a/jedi/api/keywords.py
+++ b/jedi/api/keywords.py
@@ -5,7 +5,7 @@ from jedi._compatibility import is_py3
 from jedi import common
 from jedi.evaluate import compiled
 from jedi.evaluate.helpers import FakeName
-
+from jedi.parser.tree import Leaf
 try:
     from pydoc_data import topics as pydoc_topics
 except ImportError:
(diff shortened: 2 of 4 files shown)


davidhalter:
Nice. I like the general idea!

Note that I don't like the general way of how the completion code looks like. So don't be imitated by it. I don't get it either :)

> I'm not totally sure what the behavior for the case of b + continue ought to be.

`b + continue` is not a valid statement is therefore not fully parsed (parser error recovery).

squidarth:
Ok cool.  Just thinking about next steps here:

> b + continue is not a valid statement is therefore not fully parsed (parser error recovery).

What's the best way forward with this then? Seems like the options are either:

a. Handling it by including "continue" as an autocomplete term only if the user_stmt is a `Leaf` but not `Whitespace`
b. Not handle it.  In this case, I can remove the test cases I have for these and just submit this branch as is.

What do you think?

davidhalter:
Hmm. Good question!

You could check `Parser.error_statements` in there you would have more information about leaves that haven't been parsed. Side note: Not all ignored tokens are error statements yet, so if something's missing you're free to add it or report it.

That could help in figuring out what's under the cursor. Another way is to check `Script._parser.user_stmt()`, which doesn't get its statement from the syntax tree. It's a way to also get statements in docstrings and invalid code.

Just two thoughts, tell me if you don't get those two things. They are both pretty special!

squidarth:
@davidhalter Cool, I went with checking the error statements.  Let me know what you think of my approach, and my classification of `break` and `continue` as "leaf keywords".

This change resolves the issue that I had with the `b + continue` case (take a look at the test case I committed).

If this looks good, I'll start working on `try`

squidarth:
Weird failure.  having trouble running `tox` locally, but was able to run all the tests with `py.test` under python3...

davidhalter:
I think it looks pretty good, because its a really small solution. As long as it works, it should be fine I think.

So good job! :+1:

squidarth:
I tracked down the failure, and was finally able to reproduce it locally.

Interestingly enough, I'm getting the same failure on master, and it seems completely unrelated to my changes.

For some reason, when `py.test` is run with tox, it cannot read the file `jedi/test/completion/usages.py` due to the unicode character.  The code that is failing is in `jedi/test/run.py:293`.  When it reads the file, it thows a `UnicodeDecodeError`.

squidarth:
@davidhalter Ok, I fixed the build in  de7273e.  The change is somewhat unrelated to the other changes here, so I'd be happy to make it in a separate pull request..

To summarize that change briefly, in python 3, the encoding that `open` uses is platform dependent.  I set the encoding to use to be `utf-8`, as some of the files that we open have unicode characters.  This way, the test won't fail in environments where utf-8 is not the default encoding.

If you're happy with this change, this is good to merge.

davidhalter:
It looks very good! Thanks a lot!

If you're adding more changes in the coming two weeks, know that I'm on vacation. I will be back in June.

squidarth:
Excellent, thanks.

Have fun!

PR #612, opened 2015-07-28
fix a set comprehension issue

tjwei:
when running the following code
```python
import jedi
statement = jedi.names(source='b = {i for i in range(10)}')[0]
scope, evaluator = statement.parent()._definition, statement._evaluator
evaluator.find_types(scope, statement.name)
```
jedi gets an error
```
...
  File "/usr/local/lib/python2.7/dist-packages/jedi/evaluate/__init__.py", line 246, in _eval_atom
    return [iterable.Comprehension.from_atom(self, atom)]
  File "/usr/local/lib/python2.7/dist-packages/jedi/evaluate/iterable.py", line 125, in from_atom
    return mapping[atom.children[0]](evaluator, atom)
KeyError: <Operator: {>
```
This is because in _eval_atom(self, atom) of jedi/evaluation/__init__.py, set comprehension is treated just like list and generator comprehension at this line
```python
if isinstance(comp_for, tree.CompFor):
                    return [iterable.Comprehension.from_atom(self, atom)]
```
however, set comprehension is not yet supported in iterable.py 
So this patch fix the issue, allowing set comprehension to be treated like dict comprehension.


davidhalter:
Thanks! Could you move the tests to `jedi/test/completion/comprehensions.py`? Since all the other tests are there I think it will be cleaner.

tjwei:
Sure, done. It is much cleaner to put them together.

test/completion/comprehensions.py
```diff
@@ -50,6 +50,13 @@
 ##? str()
 {a-1:b for a,b in {1:'a', 3:1.0}.items()}[0]
 
+# with a set literal
+#? int()
+[a for a in {1, 2, 3}][0]
+
+##? int()
```
davidhalter: Is this intentional? Because with two `##` the test is not enabled.

tjwei: It was not intentional, my mistake. I updated the test.
A few additional tests are disabled intentionally, because these tests needs full support of dict/set generator class.
I am happy to help to add this generator class, but really need to study the source further.

davidhalter: Would be awesome if you did. :)

davidhalter:
Oh btw: Please make the next pull request against dev. Thanks :)

PR #584, opened 2015-05-09
AttributeError with ValueError in _compatibility.py:find_module_py33

blueyed:
I came across the following error (via jedi-vim).

      File "…/vim/neobundles/jedi/jedi/jedi/_compatibility.py", line 28, in find_module_py33
        raise ImportError("Originally ValueError: " + e.message)

If I remember correctly it should use `e.args` here.

Traceback:

> AttributeError: 'ValueError' object has no attribute 'message'

    Traceback (most recent call last):
      File "…/pyenv/3.4.3/lib/python3.4/importlib/__init__.py", line 68, in find_loader
        loader = sys.modules[name].__loader__
    AttributeError: 'Nvim' object has no attribute '__loader__'
    During handling of the above exception, another exception occurred:
    Traceback (most recent call last):
      File "…/vim/neobundles/jedi/jedi/jedi/_compatibility.py", line 24, in find_module_py33
        loader = importlib.find_loader(string)
      File "…/pyenv/3.4.3/lib/python3.4/importlib/__init__.py", line 76, in find_loader
        raise ValueError('{}.__loader__ is not set'.format(name))
    ValueError: vim.__loader__ is not set
    During handling of the above exception, another exception occurred:
    Traceback (most recent call last):
      File "…/vim/neobundles/jedi/jedi_vim.py", line 107, in wrapper
        return func(*args, **kwargs)
      File "…/vim/neobundles/jedi/jedi_vim.py", line 206, in goto
        definitions = script.goto_definitions()
      File "…/vim/neobundles/jedi/jedi/jedi/api/__init__.py", line 362, in goto_definitions
        definitions = set(self._prepare_goto(goto_path))
      File "…/vim/neobundles/jedi/jedi/jedi/api/__init__.py", line 294, in _prepare_goto
        scopes = self._evaluator.eval_element(eval_stmt)
      File "…/vim/neobundles/jedi/jedi/jedi/evaluate/cache.py", line 41, in wrapper
        rv = function(obj, *args, **kwargs)
      File "…/vim/neobundles/jedi/jedi/jedi/evaluate/__init__.py", line 169, in eval_element
        return self._eval_atom(element)
      File "…/vim/neobundles/jedi/jedi/jedi/evaluate/__init__.py", line 230, in _eval_atom
        return self.find_types(scope, atom, stmt.start_pos, search_global=True)
      File "…/vim/neobundles/jedi/jedi/jedi/evaluate/__init__.py", line 120, in find_types
        return f.find(scopes, search_global)
[...]

PR #616, opened 2015-08-13
find_module_py33: use str(e) with ValueError

WoLpH:
ValueError has no message attribute.

Fixes https://github.com/davidhalter/jedi/issues/584


WoLpH:
Ok... ignore this. Apparently the hub command defaults to merging to master ;)

PR #624, opened 2015-09-08
imports - module_path - referenced before assignment

alyssackwan:
Ran into this via Elpy / Emacs.

### DIFF
--- a/jedi/evaluate/imports.py
+++ b/jedi/evaluate/imports.py
@@ -304,6 +304,7 @@ class Importer(object):
                 return []
             else:
                 debug.dbg('search_module %s in paths %s', module_name, paths)
+                module_path = None
                 for path in paths:
                     # At the moment we are only using one path. So this is
                     # not important to be correct.


davidhalter:
Thank you! :)

Though, I'm not sure why this would even happen. Do you have an example? Or a test?

alyssackwan:
Here's the Elpy output. Haven't looked into this deeply enough to come up with a real unit test:

Elpy Error

The backend encountered an unexpected error. This indicates a bug in
Elpy. Please open a bug report with the data below in the Elpy bug tracker:

https://github.com/jorgenschaefer/elpy/issues/new

```
Error Message

local variable 'module_path' referenced before assignment

Configuration

Virtualenv........: lib.util (/Users/alyssakwan/.pyenv/versions/lib.util)
RPC Python........: 3.4.3 (/Users/alyssakwan/.pyenv/versions/lib.util/bin/python)
Interactive Python: ipython (/Users/alyssakwan/.pyenv/versions/lib.util/bin/ipython)
Emacs.............: 24.5.1
Elpy..............: 1.9.0
Jedi..............: 0.9.0
Rope..............: Not found (0.10.2 available)
Importmagic.......: 0.1.3
Autopep8..........: 0.1.3
Syntax checker....: flake8 (/Users/alyssakwan/.pyenv/versions/lib.util/bin/flake8)

Traceback

  File "/Users/alyssakwan/.emacs.d/elpa/elpy-20150830.510/elpy/jedibackend.py", line 216, in run_with_debug
    return getattr(script, name)()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/api/__init__.py", line 521, in call_signatures
    self.source, self._pos)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/cache.py", line 93, in wrapper
    value = next(generator)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/cache.py", line 115, in cache_call_signatures
    yield evaluator.eval_element(call)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 169, in eval_element
    return self._eval_atom(element)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 230, in _eval_atom
    return self.find_types(scope, atom, stmt.start_pos, search_global=True)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 120, in find_types
    return f.find(scopes, search_global)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/debug.py", line 52, in wrapper
    result = func(*args, **kwargs)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 87, in find
    types = self._names_to_types(names, search_global)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 233, in _names_to_types
    new_types = _name_to_types(self._evaluator, name, self.scope)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 282, in _name_to_types
    types += imports.ImportWrapper(evaluator, name).follow()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 94, in follow
    types = importer.follow()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 252, in follow
    return self._do_import(self.import_path, self.sys_path_with_modifications())
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 281, in _do_import
    bases = self._do_import(import_path[:-1], sys_path)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 316, in _do_import
    if module_path is None:
UnboundLocalError: local variable 'module_path' referenced before assignment

  File "/Users/alyssakwan/.emacs.d/elpa/elpy-20150830.510/elpy/jedibackend.py", line 306, in run_with_debug
    return getattr(script, name)()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/api/__init__.py", line 521, in call_signatures
    self.source, self._pos)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/cache.py", line 93, in wrapper
    value = next(generator)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/cache.py", line 115, in cache_call_signatures
    yield evaluator.eval_element(call)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 169, in eval_element
    return self._eval_atom(element)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 230, in _eval_atom
    return self.find_types(scope, atom, stmt.start_pos, search_global=True)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/__init__.py", line 120, in find_types
    return f.find(scopes, search_global)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/debug.py", line 52, in wrapper
    result = func(*args, **kwargs)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 87, in find
    types = self._names_to_types(names, search_global)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 233, in _names_to_types
    new_types = _name_to_types(self._evaluator, name, self.scope)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/finder.py", line 282, in _name_to_types
    types += imports.ImportWrapper(evaluator, name).follow()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 94, in follow
    types = importer.follow()
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 252, in follow
    return self._do_import(self.import_path, self.sys_path_with_modifications())
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 281, in _do_import
    bases = self._do_import(import_path[:-1], sys_path)
  File "/Users/alyssakwan/.pyenv/versions/lib.util/lib/python3.4/site-packages/jedi/evaluate/imports.py", line 316, in _do_import
    if module_path is None:
UnboundLocalError: local variable 'module_path' referenced before assignment

Jedi Debug Information

[N] dbg: Parsed /Users/alyssakwan/home/_/projects/lib.util/src/cloverhealth/lib/util/process.py, with 0 parsers in 4 splits.
[N] dbg: eval_element <Name: Config@10,8>@(10, 8)
[N] dbg: finder.filter_name "<Name: Config@10,8>" in (<eFunction of <Function: __init__@9-13>>): '[<Name: Config@4,41>]'@(10, 8)
[N] dbg: search_module 'cloverhealth' in '/Users/alyssakwan/home/_/projects/lib.util/src/cloverhealth/lib/util/process.py'
[N] dbg: search_module 'cloverhealth.lib' in paths ['/Users/alyssakwan/home/_/projects/lib.util/src/cloverhealth']
[N] dbg: search_module 'cloverhealth.lib.util' in paths []

```

Reproduction:

```Python
import jedi

source = '''\
"""OS process framework."""


from cloverhealth.lib.util.config import Config


class Process(object):

    def __init__(self, config_handlers):
        self._config = Config()
        self._whoami()

    def _whoami(self):
        pass
'''

script = jedi.Script(column=30, source=source, encoding='utf-8', line=10, path='/Users/alyssakwan/home/_/projects/lib.util/src/cloverhealth/lib/util/process.py')
script.call_signatures()
```

davidhalter:
Hmm might be a general issue with imports. Need a test. If you don't do it I will. The test will be your example with a different library (that Jedi is able to import). But until then I will leave this issue open and merge it once the test is here.

alyssackwan:
I'd be more than happy to pitch in. New to Python so better understanding metaprogramming and the AST is a good thing. Let me know where to start.

davidhalter:
Hmm I guess first we have to find the actual problem (in Jedi logic).

Obviously it has to do something with the missing `paths` (see also debug output). However, we don't know yet, why that list is empty.

alyssackwan:
Ah. So my setup:

1. ``setup.py`` gets everything from ``src/``.
2. In Emacs, I'm using Elpy with Jedi as a backend. I also use Projectile.
3. To manage multiple projects in the same instance of Emacs, I have a ``.dir-locals.el``. It calls ``pyvenv-workon`` from the pyvenv Elisp package. Therefore, Emacs will load it for every file / buffer in that directory tree.
3. I also do automagic by having Emacs use the Elpy and Projectile hooks to make sure that the Python project root has a ``.dir-locals.el`` in it. If it doesn't, it creates a ``pyenv virtualenv`` (I'm on OS X with Homebrew), and writes the ``.python-version`` file for pyenv and creates the ``.dir-locals.el`` file for Emacs.
4. The Python package is __not__ installed, either with ``python setup.py install`` or ``python setup.py develop``. I use Tox to drive my tests so I don't need to side-effect the virtualenv with package installation. __(This can be changed.)__

Therefore, it seems to me that there's several possible failure modes:

- Jedi may be running against a Python env that is not the virtualenv set by ``pyvenv-workon``.
   - I assume this is the responsibility of Elpy to invoke Jedi and therefore use the right Python env.
   - Given that Elpy and Elisp pyvenv was written by the same person (the latter to support the former) I doubt this is the case.
- If Jedi is using the virtualenv, the package I'm working on just needs to be ``python setup.py develop``.
- Alternately, Jedi is doing magic on its own to figure out PYTHONPATH.
   - If so, the fact that my ``$PWD`` for the file / buffer __should__ be the project root breaks things, since everything is tucked under ``src/``.

Thoughts?

I'm going to just ``python setup.py develop`` and see if it works. I'll let you know.

alyssackwan:
No joy. But potential breakthrough! You comment on line 297 re: namespace packages. I'm using nested namespace packages. The debug output shows that the first concrete package under the nested packages is the one that's empty.

I'm going to trace your code further tonight. Too many fires to put out today. (Symbol completion at point would be __really__ helpful right now, LOL.)

davidhalter:
I think your setup is not really important. It's nice to know, but I think it's an actual Jedi issue that will be reproducible in every environment.

> I'm using nested namespace packages. The debug output shows that the first concrete package under the nested packages is the one that's empty.

So this might actually be the problem?! Maybe we have to solve the paths issue there. :) So it's all about why that paths variable is empty in the first place. If namespace packages are the reason, we should fix those, because it would be something more that Jedi understands. (And there's no need for this small fix).

alyssackwan:
ITA. I'm new to Python (only 2 months), and only Python 3, and only with setuptools, so my knowledge of namespace packages is limited to that context. So the syspath stuff is a little opaque to me right now (and do distutils or setuptools have utility methods to aid with this?). But I'll try to pitch in. LMK where to head and I'll do the legwork.

davidhalter:
> I'm new to Python (only 2 months), and only Python 3

Time to learn about namespace packages :-P Actually not. It's not that important, but for Jedi it might be :)

> So the syspath stuff is a little opaque to me right now (and do distutils or setuptools have utility methods to aid with this?).

I'm not sure what you want to know. I wrote a bit about namespace packages below. In general the `sys.path` is generated by a python module called `site.py`. `distutils` and `setuptools` are just tools to push stuff into the `sys.path` (or to add new paths via `.pth` files). 

------

This is the place where namespace packages are tested:
https://github.com/davidhalter/jedi/blob/master/test/test_evaluate/test_namespace_package.py

And this is how a namespace package typically looks like:

https://github.com/davidhalter/jedi/blob/master/test/test_evaluate/namespace_package/ns1/pkg/__init__.py
```
try:
    import pkg_resources
    pkg_resources.declare_namespace(__name__)
except ImportError:
    import pkgutil
    __path__ = pkgutil.extend_path(__path__, __name__)
```

So it uses pkgutil to extend the `__path__` attribute that happens to be an attribute for every module.

Also Jedi doesn't support implicit namespace packages, that have been introduced in Python 3.3, AFAIK. In Python 3.3 you can just skip the `__init__.py` file and you have a namespace package.

alyssackwan:
Thank you. I'll look at this over the weekend. Re: implicit namespace packages - these break pretty much everything right now, so I wouldn't worry about supporting it yet. Plus most libs which need 2 and 3 compatibility can't do it, so whatevs.

scribu:
Here's an easy to reproduce test:

1. Run `pip install Flask-Admin==1.4.2`
2. Run the following script:

```python
import jedi

CODE = """
from flask_admin.contrib.sqla import ModelView
"""

script = jedi.Script(source=CODE, line=2, column=38)

print(script.goto_definitions())
```

You should get this error:

```
  File "test.py", line 9, in <module>
    print(script.goto_definitions())
  File "/Users/scribu/git/jedi/jedi/api/__init__.py", line 188, in goto_definitions
    definitions = helpers.evaluate_goto_definition(self._evaluator, leaf)
  File "/Users/scribu/git/jedi/jedi/api/helpers.py", line 161, in evaluate_goto_definition
    return evaluator.goto_definitions(leaf)
  File "/Users/scribu/git/jedi/jedi/evaluate/__init__.py", line 457, in goto_definitions
    return imports.ImportWrapper(self, name).follow()
  File "/Users/scribu/git/jedi/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/scribu/git/jedi/jedi/evaluate/imports.py", line 94, in follow
    types = importer.follow()
  File "/Users/scribu/git/jedi/jedi/evaluate/cache.py", line 41, in wrapper
  File "/Users/scribu/git/jedi/jedi/evaluate/imports.py", line 252, in follow
    return self._do_import(self.import_path, self.sys_path_with_modifications())
  File "/Users/scribu/git/jedi/jedi/evaluate/imports.py", line 316, in _do_import
    if module_path is None:
UnboundLocalError: local variable 'module_path' referenced before assignment
```

scribu:
`flask_admin/contrib/__init__.py` looks like this:

```
try:
    __import__('pkg_resources').declare_namespace(__name__)
except ImportError:
    pass
```

The problem is in `ModuleWrapper.py__path__`.

If I make the following brute-force change, it correctly finds the subpackage:

```diff
- other = os.path.join(s, unicode(self.name))
+ other = os.path.join(s, 'flask_admin', unicode(self.name))
```

Not sure what defining a namespace in a subpackage is supposed to mean. Nevertheless, the python interpreter doesn't have any issues with it.

scribu:
Looks like I've confirmed this comment :P

```
                # It's possible that by giving it always the sys path (and not
                # the __path__ attribute of the parent, we get wrong results
                # and nested namespace packages don't work.  But I'm not sure.
                paths = base.py__path__(sys_path)
```

davidhalter:
#727 is closer to solving this issue. Therefore I'm closing this one.

dzhwinter:
hha

PR #637, opened 2015-10-21
Couple build system improvements

immerrr:
Separate PR for couple of build system issues:
- work around an issue in pytest doctest plugin
- enable [new travis infrastructure](http://docs.travis-ci.com/user/migrating-from-legacy/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 env:
     - TOXENV=py26
     - TOXENV=py27
--- a/conftest.py
+++ b/conftest.py
@@ -1,8 +1,9 @@
 import tempfile
 import shutil
 
-import jedi
+import pytest
 
+import jedi
 
 collect_ignore = ["setup.py"]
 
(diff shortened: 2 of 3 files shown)


pytest.ini
```diff
@@ -8,3 +8,4 @@ norecursedirs = .* docs completion refactor absolute_import namespace_package sc
 # fine as long as we are using `clean_jedi_cache` as a session scoped
 # fixture.
 usefixtures = clean_jedi_cache
+testpaths = test
```
davidhalter: The Jedi directory needs to be included as well, because it contains doctests.

immerrr: Ah, that starts to make sense. When I added the directory back, the failure about missing fixture returned and now it started making sense: `clean_jedi_cache` fixture is not found for doctests module running in `jedi` directory, because it is in a `conftest.py` in a  different directory.

immerrr: It is simpler to reproduce with `py.test --fixtures jedi/`.

immerrr: Fixed by moving `clean_jedi_cache` to top-level `conftest.py`.  Now there are two pieces of code that manage temporary directories for cache: as a session-level fixture and in `pytest_configure`/`pytest_unconfigure` hooks. I'm not sure which one is better to stay, so left them both in for now.

davidhalter:
Nice! Thanks for the work. It makes things simpler.

PR #635, opened 2015-10-21
Function return type in completion results

elad661:
Hi!

For GNOME Builder, we want to have the completion suggestion list show the return value type of each function. Right now, jedi doesn't export this information in its `Completion` class. 

Would it be possible for you to add this as an attribute to the class? Jedi can already parse docstrings and such quite well to guess the return types of functions, so it'd be really convenient (and probably better for performance) if jedi would provide this information as an attribute of the `Completion` class.

btw, feel free to add GNOME Builder to the "jedi users" section of your README file :)

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -33,5 +33,6 @@ Savor d'Isavano (@KenetJervet) <newelevenken@163.com>
 Phillip Berndt (@phillipberndt) <phillip.berndt@gmail.com>
 Ian Lee (@IanLee1521) <IanLee1521@gmail.com>
 Farkhad Khatamov (@hatamov) <comsgn@gmail.com>
+Elad Alfassa (@elad661) <elad@fedoraproject.org>
 
 Note: (@user) means a github user name.
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -308,6 +308,17 @@ class BaseDefinition(object):
 
         return '.'.join(path if path[0] else path[1:])
 
+    @property
+    def return_type(self):
+        if self.type != 'function':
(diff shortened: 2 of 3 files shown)


elad661:
I hope this is good enough. Let me know if there's anything you need me to change.

patrys:
pytest >= 2.8.0 seems to be broken, this is what causes strange output on Travis.

@elad661 you may need to add `return_type` to `dont_scan` list in `test.test_api.test_call_signatures.test_signature_is_definition`. It's failing for me as the test case hits the `AttributeError` explicitly raised for non-functions.

elad661:
@patrys done.

immerrr:
Pytest issue is fixed on current dev, if you rebase on top of it you'll be able to make it green.

davidhalter:
Ok. I would like to have a few tests for this. Also when is `r.name` `None`?

elad661:
Regarding tests: I'm a bit lost trying to understand your test layout... would you want this kind of test in a new file starting with `test_integration_` or someplace else?

Regarding `r.name` being `None`, you said "Not everything might have a name" and I missed the word "attribute" there, so I thought you mean the name might be `None` in some cases. Should I change it to use hasattr(r, "name")?

davidhalter:
> Regarding r.name being None, you said "Not everything might have a name" and I missed the word "attribute" there, so I thought you mean the name might be None in some cases. Should I change it to use hasattr(r, "name")?

I'm actually just not sure if everything has a name attribute. For example what would the name be for lambdas? I don't really know. It's probably None or an AttributeError. It's something that you would have to test. :)

Issue #589, opened 2015-05-19
AttributeError: 'MergedNamesDict' object has no attribute 'get'


jorgenschaefer:
A user reported this error using Jedi 0.9.0:

```
Traceback

Traceback (most recent call last):
  File "/home/user/.emacs.d/elpa/elpy-20150226.1148/elpy/rpc.py", line 100, in handle_request
    result = method(*params)
  File "/home/user/.emacs.d/elpa/elpy-20150226.1148/elpy/server.py", line 86, in rpc_get_calltip
    get_source(source), offset)
  File "/home/user/.emacs.d/elpa/elpy-20150226.1148/elpy/server.py", line 48, in _call_backend
    return meth(*args, **kwargs)
  File "/home/user/.emacs.d/elpa/elpy-20150226.1148/elpy/jedibackend.py", line 126, in rpc_get_calltip
    "index": call.index,
  File "/home/user/.local/lib/python2.7/site-packages/jedi/api/classes.py", line 653, in index
    if self._index >= len(self.params):
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/user/.local/lib/python2.7/site-packages/jedi/api/classes.py", line 348, in params
    sub = followed.get_subscope_by_name('__init__')
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/representation.py", line 466, in get_subscope_by_name
    for s in self.py__mro__(self._evaluator):
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/representation.py", line 429, in py__mro__
    for cls_new in mro_method(evaluator):
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/representation.py", line 409, in py__mro__
    for cls in self.py__bases__(self._evaluator):
  File "/home/user/.local/lib/python2.7/site-packages/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
[...]

PR #644, opened 2015-11-01
implemented .get() for MergedNamesDict, see #589

avkoval:
Made .get() and added test for this use-case for issue #589

### DIFF
--- a/jedi/parser/fast.py
+++ b/jedi/parser/fast.py
@@ -74,6 +74,12 @@ class MergedNamesDict(object):
     def __iter__(self):
         return iter(set(key for dct in self.dicts for key in dct))
 
+    def get(self, key, replacement=None):
+        for d in self.dicts:
+            if key in d:
+                return d[key]
+        return replacement
+
     def __getitem__(self, value):
         return list(chain.from_iterable(dct.get(value, []) for dct in self.dicts))
 
--- a/test/test_parser/test_fast_parser.py
+++ b/test/test_parser/test_fast_parser.py
@@ -438,3 +438,8 @@ def test_string_literals():
 
     script = jedi.Script(dedent(source))
     assert script.completions()
+
+def test_nested_get():
+    """Test merged dict get()"""
+    s = jedi.Script(u'from email import message', 1, 22, None, 'utf-8')
+    s.goto_assignments()


davidhalter:
But what's the actual problem? How and where does this occur? This is what I want to know first.

davidhalter:
But thanks anyway for the implementation.

Sorry for not just merging. I have realized along the line that bugs like this one haunts you somewhere later in point. So I always want to know what exactly the issue is. It's so easy to hot-fix something in Python with try/except.

RichardBronosky:
Thank you for this patch! Confirmed to fix https://github.com/davidhalter/jedi/issues/589#issuecomment-161541334 and https://github.com/davidhalter/jedi/issues/589#issuecomment-152909087 added to https://github.com/sadovnychyi/autocomplete-python/commit/b4e575b08cc8acce6fb640a6a3afc92bff78ab70

mathieulongtin:
Will this get merged eventually? I get this error using YouCompleteMe's latest as of today.

davidhalter:
No it will not. You could as well be doing:

   try:
       jedi.Script.completions
   except:
       pass

This pull request is not a fix, because it doesn't solve the problem. It might be the fix, but I want to know why this happens, first.

mathieulongtin:
My stack trace is similar to issue #589. I'm not sure what causes it to happen in the first place, but it seems that a MergedNamesDict ends up in another MergedNamesDict. Since the `__getitem__` of that class uses `get`, probably assuming that it only has dict inside, it fails.

When I saw that stack trace, I added a `get` like this, so it would behave like a dict.
```
    def get(self, name, default=None):
        try:
            return self[name]
        except KeyError:
            return default
```
Not quite the same as for this patch, but anyway. My `jedihttp*stderr.log` have been empty since then, and I get much better name suggestions than before, because my jedi process doesn't die half the time.

davidhalter:
> MergedNamesDict ends up in another MergedNamesDict

Yes, but I this shouldn't even happen. I think we have a bigger issue there.

davidhalter:
I have fixed it in a branch.

What this issue showed me is that a lot of people try to solve stuff with try/except in Python instead of fixing the actual issue, which makes me a bit sad.

avkoval:
Right, I've seen your comment (how to make correct fix) but it would take a lot of time for me to get into the whole picture/details. Thank you for applying correct fix here!

davidhalter:
Well, it's not very much a critic of your approach here, I'm just seeing this again and again :)

Issue #170, opened 2013-03-14
Use function annotations


davidhalter:
Jedi basically supports function annotations. I never implemented the final stage of autocompletion, because function annotations have not been defined to be anything specific in Python 3. But probably it would be a good idea to implement it. (~ 10 line patch for Jedi)

I thought about it for a little while and then remembered that [pep 3107][1] talks about this:

> Despite yet more discussion, it was decided not to standardize a mechanism for annotation interoperability. Standardizing interoperability conventions at this point would be premature. We would rather let these conventions develop organically, based on real-world usage and necessity, than try to force all users into some contrived scheme.

I realized that Jedi is exactly the software that should try to define the usage of these function annotations.

I thought about this again, because @ceronman posted this article: http://ceronman.com/2013/03/12/a-powerful-unused-feature-of-python-function-annotations/

[1]: http://www.python.org/dev/peps/pep-3107/

leebrooks0:
What is the status here re autocompletion? It seems like it is easy to implement (first comment says about 10 lines of code).

dbrgn:
For the record:

https://www.python.org/dev/peps/pep-0483/
https://www.python.org/dev/peps/pep-0484/

davidhalter:
@leebrooks0 The status is that we haven't implemented anything. A path at this point would be more complicated than 10 LOC, because there's the whole mypy stuff now (as defined in PEP 484).

@dbrgn Thanks!

mikn:
With https://github.com/ambv/typehinting now accepted and slated for release in 3.5, it would be really cool if it could go into Jedi. Is there any plan for it? @davidhalter

davidhalter:
Well, there's no specific plan, because I don't really have time right now to do it. But if someone's willing to do it, I would be happy to support.

mikn:
If you have a concrete outline or idea of how to implement it within jedi,
I may have time. :)

davidhalter:
Sorry for the big delay :) I forgot about it.

But: I guess we should start by ignoring the `typing` module. We should just make it possible for Jedi to understand normal references (should be fairly easy). The annotations are already scanned. Maybe you need to improve the parser as well.

Once that's merged, we can start dealing with the special cases: e.g. https://www.python.org/dev/peps/pep-0484/#forward-references. And in the end we start to support the typing module.

carlsverre:
[...]

PR #659, opened 2015-12-13
initial poc pep-0484 type hints

reinhrst:
I'm taking a swipe at #170 , and this is just to see if I'm on the right track. jedi/evaluate/pep0484.py contains a checklist with functionality that needs to be implemented on this end (I believe that the first 4 or 5 should be relatively straight-forward to implement, maybe even all with exception of the stub files, especially if they live in a different directory). Then in addition, jedi needs to understand all the subtleties of the typing module, but that's another chapter. Even the current very small change would already fix 50% of the times that I don't get completions in my current project.

Things that I'd like some feedback on (although feel free to shoot at other things as well):
- I'm currently getting the annotations from the parsed code. Alternatively one could ask the python interpreter and use the `.__annotations__`. Which is preferred?
- My `if` and `assert` statements to look for annotations seems to hold in all the tests (in python 3.4). Are the assumptions there correct? Should I perhaps use some less-defensive programming style, to allow for future python-syntax additions?
- I can't get tox to run, so only ran the tests in python 3.4, but I'm pretty sure the test will fail in 2.x because of the syntax. How do I guard the tests such that they only run in >=3.2 ?
- Right now I have annotations take precedence over docstring in case they are present. Is this the way to go, or should we perhaps merge them?
- PEP 0484 supports decorators for switching off type hinting in some modules/functions. For now I didn't include them (even on the todo list) since worst case one will just get wrong completions, right? Or should we allow for them?

### DIFF
--- a/test/completion/pep0484.py
+++ b/test/completion/pep0484.py
@@ -1,0 +1,14 @@
+""" Pep-0484 type hinting """
+
+# -----------------
+# sphinx style
+# -----------------
+def typehints(a, b: str, c: int, d:int = 4):
+    #? 
+    a
+    #? str()
+    b
+    #? int()
+    c
+    #? int()
+    d
--- a/jedi/evaluate/finder.py
+++ b/jedi/evaluate/finder.py
@@ -22,6 +22,7 @@ from jedi.evaluate import representation as er
 from jedi.evaluate import dynamic
 from jedi.evaluate import compiled
 from jedi.evaluate import docstrings
+from jedi.evaluate import pep0484
 from jedi.evaluate import iterable
 from jedi.evaluate import imports
 from jedi.evaluate import analysis
@@ -354,6 +355,11 @@ def _eval_param(evaluator, param, scope):
     if isinstance(func, er.InstanceElement) \
             and func.instance.is_generated and str(func.name) == '__init__':
         param = func.var.params[param.position_nr]
+
+    # Add pep0484 type hints
+    pep0484_hints = pep0484.follow_param(evaluator, param)
+    if pep0484_hints:
+        return pep0484_hints
 
     # Add docstring knowledge.
     doc_params = docstrings.follow_param(evaluator, param)
--- a/jedi/evaluate/pep0484.py
+++ b/jedi/evaluate/pep0484.py
@@ -1,0 +1,34 @@
+"""
+PEP 0484 ( https://www.python.org/dev/peps/pep-0484/ ) describes type hints
+through function annotations. There is a strong suggestion in this document
+that only the type of type hinting defined in PEP0484 should be allowed
+as annotations in future python versions.
+
+The (initial / probably incomplete) implementation todo list for pep-0484:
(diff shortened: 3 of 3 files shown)


test/completion/pep0484.py
```diff
@@ -0,0 +1,14 @@
+""" Pep-0484 type hinting """
+
+# -----------------
+# sphinx style
```
reinhrst: Oops, that's what you get for copy-pasting other files :(

davidhalter:
If it's not a big hassle for you, I would be very grateful if you start working
on this on the linter branch. I'm going to merge it upstream very soon. This
would be easier for you (easier to understand) as well.

> I'm currently getting the annotations from the parsed code. Alternatively one could ask the python interpreter and use the .__annotations__. Which is preferred?

Jedi only supports the parser, you cannot just magically use `__annotations__`.
However we still want type inference there. What I would suggest is something
that already exists within Jedi in some other forms (e.g. `py__mro__`,
`py__bases`, etc) - we recreate `__annotations__` in a static analysis kind of
way (https://github.com/davidhalter/jedi/blob/linter/jedi/evaluate/representation.py#L17):

'
```

class Function():  # In evaluate.representation
    @memoize_default()
    def py__annotations__(self):
        parser_func = self.base
        dct = {'return': self._evaluator.eval_element(parser_func.annotation)}
        for param in parser_func.params:
            dct[param.name.value] = self._evaluator.eval_element(param.annotation)
        return dct
```

With this function you basically have everything you need.

> My if and assert statements to look for annotations seems to hold in all the tests (in python 3.4). Are the assumptions there correct? Should I perhaps use some less-defensive programming style, to allow for future python-syntax additions?

Don't care for future additions. That's currently fine. Any future
modifications should be tested separately. But as hinted above, you shouldn't
be using the parser tree that much in the first place. Just use the
`annotation` properties (they are already defined in parser.tree`). I think you
stil have to implement `Param.annotation`, though.

> I can't get tox to run, so only ran the tests in python 3.4, but I'm pretty sure the test will fail in 2.x because of the syntax. How do I guard the tests such that they only run in >=3.2 ?

Very good question again. Yes we need to limit it somehow. I don't know yet how
we would do that best. If you have a clever idea, let me know that. (Maybe
something like `py > 2` in the first line of the file? And then parser for it?)
But if you're not checking for that, I will still merge. It's good enough if
the tests for any Python version passes. I can then do the rest from there.

> Right now I have annotations take precedence over docstring in case they are present. Is this the way to go, or should we perhaps merge them?

Wow you're on the right track! :) I'm not sure if I can  really answer that.
What do you think? I would tend to say they hold about the same level of
priority.

> PEP 0484 supports decorators for switching off type hinting in some modules/functions. For now I didn't include them (even on the todo list) since worst case one will just get wrong completions, right? Or should we allow for them?

I agree that ignoring them for now is probably best. This would be a "goodie"
in the end, but we really don't need to do that now. I think they are very
rarely used, anyways.

reinhrst:
Will open new PR on that branch.

PR #664, opened 2015-12-22
simple typehints in comments

reinhrst:
I'm not yet happy with it, but does this look halfway decent to you? in `finder.py:_remove_statements` there is actually a comment that docstrings should be handled in the evaluator, but I didn't quite see how that made sense.


davidhalter:
You're right! I don't see how that makes sense either. I think we returned something very weird there, previously. 

IMO this is too complicated: We shouldn't be doing all that stuff by hard-coding. This stuff should be done by the logic of the tree:

```leaf = stmt.last_leaf().get_next()```

Both these methods don't exist (but `first_leaf` and `get_previous` do exist, so you get the idea).

I have also realized that just fetching the comments is probably not the best idea (AFAIK it needs to be on the same line). So for this I would probably just check the prefix itself: `re.match(r'[ \t]*#\s*type:\s*([^#\r\n]*)', leaf.prefix)`.

Please also test the cases where the comment is on the next line (should be ignored).

**Comment by reinhrst** [comment]
Right, makes sense.

I also notice that there's an issue with python 2.7 in the test; will look at that too.

**Comment by davidhalter** [comment]
Let me know when you're done with this.

**Comment by reinhrst** [comment]
Yep --- will be past the `typing` module.

**Comment by reinhrst** [comment]
I thought about quickly adding this one as well :)

> I have also realized that just fetching the comments is probably not the best idea (AFAIK it needs to be on the same line). So for this I would probably just check the prefix itself: re.match(r'[ \t]*#\s*type:\s*([^#\r\n]*)', leaf.prefix).

I see what you're doing here. Just wondering whether this isn't something that should be fixed in the parser itself? I see how it makes sense in may cases to add a comment as a prefix to the next statement:
```
# do something with x
x.doSomething()
```
However when the comment is on the same line as a statement, then clearly it says something about that statement. Going to the next node and checking if perhaps it has some data that pertains to this node, just feels wrong. Let me know what you think.

**Comment by davidhalter** [comment]
> Just wondering whether this isn't something that should be fixed in the parser itself?

True. It can look like that. I have been there myself. At the same time there's a reason behind it. It's the way how the tokenizer handles whitespace/comments. I'm open for changes there, but I don't have an idea myself. If we're ok with the datastructure we could simply add a `get_suffix()` to be able to access the suffix.

**Comment by davidhalter** [comment]
Hmm, are you done here? I'm not sure what the state is and too lazy to review it in detail :)

**Comment by reinhrst** [comment]
Sorry, flu's put me offline for a couple of days. Nothing more happened here, just a rebase. First need to get tests running on 3.5 so that I can actually develop on it again :)

**Comment by reinhrst** [comment]
OK.... I just rebased my PR on the `python35` branch... Perhaps this was not the best idea (then again, perhaps all these python35 commits will disappear here once that branch get merged to linter). In any case, ignore all commits above :)

**Comment by reinhrst** [comment]
I have most of the stuff working. Not included in this PR:

- Python 2 backwards-compatibility [function annotation][1], which I feel is large enough to be put in its own PR:
```
def embezzle(self, account, funds=1000000, *fake_receipts):
    # type: (str, int, *str) -> None
    """Embezzle funds from account using fake receipts."""
    <code goes here>
```
- [Placeholder type comments][2], which I think will not work easily in jedi, since jedi is smart enough to spot the second assignment of a variable and then ignore the first. I'd be happy to not support this sort of hinting:
```
from typing import IO

stream = None  # type: IO[str]
[...]
stream = get_stream()  #  <--- should retain the type: IO[str] binding
```


[1]: https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code
[2]: https://www.python.org/dev/peps/pep-0484/#type-comments

davidhalter:
Yeah the latter would probably be hard. I agree that the former is worth a PR.

Do you think you're ready to merge?

reinhrst:
Yep go ahead!

davidhalter:
Awesome!

PR #665, opened 2015-12-28
Fix UnboundLocalError

messense:
Fix ``UnboundLocalError: local variable 'module_path' referenced before assignment`` when ``paths`` is empty.


messense:
Duplicated of #624

PR #663, opened 2015-12-28
PEP 484 typing library

reinhrst:
Since we got (limited) pep-0484 annotation support now in linter, let's take it a step further and look at the `typing` module understand. `typing` contains many parts, and looking at it prompted more questions than answers, so here an attempt to discuss them, and possibly get to a small plan to move forward

### scope

Intention here is to only look at the use of the `typing` module in already working type hints. For now this would be the function parameter and return value hints. Probably we'll get for free that it will also work in docstring hints, but not sure anyone will ever do that. So out of scope for this discussion is other PEP-0484 ways of annotating (through comments or stub files).

### typing module
The `typing` module is part of python 3.5+, and can be used on older versions through the pypi version (`typing`). So far only looked at the "simple" classes (e.g. `Union`), but there doesn't seem to be any magic there; with which I mean, something jedi can use to automagically infer what it should do. Perhaps I'm mistaken here, in that case, please put me straight.

### steps
First we have to figure out when we're actually dealing with the `typing` module. I recon that in python 3.5 it's easier to see if `import typing` is actually importing the standard typing module (or some custom module also named typing). In python <3.5, it will always be a custom module, and I think that jedi can/should only interpret type-hints using the `typing` module, if the module is imported toplevel, with the name `typing` (I may mess up some python-terms here, but I think it's clear what I mean). It also means that any such module will be interpreted that way. This means that as long as `typing` is imported in one of the ways below, we will recognise it, and assume it to be the (python 3.5-backported) `typing` module:

```
import typing
def test(x: typing.Any): ...
```

```
from typing import Any
def test(x: Any): ...
```

```
import typing as t
def test(x: t.Any):...
```

It means that not supported situations are where one installs the `typing` module under some other name or as part of a package (in these cases jedi will not realise that it should interpret the module as the `typing` module). It also means that if your do the following, you will probably get wrong behaviour:

```
[main.py]
import typing
def test(x: typing.Any): ...

[typing.py]
Any=str
```

I think the last case can be supported by jedi if we would really really want to, but I don't think it will be a major problem to have this specific case fail.

### Items within typing module
My current estimation of no "force" that will jedi-magically make the `typing` module work, probably means manually implementing each functionality manually. In the following list I try to determine what priority things should have. Probably this list is far from complete; I guess everything I leave off means no priority for now. Priorities are based in how useful I would expect them to be in the codebase of my current daytime project --- a very subjective measurement indeed, so feedback is appreciated. Note that priority does not imply order: low priority but easy to implement things may be picked up before high-priority ones.

- [`Any`][1]: low priority, not very useful for completions, however a case can be made that a function `def test(x: typing.Any): ...` should not have the type of `x` inferred from a later call to `test(1)`.
- [`Union`][2]: high priority
- `Optional`: `Optional[X] == Union[X, None]` high priority; even through for type-hinting, using `Optional[X]` is not very useful, it must be understood by jedi (probably to be just `X`)
- [`Tuple`][3]: mid priority. Is there support for tuple-types in jedi already? Two variants: `Tuple[X, Y, etc]` and `Tuple[X, ...]` meaning undefined length of type `X`.
- [`Callable`][4]: low priority; perhaps not a very good reason, but I use very few callbacks.
- [`Typevar`][5]: high priority; It seems to me that supporting these will be very useful for correctly inferring return-types for library functions. In addition, `Typevar("A", str, int)` gives type information for any parameters of that type. I do think that correctly implementing this could be hard without diving into the inner bowels of jedi...
- `SupportsInt`, `SupportsFloat`, `SupportsAbs`, `SupportsRound`, `Reversible`, `Container`: low priority, less useful for completions, since they all only define one (or 2) method(s), that should not be called directly, and not very useful in returntype from those functions.
- `Iterable`, `Iterator`: mid priority, help infer type of loops variables and `next()` calls
- `AbstractSet`, `MutableSet`: mid priority, help infer type of loops variables
- `Mapping`, `MutableMapping`, `Dict`:  high priority, help infer type of loops variables and `__getitem__` calls, however only for homogenous maps
- `Sequence`, `MutableSequence`, `List`: high priority: help infer type of loops variables and `__getitem__`
- `ByteString`: low priority
- `MappingView`, `KeysView`, `ItemsView`, `ValuesView`: mid priority, help infer type of loops variables
- `Generator`: low priority
- `io`: low priority
- `re`: low priority
- `NamedTuple`: low priority
- `cast`: low priority
- `get_type_hints`: irrelevant
- `no_type_check`, `no_type_check_decorator`: not supported for now

### plan of approach
Nothing firm yet, first wanted some feedback on this

[1]: https://docs.python.org/3/library/typing.html#typing.Any
[2]: https://docs.python.org/3/library/typing.html#typing.Union
[3]: https://docs.python.org/3/library/typing.html#typing.Tuple
[4]: https://docs.python.org/3/library/typing.html#typing.Callable
[5]: https://docs.python.org/3/library/typing.html#typing.TypeVar


reinhrst:
So, I didn't quite get it to work the way you suggested. Problem is that pep0484 expects types/classes, not instances in the annotations. Therefore things broke if I tried to feed it a `jedi_typing,Sequence(int)` object. I could probably hack it to accept instances as well, but it will be a hack (ie: string instances should be evaluated not used to mean "str").

So the current code is the result of a good night of hacking away. It works, but that's probably all that can be said for it, and the nice inheritance-plan won't work this way I fear.

davidhalter:
> So, I didn't quite get it to work the way you suggested.

Well you got it to work in a much better way. This way you can even keep the right `__init__` signatures.

> and the nice inheritance-plan won't work this way I fear.

True. This is an issue. But I think there is also a solution! (Hopefully I'm not giving you "bad" advice again, hehe)

Rewrite it to::


    from collections import abc

    def factory(typing_name, *index_types):

        class Sequence(abc.Sequence):
            def __getitem__(self) -> index_types[0]:
                pass

        class MutableSequence(Sequence, abc.MutableSequence):
            pass

        class List(MutableSequence, abc.List):
            pass
        
        dct = {
            'Sequence': Sequence,
            'MutableSequence': MutableSequence,
            'List': List,
        }
        # and so on, once Jedi understands `locals()`, replace with getattr(..)
        return dct[typing_name]

Do you understand what I'm doing? Instead of calling `result |= replacementclass.py__getitem__(index)` you can then just call::

    cls_name_wrapped = compiled.create(class_name)
    argument_array = FakeSequence(...)
    result |= execute_evaluted(factory_func, cls_name_wrapped, argument_array)`

I currently don't have time to write how I calculate `FakeSequence`, but I might write that down later. Do you think this could be a solution?

reinhrst:
Right, that seemed to work :)

As soon as I have multiple indices working, I'll take a swing at Union (and Optional) and Mapping (and views). I think at that time I'll call it quits for now on this module, it will do 95% of the stuff I use on a daily basis then. Maybe will even finish it tonight.

I would like your thoughts on 878c95f --- do you think this is a good idea? It may look nicer in the feedback in an editor, seeing that something is a `Sequence[int]` rather than just a `Sequence`. I can't seem to convince the testing system to test for `Sequence[int]()` though....

Also, I tried for a bit to get it working on python 2. I recon it should be possible to annotate functions with `typing.Sequence[int]` from within a docstring, and when pep0484-style comment hints are implemented, you're actually doing it without trying to mix two disjoint annotation systems :). On the other hand, I'm not very motivated to struggle with python2 unicode vs non unicode strings etc. What do you think?

reinhrst:
Ah, one thing about the tests: tests will fail when there is no typing module. I'm not sure how best to add it to the test requirements. A `pip install typing` should do the trick

reinhrst:
I think code is nicer now in that last one, but I had to drop a tests, something I don't like. However seems that allowing addition of strings to become forward references, although not explicitly disallowed in pep0484, is a really bad idea:
```
class Ameta(type):
    def __add__(cls, othercls):
        return cls

class A(metaclass=Ameta):
    pass

class AA:
    pass

def x(a: "A" + "A"):
    pass
```

tell me the type of `a` in that last function.... :)

reinhrst:
OK, let me know what you think. I wouldn't be surprised at all if you would find some ways to make it crash :). I should actually include a list somewhere with which pieces of the typing library are done now (and which are still to-do).

Anyways, there are still 2 issues, noted in the tests with TODO
- I haven't quite figured out how to differentiate between `Tuple[int, str]` and `Tuple[int, ...]`. Easiest would be to just remove the `...` from the call-list if the requested class is a `Tuple`, but if possible I would like to go without separate code in pep0484.py for Tuple.
- It seems that jedi has a hard time with tuple-assignments (not sure what it's called). `x=(1,"2"); y=x[0], z=x[1]` seems to go fine, and jedi knows that y = int and z = str, however `y, z = (1, "2")` only seems to allow for y's type to be inferred, not z's.

davidhalter:
VERY NICE :) Looks very good!

> I would like your thoughts on 878c95f --- do you think this is a good idea? It may look nicer in the feedback in an editor, seeing that something is a Sequence[int] rather than just a Sequence. I can't seem to convince the testing system to test for Sequence[int]() though....

I like the idea, but the execution causes an issue: You're basically refactoring a syntax tree. This means that feature refactorings might go wrong (or basically also something in the type inference logic that compares to that name. The tree should be imutable IMO). 

I'm not exactly sure if this would fix the issue, but you could theoretically just edit the description code of `api/classes.py`, this is definitely something that I would support!

Why is it not testable? Can you give me an example?

> I recon it should be possible to annotate functions with typing.Sequence[int] from within a docstring

Yeah, this should definitely be possible. (IMO) There's really nothing that speaks against it. If you don't test it now, I will later.

> On the other hand, I'm not very motivated to struggle with python2 unicode vs non unicode strings etc. What do you think?

Why would that be a problem?

> Ah, one thing about the tests: tests will fail when there is no typing module. I'm not sure how best to add it to the test requirements. A pip install typing should do the trick

True. We should just install it by default in the `.tox` config file. Maybe there's also an option in `py.test`, but I'm not sure.

> I'll take a swing at Union (and Optional)

Nice! For `TypeVar` I think the best solution is to register your own function `stdlib._implemtented`. For this normal Python function you can ignore argument clinic, just use a plain function without the decorator like this:

    def typing_TypeVar(evaluator, obj, arguments):
         return a set of the arguments somehow.

If this PR doesn't include this, nevermind :)

> Ah, one thing about the tests: tests will fail when there is no typing module. I'm not sure how best to add it to the test requirements. A pip install typing should do the trick

> I think code is nicer now in that last one, but I had to drop a tests, something I don't like.

Yeah, that's really bad. I think forward references should actually be working the way you wrote it before. I know that this breaks `List['foo']`, but this is something we should solve in a different way. IMO we should add a function in `jedi_typing.py` that looks like this::

```
def _forward_reference(typ):
    return typ
```

This function should be called by all `__getitem__` and the like. We can then intercept all calls to this function (in `stdlib.py`) and return a custom type. However, as you've probably seen this doesn't work right now, because you cannot get the context where the string was created. I've planned to refactor Jedi in a way so it can do that, but it will require a lot of time.

Therefore, for the moment I would prefer we would stick with the simpler version of just not understanding forward references, if they are nested. Just commented the tests that are failing like this `## int()`. IMO it's more important to create good code now instead of just creating features (and amassing technical debt).

Forward references  (in typing.py) are using the stack frame. As long as Jedi doesn't understand that, we just ignore this feature.

> tell me the type of a in that last function.... :)

Should be `AA` in Python (I'm not talking about Jedi...)?! It's not like anybody is using that, really. I just think that we should follow the way how Python works. Is this bad?

> I haven't quite figured out how to differentiate between Tuple[int, str] and Tuple[int, ...]. Easiest would be to just remove the ... from the call-list if the requested class is a Tuple, but if possible I would like to go without separate code in pep0484.py for Tuple.

Well, I guess this should work::


    class Tuple(Sequence, tuple):
        def __getitem__(self, index):
            typ = indextypes[index]
            if typ == ...:
                return indextypes[0]()
            else:
                return typ()


> It seems that jedi has a hard time with tuple-assignments (not sure what it's called). x=(1,"2"); y=x[0], z=x[1] seems to go fine, and jedi knows that y = int and z = str, however y, z = (1, "2") only seems to allow for y's type to be inferred, not z's.

Hmm... Ok. This shouldn't be the case. Do you have a test?

> OK, let me know what you think. I wouldn't be surprised at all if you would find some ways to make it crash :).

I will review the patch as soon I have more time. I probably need one oru two days.

reinhrst:
> I like the idea, but the execution causes an issue: You're basically refactoring a syntax tree. This means that feature refactorings might go wrong (or basically also something in the type inference logic that compares to that name. The tree should be imutable IMO).
Right, I was afraid of that. It did feel extremely hacky what I was trying to do :). I will revert and leave it for now. Will look at the description code; probably we'd need some `name_to_show`-field on `representation.Class` then, right?
> Why is it not testable? Can you give me an example?
Ah, it seems to have been something else, fixed it

As for python2: probably not too hard, guess I will have to dust off my python2/unicode whip. Just wanted to make sure you agree that it's needed

> For TypeVar I think the best solution is to register your own function stdlib._implemtented. For this normal Python function you can ignore argument clinic, just use a plain function without the decorator like this

I was not planning on including the `TypeVar` in the PR. It's useful if done right, but lets solve one thing at a time.

> Should be AA in Python (I'm not talking about Jedi...)?! It's not like anybody is using that, really. I just think that we should follow the way how Python works. Is this bad?
I'm not sure there is a way how Python works? It comes down to the precedence of the "forward reference resolver", and this is not defined in PEP-0484 I think.

> I think forward references should actually be working the way you wrote it before. I know that this breaks List['foo'], but this is something we should solve in a different way. IMO we should add a function in jedi_typing.py
Well, the forward reference resolver should be independent from `jedi_typing.py` IMO. It's a pep0484-syntax thing, not a typing-module thing. Other (non-0484) type hints should not be influenced by it, even if they use the `typing` module (technically; although I don't see anything breaking if it does). More importantly however, if someone brings his own typing-replacement module (one that jedi can understand), and uses it with pep0484-style-hints, FR should work.

That is of course the ideal situation, and we'll also have to consider implementation. Actually, the more I think about it, the more I'm convinced that a proper solution will remain hard until the previous issue, of FR-precedence, has been decided. I will focus on that a bit more.

> Well, I guess this should work (Tuple):
See #670

> Hmm... Ok. This shouldn't be the case. Do you have a test?

OK, seems to be slightly more complex than what I thought. Pehaps you could point me in the right direction:
```
x = 1, "b"
a, b = x
#? int()
a
#? str()
b
def tupleassign(x: typing.Tuple[int, str]):
    a, b = x[0], x[1]
    #? int()
    a
    #? str()
    b
    a, b = x
    #? int()
    a
    #? str()   <--- this one fails ( actual = set() )
    b
```
So there is something special going on in tuple assignment beyond a `__getitem__(1)` I guess....

davidhalter:
I've gotten a bit sick (fever) and I'm therefore probably not really able to fix the tuple issues this weekend. I will try to fix it as soon as possible.

reinhrst:
No worries, I'm away for most of the weekend myself. Let me know when it's done, and get well soon!

davidhalter:
Ok. I'm back, not fully functional, but working :) Lost 9 kg. It was quite intense, hehe.

> probably we'd need some name_to_show-field on representation.Class then, right?

Not even necessary. You can also just do it in the api logic itself. But I don't know how big the code would be. I'll let you decide.

> I'm not sure there is a way how Python works? It comes down to the precedence of the "forward reference resolver", and this is not defined in PEP-0484 I think.

You're right. But I would not go too far away from mypy's way of doing things. Mypy evals those strings (which means that all the operations (additions and so on) are performed. Of course it's 

> So there is something special going on in tuple assignment beyond a __getitem__(1) I guess....

Not really, it's just that if you call it like that, Jedi will basically call `py__iter__`. `__iter__` is something you have defined yourself:

    class Iterable(abc.Iterable):
        def __iter__(self):
            yield indextypes[0]()

And there you have it, it takes the first type and returns it. For everything after it there's nothing to unpack.

jedi/evaluate/iterable.py
```diff
@@ -443,37 +444,56 @@ def py__iter__types(evaluator, types, node=None):
     return unite(py__iter__(evaluator, types, node))
 
 
-def py__getitem__(evaluator, types, index, node):
+def py__getitem__(evaluator, types, trailer):
```
davidhalter: Just wondering: Why did you change the signature?

jedi/evaluate/pep0484.py
```diff
@@ -60,3 +81,81 @@ def follow_param(evaluator, param):
...
+    # actually the PEP-0484 typing module and not some other
+    if tree.is_node(node, "subscriptlist"):
+        nodes = node.children[::2]  # skip the commas
+    else:
+        nodes = [node]
+    del node
+
+    # hacked in Union and Optional, since it's hard to do nicely in parsed code
+    if typ.name.value == "Union":
+        return set().union(*[evaluator.eval_element(node) for node in nodes])
```
davidhalter: I think `iterable.unite` would be pretty nice for this. I'm using it quite often. I know its name kind might be confusing, if you have a name in mind, please tell me.

jedi/evaluate/pep0484.py
```diff
@@ -60,3 +81,81 @@ def follow_param(evaluator, param):
...
+    valid_classnames = set(child.name.value
+                           for child in function_body_nodes
+                           if isinstance(child, tree.Class))
+    if typ.name.value not in valid_classnames:
+        return None
+    compiled_classname = compiled.create(evaluator, typ.name.value)
+
+    result = set()
+    # don't know what the last parameter is for, this seems to work :)
+    args = FakeSequence(evaluator, nodes, "x-type")
```
davidhalter: Haha. The last parameter is the sequence type (`list` or `tuple`) :-) This is probably not very important for you, because you're not using functions on this sequence, just lookups.

reinhrst:
Glad to hear you're doing better and thanks for the feedback. I'm afraid this week (and possibly next as well) I have some other pressing tasks (something has to pay the bills ;)), will spend some time finishing and polishing this when I can find some!

jedi/evaluate/iterable.py
reinhrst: Ok, I hope I still fully recall... I went through a number of iterations here (not all of them well-informed), and by now it might or might not be necessary to have the changed signature.

I wanted the raw trailer. Note that `node` in the original signature was always the closing `]` (I think). My first reason for that was that I wanted the index before it went through `create_index_types` --- I seem to recall that in the end this might not have been necessary.
The other reason was that I wanted to feed the index-node (not just the index) to `get_types_for _typing_module`. This was to have somewhere to get the parent context from, when doing the forward reference resolving. Since later I moved the forward reference resolving back to parsing the annotations, this is not necessary anymore. However it may become necessary again if we revert that code change, and resolve FRs in `get_types_for _typing_module` once more.

reinhrst:
> Not really, it's just that if you call it like that, Jedi will basically call py__iter__. __iter__ is something you have defined yourself:
Ah, sure, that's wrong anyways. Before solving the tuple-case, how can I fix this one (it should return an undefined number of items of type `indextypes[0]`)?

```
class Iterable(abc.Iterable):
    def __iter__(self):
        while True:
            yield indextypes[0]()
```
Doesn't seem to do the trick. 

Neither does
```
    class Iterable(abc.Iterable):
        def __iter__(self):
            class Iter:
                def __next__(self):
                    return indextypes[0]()
            return Iter()
```

Both do the same as the current code, only returning the first item

```
class Iterable(abc.Iterable):
    def __iter__(self):
        yield indextypes[0]()
        yield indextypes[0]()
```
Correctly assigns the first two, no more.... (as expected)

>You're right. But I would not go too far away from mypy's way of doing things. Mypy evals those strings (which means that all the operations (additions and so on) are performed. Of course it's ok to do it like you did, but in the longterm I would probably like to switch, simply because it's more complicated the way we're doing it now. But let's leave it for now.

I will try to think a bit more about it. There is more different behaviour between the two approaches:

- Using a different typehinting scheme (i.e. docstring) with the `typing` module, I think the correct behaviour should be to not resolve FRs (which it doesn't in the current code, it did in the code where the FRs where fixed in the typing module). There is no reason to use FRs in docstrings. Then again, there is no reason not to support it, it will not break any other behaviour. So not much of an argument

- Using FR with one's own typing module should work, I think. However I'm really trying to see if any such typing module would just be a contrived example that nobody would actually ever make. So not too sure this is a good argument either.

So I tend to agree with you that perhaps the best course of action for now it to revert 10f5e1, since that code is easier, and actually seems to better follow the spirit (although not the letter) of pep0484.

In the coming weeks I should have some more time to spend on this, hopefully we can get this stuff to merge soon. Just out of curiosity, how mature do you think the linter-branch currently is? Do you feel it's on par with the current master/stable?

davidhalter:
> while True:
>             yield indextypes[0]()

> Doesn't seem to do the trick.

There's not really a good answer for that. The implementation of py__iter__ is basically not good enough. I would not care about this too much. It's something that I have to fix somewhere along the lines (it's kind of planned). Maybe comment the corresponding tests like this: `##?`.

> Using FR with one's own typing module should work, I think. However I'm really trying to see if any such typing module would just be a contrived example that nobody would actually ever make. So not too sure this is a good argument either.

> So I tend to agree with you that perhaps the best course of action for now it to revert 10f5e1, since that code is easier, and actually seems to better follow the spirit (although not the letter) of pep0484.

> In the coming weeks I should have some more time to spend on this, hopefully we can get this stuff to merge soon. Just out of curiosity, how mature do you think the linter-branch currently is? Do you feel it's on par with the current master/stable?

I think it's very mature and the linter branch basically is the new dev, so after this PR I'll probably merge with dev and then we should work towards master quickly.

reinhrst:
>>> I'm not exactly sure if this would fix the issue, but you could theoretically just edit the description code of api/classes.py, this is definitely something that I would support!

>>probably we'd need some name_to_show-field on representation.Class then, right?

>Not even necessary. You can also just do it in the api logic itself. But I don't know how big the code would be. I'll let you decide.

I'm looking into this (looking into how much work it would be). Would be an amazing feature to have included in the interface of any completer.

I'm afraid I'm not fully getting your line of reasoning. I have the data available when I'm in `pep0484.get_types_for_typing_module`, and I need it in the `api/classes`. Doesn't this mean that I should somehow store it in the meantime on the return value from `pep0484.get_types_for_typing_module`, which I think is a (set of) `representation.Class`?

davidhalter:
Good question. I'm actually not sure. I don't like "storing" additional data if we don't have to :) That's I guess why I suggested that. But you're probably right that we don't have access to that. Maybe we could implement a special function in `jedi_typing.py` that would return the internal types? So we don't have to modify anything when creating the classes?

reinhrst:
> Good question. I'm actually not sure. I don't like "storing" additional data if we don't have to :) That's I guess why I suggested that. But you're probably right that we don't have access to that. Maybe we could implement a special function in jedi_typing.py that would return the internal types? So we don't have to modify anything when creating the classes?

Gave it some thought over the last week. It would be ideal if somehow it could be deduced from the class again. And the more I thought about it, I see how this description does not belong on the  `representation.Class`. For now removed it.

I also fixed it to work in python 2.7. Didn't check 2.6, but I feel that 2.6 support has been discontinued in this branch, right? The only tests failing in 2.7 are the tests that fail in the upstream branch.

I feel that that's the last open issue... Ready to merge I'd say, or did I miss something?

davidhalter:
Hmm. I haven't actually discontinued 2.6 support. But it's ok if it passes one version. I can fix the errors then. Typically it's something like missing unicode calls and so on. 2.6/2.7 issues seem to be my fault anyway (in the linter branch). I will merge after I have done a small review again and since I'm on holidays, it might take a few days :)

reinhrst:
> Hmm. I haven't actually discontinued 2.6 support.

Ah, I was thinking this since the test complains that there is no grammar2.6.txt. However this is not there in master either, probably should use 2.7. Anyways, let me know when the tests run on the upstream, and I can check what is needed to get them to run in this PR.

>  I will merge after I have done a small review again and since I'm on holidays, it might take a few days :)

No worries, take your time and enjoy your holidays.

davidhalter:
Awesome. I reviewed it again. Thanks so much!

PR #676, opened 2016-01-09
Refactoring WIP

purpleP:


davidhalter:
Did you merge with linter? If so please create a pull request against linter. This PR is still not readable at all.

purpleP:
@davidhalter I don't remember, but you just need to see one file. refactoring.py. I don't see why generate pull request for that purpose in the first place. But I'll make another one.

PR #678, opened 2016-01-09
Refactoring WIP

purpleP:


jedi/common.py
```diff
@@ -155,6 +154,11 @@ def splitlines(string):
     return re.split('\n|\r\n', string)
 
 
+def content(file_path):
+    with open(file_path, 'r') as f:
+        return f.read()
```
davidhalter: You shouldn't even be reading files. This will cause you issues with unicode. There's quite a bit of logic within Jedi that cares for this stuff. There's probably a code attribute in all the parsers that you can use.

jedi/api/__init__.py
```diff
@@ -51,6 +53,8 @@ class NotFoundError(Exception):
     .. todo:: Remove!
     """
```
davidhalter: Remove please. You shouldn't modify this file at all.

jedi/refactoring.py
```diff
@@ -12,86 +12,146 @@
...
+    def real_line(self):
+        return self.line - 1
 
-    def old_files(self):
-        dct = {}
-        for old_path, (new_path, old_l, new_l) in self.change_dct.items():
-            dct[new_path] = '\n'.join(new_l)
-        return dct
+
+class PosRange(namedtuple('PRange', ['start', 'stop'])):
```
davidhalter: Since we're not using this in Jedi in general and `start <= x <= stop` is really good enough, please remove this.

jedi/refactoring.py
```diff
@@ -12,86 +12,146 @@
...
+        return script.module_path
 
-    def process(path, old_lines, new_lines):
-        if new_lines is not None:  # goto next file, save last
-            dct[path] = path, old_lines, new_lines
+    usages = (u for u in sorted(script.usages(), key=by_module_path)
+              if not u.in_builtin_module())
+    usages_by_file = groupby(usages, by_module_path)
+    return [change_for_rename(m_path, usages, new_name)
+            for m_path, usages in usages_by_file]
```
davidhalter: Hmm I think you haven't understood how the parser works, yet. What you should do is use the `usages` to modify the AST and then call `get_code()` on the AST. This is way simpler than what you did. Clearly this is also possible, but you make your life very complicated. Basically you don't need a Content and FileState class. It's way simpler. Do you understand what I mean?

Maybe just start with something simple: Parse something, change a Leaf.prefix and call `Parser.module.get_code()` and you will maybe see what I mean.

purpleP:
Oh, I see. I basically refactored what's already been there. Haven't thought about using parser to manipulate AST. I will play with parser and refactoring it then. And when I will be ready, I will implement 'refactoring' functions

davidhalter:
@purpleP I would be very happy to see this! I'm closing this PR, because it's not using the AST. However I would be really excited to see a new one coming!

purpleP:
@davidhalter Yeah, I don't have time right now, because I'm working on library that can create rest interface for some sqlalchemy models hierarchy, but as soon as I will somewhat finish it I will get back to refactoring. I really want to use Jedi for refactorings. By the way, One of the PyCharm weak points is when I want to rename argument. It renames its usages in function, but isn't smart enough to rename every occurance of passing this argument as kwarg. Will this be possible with Jedi?

davidhalter:
Can you give me an example? I'm not exactly sure what you mean.

purpleP:
suppose we have a function

``` python
def f(a, b):
   return a + и
```

and we use it somewhere.

`print(f(a=1, b=2))`


Now If I would perform rename refactoring in PyCharm to rename parameters a and b to с and d it will do the following.

```python
def f(c, d):
   return c + d

print(f(a=1, b=2))
```
Sorry, I've just retested this behaviour and I've been surprised, that PyCharm behaves as it should and renames

```python
print(f(a=1, b=2))
```
to
```python
print(f(c=1, d=2))
```
Very strange, because I remember that such things have given me some pain.
I've discovered also, that PyCharm can rename parameters in class hierarchy, which is pretty nice.

One thing that also disappointed me in PyCharm previously is when performing move refactoring of some package PyCharm sometimes (or maybe always, I don't remember) couldn't produce correct imports. I hope that we could make that possible with Jedi.

davidhalter:
I'm not sure if the case above is handled correctly, yet. Actually I'm pretty sure it's not. But it shouldn't be very hard to handle it correctly.

PR #680, opened 2016-01-11
Support type hinting using docstring of method interface (i.e. parent class)

emacsway-qarea:


davidhalter:
I don't understand, what does this solve? Also I will not accept PRs on master branch. Please try against the linter branch.

emacsway-qarea:
Example:

```python
    # Class Intrface
    class ISample(object):
        def a_method(self, a_arg):
            """:type a_arg: random.Random"""


    # Class realization
    class Sample(ISample):  
        def a_method(self, a_arg):
            a_arg.|no completion (fixed in fork)
```

davidhalter:
I won't merge this, sorry. There's already enough complexity in the docstring matching (and this goes over the top). There's no guarantee in Python that a sub class has the same signature.

emacsway-qarea:
The bad type hinting is better than nothing.

> There's no guarantee in Python that a sub class has the same signature.

First of all, in this case we can define new docstring with new types.

The second, - subclasses should follow interface.

Third, - it's a code smell and bad practice:

"Refused Bequest
The smell of refused bequest is much stronger if the subclass is reusing behavior but does not
want to support the interface of the superclass. We don't mind refusing implementations, **but
refusing interface gets us on our high horses**. In this case, however, don't fiddle with the
hierarchy; you want to gut it by applying Replace Inheritance with Delegation."

M.Fowler, "Refactoring".

> I won't merge this, sorry.

No problem.

davidhalter:
> The bad type hinting is better than nothing.

Maybe for completion but not for all the other usages of static analysis.

> The second, - subclasses should follow interface.

Of course, but this is not a must, I can do whatever I want with an interface in Python. Once one of the open PRs is merged you can just use type annotations (PEP 484) to be more specific about your types.

emacsway-qarea:
> Maybe for completion but not for all the other usages of static analysis.

I use type hinting for refactoring (every day), and it's working great, hundred percent hit!

> Once one of the open PRs is merged you can just use type annotations (PEP 484) to be more specific about your types.

I don't find how to use type comments of PEP-0484 for function params and rtypes. Only annotation, - but it's not acceptable for Python2.

Also type comments of PEP-0484 does not supports by PyCharm for Python2. So, it can't be used by whole team.

> Of course, but this is not a must

There is exists a simple solution. No problem to check, is this class really an interface, i.e. has type abc.ABCMeta.

Btw, python is polymorphic language, which means that the objects must follow interface. I can not imagine a case when a subclass changes the signature of superclass (except constructor). It's really messy approach. I don't know programmers who would do it.

I do not insist on resolving this issue. I've already resolved this issue privately for me by patched rope. I just suggested how we can improve jedi. There is no problem to decline this proposal. I agree.

m3nthal:
@emacsway-qarea, [PyCharm supports PEP-0484](https://www.python.org/dev/peps/pep-0484/). I use type hints in Python2 comments a lot because there is now other way to do it: `# type: (str, dict) -> [WebElement]` works in PyCharm.

davidhalter:
This has nothing to do with each other.

m3nthal:
@davidhalter I was replying to @emacsway-qarea comment:

> Also type comments of PEP-0484 does not supports by PyCharm for Python2. So, it can't be used by whole team.

Issue #372, opened 2014-01-24
No completions of the functions of SciPy, NumPy, Matplotlib, and PyLab


changyuheng:
This issue is very similar to [this](http://stackoverflow.com/questions/14708453/emacs-jedi-does-not-find-numpy-completions).

I'm currently using Vim with YouCompleteMe. I found that Python's code completion went well except of the follow modules' functions: SciPy, NumPy, Matplotlib, and PyLab.

I traced the YouCompleteMe, finding it just pass the code completion job to Jedi, so then gave a quick try:

``` python
import jedi
print(jedi.Script('import numpy; a = numpy.array([0]); a.').completions())
```

Got output: '[]'.

But

``` python
import jedi
print(jedi.Script('import numpy; numpy.').completions())
```

went well.

I also found that IPython completes both 'numpy.' and 'a.' well.

There is a test unit for NumPy in the file 'test/completion/thirdparty/pylab_.py' around the line containing 'na.shape', so I think the root cause may be the wrong configuration on my machine. But I cannot find it out from the documents.

I'm using Vim with +python flag and YouCompleteMe plug-in on Ubuntu 13.10. I also installed SciPy by apt-get:

``` sh
apt-get install python-numpy python-scipy python-matplotlb
```

And then upgrade them by pip (without virtualenv)

``` sh
pip install --upgrdae numpy scipy matplotlib
```

Knusper:
any updates on this?

davidhalter:
No.

ronxin:
@davidhalter Does jedi currently support extracting the return types specified in numpydoc-style docs? @immerrr mentions this at https://github.com/davidhalter/jedi/issues/372#issuecomment-50237535 , and you mentioned that this "is now a part of Jedi", but in https://github.com/davidhalter/jedi/blob/master/jedi/evaluate/docstrings.py I still only see extracting parameters out of numpydocs, and in testing jedi doesn't seem to be tracking the return types (e.g., of matplotlib). Am I missing something?

davidhalter:
@roxin Maybe it's working in the linter branch?

ronxin:
@davidhalter I didn't seem to find it in the linter branch either.

Gansgar:
Why are you just not running a python shell in the background. I know that sounds ambiguous, because there are many cases which could go wrong. But having instead a complete autocompletion is something I would really like. You could also give the option to activate / deactivate it or something like that.

bcolsen:
[...]

PR #682, opened 2016-01-12
fix exception throwing when dealing with numpy

21paradox:
Add Null check, Just passthrough empty result instread of throwing errors
This would fix https://github.com/davidhalter/jedi/issues/372 
Tested in smartos.

### DIFF
--- a/jedi/evaluate/finder.py
+++ b/jedi/evaluate/finder.py
@@ -82,6 +82,8 @@ class NameFinder(object):
 
     @debug.increase_indent
     def find(self, scopes, search_global=False):
+        if not scopes:
+            return []
         # TODO rename scopes to names_dicts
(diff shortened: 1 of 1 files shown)


davidhalter:
Why would scopes be None? And why would self.scope not have a names_dicts attribute? These are the really important questions here.

21paradox:
self.scope doesn't have names_dicts attr maybe because of c extensions.
I'm not 100% sure.
I just have tested one module  and numpy, they both not 100% written in python

davidhalter:
You have to get behind the reason why this would happen. I'm not going to merge assumptions.

21paradox:
Very sorry for that pr. I will take a deep looking at the problem.

Issue #122, opened 2013-02-06
Namespace packages are not supported


flupke:
Goto is tricked by namespace packages and fails.

I made tests to demonstrate this issue in https://github.com/flupke/jedi/commit/ad95279f1f5d160a22d5e07c1ced2a3011f39777

The problem is that `imp.find_module()` does not support namespace packages.

Should `imports.ImportPath._follow_file_system()` fiddle with `pkg_resources` (see http://stackoverflow.com/questions/5741362/alternatives-to-imp-find-module for a beginning of solution)? Or is `imp.find_module()` something you would like to discard in the future? (maybe by using `importlib.import_module()`, which does work with namespace packages, but has the disadvantage of actually importing the modules)

sherwoodwang:
I'm using latest jedi-vim. It seems completion for `zope.interface` doesn't work.

When I use \<lead>d to find definition, it even throws an exception.

    Traceback (most recent call last):
      File "/Users/username/.vim/bundle/jedi-vim/jedi/jedi/_compatibility.py", line 24, in find_module_py33
        loader = importlib.find_loader(string)
      File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/importlib/__init__.py", line 70, in find_loader
        raise ValueError('{}.__loader__ is None'.format(name))
    ValueError: zope.__loader__ is None
    During handling of the above exception, another exception occurred:
    Traceback (most recent call last):
      File "/Users/username/.vim/bundle/jedi-vim/jedi_vim.py", line 110, in wrapper
        return func(*args, **kwargs)
      File "/Users/username/.vim/bundle/jedi-vim/jedi_vim.py", line 211, in goto
        definitions = [x for x in script.goto_definitions()
      File "/Users/username/.vim/bundle/jedi-vim/jedi/jedi/api/__init__.py", line 364, in goto_definitions
        definitions = resolve_import_paths(definitions)
      File "/Users/username/.vim/bundle/jedi-vim/jedi/jedi/api/__init__.py", line 338, in resolve_import_paths
        scopes.update(resolve_import_paths(set(s.follow())))
      File "/Users/username/.vim/bundle/jedi-vim/jedi/jedi/evaluate/cache.py", line 41, in wrapper
        rv = function(obj, *args, **kwargs)
      File "/Users/username/.vim/bundle/jedi-vim/jedi/jedi/evaluate/imports.py", line 94, in follow
        types = importer.follow()
[...]

PR #689, opened 2016-01-26
fix for #122

sergeyglazyrindev:
Hey David
I use widely jedi on my emacs on python 3 and I persistently getting error reported in #122.
I think this could be at least a fix to improve our developer's life
What do you think ?


davidhalter:
Awesome, thank you!

PR #690, opened 2016-01-31
Read function annotations in python3

miyakogi:
Function annotations are not read in python3 environment.

When run the below code with python2, it shows lots of Completion objects, but with python3, no completions are shown. This PR fixes this.

```py
from jedi import Script

src = 'def a(i:int):\n    i.'
s = Script(src, 2, 6)
print(s.completions())
```

I have added `tname` in grammer for python3, and `tfpdef` is now juast an alias to it.
Although renaming all tfpdef to tname might be better, it makes differences between py2 and py3 grammer files bigger.


davidhalter:
I don't get at all what that should help with. There are no tests and function annotations are definitely read. At least in the latest linter branch.

miyakogi:
Sorry, I have missed linter branch, since it does not work with python3.5.
Now I tested in linter branch with python3.4, completions are correctly shown.

Sorry again for bothering you, I close this PR, thank you.

davidhalter:
No problem :) Glad you tried to debug it.

PR #691, opened 2016-02-09
python 3.5 support status

reinhrst:
In a mood of great optimism, I've switched all my projects to python 3.5 :). Hence this question...

I was wondering what the current status is for python3.5. Right now he linter branch crashes right away, but this is easy to fix (add the `@=` operator to the `parser/token.py` file, but then still there are a bunch of errors when running the tests. I think as a first step we should get the current tests running on python 3.5, only afterwards fix the `async/await` stuff (which shouldn't be to hard, since from a completer point of view, we can just ignore all `await/async` keywords :)).

Just wondering if you're already looking into this, or whether it would make sense for me to start squashing the errors one by one, until I have a working development environment again (and we have python3.5 support in jedi as added bonus :))?


reinhrst:
Passes all tests :)

Things that were solved but possibly in the wrong way, two `grammar.txt` changes, and one change in the tokenizer:

 - most things that got parsed as `power` nodes in 3.4, are now `atom_expr`. This resulted in pretty much me replacing every check for a `power` node with an or-clause that checked for `atom_expr`
- In 3.4, the `*` and `**` argument-parts where part of the `arglist`. In 3.5, they are `argument`s. Some fiddling was needed to support both.
- There is some magic in tokenizing `async/await` stuff: https://www.python.org/dev/peps/pep-0492/#transition-plan. I implemented this in `tokenize.py`, and it seems to be working. Would that be the spot to do that?

davidhalter:
> Finally, I suppose you'lll have to manually add python3.5 as test environment in travis?

Just add it to `.travis.yml`, I guess.

> This resulted in pretty much me replacing every check for a power node with an or-clause that checked for atom_expr

That sounds kind of bad. Not that I don't agree with that change, it just makes it more complicated. But I will have to review the patch anyway (hopefully in the next few days).

reinhrst:
> Just add it to .travis.yml, I guess.

Indeed.... How did I miss that file.... However it required some more tweaking (since python 3.5 doesn't seem to be in the default travis image), but it seems to work now.

> That sounds kind of bad

Agree, don't like it either. Alternatives that I could think of is always make it generate `power` nodes, even if they have only 1 child, add the `atom_expr` to the grammar for older python's (I like this a lot from a "makes-sense" point of view, since `atom_expr` is a much better name for the node than `power`, since it's not a power in 99.9% of the cases; however editing `grammar.txt` is in itself a terrible thing), or making some `_compatibility` variable that is either `'power'` or `'atom_expr'` depending on python version. btw, later I learned that `is_node()` can take multiple node types, if we decide to go this way, I'll fix that in places where now I made it an or-clause.

jedi/evaluate/analysis.py
```diff
@@ -174,7 +174,7 @@ def check_try_for_except(obj, exception):
     def check_hasattr(node, suite):
         try:
             assert suite.start_pos <= jedi_obj.start_pos < suite.end_pos
-            assert node.type == 'power'
+            assert node.type == 'power' or node.type == 'atom_expr'
```
davidhalter: Why not `in`?

davidhalter:
Looks really good!

However I'm very confused about Python itself:

Why is `async for p in [1]:` yielding a syntax error in Python 3.5? How would I need to write something that uses `async for` or `await` in any possible way? I'm confused, because I'm using samples from the PEP and they are not working, but it's clearly addressed towards Python 3.5.

reinhrst:
Not sure you're referring to that, but:

If you use async or await outside of a 'async def' block, they are just
normal identifiers. It's scary magic that did already byte me before.

    def a():
        await = 1

Is fine, while

    async def a():
        await = 1

Is a syntax error.....

reinhrst:
OK, back at computer so I could test something before I write it: 
`async for p in [1]:` is a syntax error, unless it's within a `async def NAME` block. The reason for this is that `async` and `await` are only keywords when they're within an `async def` block, else they're just identifiers (with the obvious exception of `async def NAME` itself. So:
```
def a():
    async for i in [1]: ...
```
is a syntax error, while 
```
async def a():
    async for i in [1]: ...
```
is allowed (probably will give you an error when run, because [1] is not an asynchronous iterator).

I wrote things in the jedi tokenizer such that
```
async def a():
    def b():
        async for i in [1]: ...
```
or
```
async def a():
    def b():
        return await foo()
```
is not allowed. On the other hand
```
async def a():
    b = lambda x: await foo()
```
will be allowed.

Actually I just checked, and python 3.5 does allow an `await` in a nested non-async function (at least it's not a syntax error), so I think Jedi should also parse that.

davidhalter:
Wow. Thats... complicated.

I think the checks if async/await are used within a function shouldn't be in the tokenizer, because we currently also don't do it with `continue`, `break, etc. This would need to be done in a second pass (which we currently are not doing, but it's not important for autocompletion).

This leaves us with the ugly distinction of keyword vs. normal variable. Why are we not parsing it normally and are not adding a keyword (+ mapping ASYNC/AWAIT to async/await)? I think because of ambiguities. But maybe that's still an option.

I also think it might be clever to just implement it the way it will be implemented in `3.7`. This means we would use `async` and `await` already as keywords. This would break autocompletion obviously in a few places. But if you're already using `3.5` I don't think that's a big issue since you're not going to write `async` and `await` a lot. The impact of such a change would be minor and our code would be way cleaner.

What do you think?

reinhrst:
> I think the checks if async/await are used within a function shouldn't be in the tokenizer
Actually I put it there since that is what is described in PEP 492: https://www.python.org/dev/peps/pep-0492/#transition-plan . Indeed it's different from pretty much any other keyword in the language. Changing this would mean changing `grammar.txt`

> I also think it might be clever to just implement it the way it will be implemented in 3.7

That's certainly an idea. Places where this might be an issue is in `asyncio.await()` (which is deprecated now, but I still had it in my code once or twice). It would probably be bad for any sort of linting if we do this.

davidhalter:
> It would probably be bad for any sort of linting if we do this.

True, but linting is not something that will be good in the next few months (I'm quite busy.) So I would just not care, really. If we really need it then, we could still grab your code?

And btw. I'm not complaining about your solution. It's just this PEP makes the tokenizer way more complicated and I dislike that.

reinhrst:
Hmmm, got it, makes sense. It would mean removing a bunch of tests (that were added in 2a691ee), and I think changes to `grammar.txt`. Do we already have an idea what the grammar.txt will look like for 3.7? Will it contain ASYNC / AWAIT tokens, or use the literals like with other keywords?

davidhalter:
In 3.7 `async` and `await` will be keywords. So we could just edit `grammar.py` a bit. We just have to make a comment there.

reinhrst:
OK, committed!

Do you have a suggestion for the power <--> atom_expr ?

davidhalter:
I don't have a solution for that. I think that just changing it to `in` is cleaner. But that's really a detail.

For certain stuff it would probably be possible to add methods to the parser.tree (that check for certain conditions). However this is not in the scope of this PR.

reinhrst:
>  I think that just changing it to in is cleaner. But that's really a detail.

True, but I didn't want to add any commits doing that if we would decide afterwards to `atom_expr` the thing some other way... Anyways, done now. I'd say its ready to be merged! Tests fail on 2.6 and 3.2, but they do the same on the linter branch, so I don't think this PR broke that.

Furthermore, travis very unhelpfully, now says that all tests are done in python 3.5. This is not the case, but you need to specify python3.5 as language, or travis won't install 3.5 on the test machine. The tests are still run in their respective python versions (e.g. see the skips vs the xfails in `test_pgen2.py` in python 3.4 vs 3.5 test, just what one would expect)

davidhalter:
> Furthermore, travis very unhelpfully, now says that all tests are done in python 3.5. 

Yeah. I think it looks pretty broken now when using Python (even if you don't set the language).

Thank you very much. It looks indeed good!

PR #696, opened 2016-02-16
update documentation

reinhrst:
Include python 3.5 support, and PEP0484 support


davidhalter:
Wow, I really totally forgot about that! Thanks for documenting it.

PR #699, opened 2016-02-29
add python 3.5 classifier to setup.py

mfussenegger:
Since 5a555076bbfe294efb40293d162ddd409b7f84e6 3.5 is listed as
supported python version in the docs


davidhalter:
True, thank you!

PR #701, opened 2016-03-03
Add vscode to editors using jedi

mfussenegger:
The python extension for Visual Studio Code uses Jedi.

As seen here:

    https://github.com/DonJayamanne/pythonVSCode/tree/master/pythonFiles/jedi


davidhalter:
Thanks :)

PR #705, opened 2016-03-10
Update unsupported features list with "implicit namespace packages"

tbradshaw:
I tried to blend into the writing style of the document, while still providing a link to the specific feature.  (Since "namespace packages" have been all over the place in Python over the years, hopefully referencing PEP 420 will prevent readers from getting confused about other solutions.)


davidhalter:
Thanks!

PR #711, opened 2016-03-31
Enabled python 3.5 support

WoLpH:
The tests weren't testing Python 3.5 yet, now they do :)

I didn't have any other Python 3 available on my system so couldn't use tox otherwise


davidhalter:
Thanks!

PR #719, opened 2016-04-15
Add gedit to supported editors list

isamert:
I made a little plugin for Gedit that uses jedi.

https://github.com/isamert/gedi


davidhalter:
Thanks mate :)

PR #721, opened 2016-04-29
Fix jedi-vim issue #385: extra '=' sign for params

StephenHesperus:
jedi-vim issue #385:
  https://github.com/davidhalter/jedi-vim/issues/385
  Completion of function arguments get an extra = sign

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -385,8 +385,9 @@ class Completion(BaseDefinition):
         if settings.add_dot_after_module:
             if isinstance(self._definition, tree.Module):
                 append += '.'
-        if isinstance(self._definition, tree.Param):
-            append += '='
+        # Remove those two lines should fix jedi-vim issue #385: https://github.com/davidhalter/jedi-vim/issues/385.
+        # if isinstance(self._definition, tree.Param):
+        #     append += '='
 
         name = str(self._name)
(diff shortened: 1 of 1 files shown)


davidhalter:
I appreciate the effort! While this is solving the problem in jedi-vim, this is clearly not solving the issue completely. We want the `=` to be there in some cases (when it's part of an argument). I'm pretty sure the tests would also be failing (if not for the issue on master). 

Also please read CONTRIBUTING.md, make pull requests to the dev branch.

StephenHesperus:
Thanks for the response! Sorry for the inconvenience. 

As of the '=' sign, I think it's very nice if it's there when we need it, but right now, I do appreciate it if it just gives me the parameter name correctly, I can add the '=' sign by myself when I need it :smile:

And thank you for writing jedi and jedi-vim as such awesome tools! :+1:

davidhalter:
> As of the '=' sign, I think it's very nice if it's there when we need it, but right now, I do appreciate it if it just gives me the parameter name correctly, I can add the '=' sign by myself when I need it :smile:

I can understand this, but this has worked once. It's a bug that is definitely fixable by checking if you're inside a function definition. I'm pretty sure it will not work as it once did, but it will be better.

Just check the tests, if those are almost passing you would be on a good way.

Issue #724, opened 2016-05-17
Do not init colorama on import. 


Carreau:
[This line](https://github.com/davidhalter/jedi/blob/6266678064cdae5fa9d71a2d8854aac7503f1db9/jedi/debug.py#L17) activate colorama on jedi import. 

Which lead to a rabbit hole of things which in then lead to `ipython file.py > otherfile` to crash while trying to write to stdout https://github.com/ipython/ipython/issues/9402

Would it be possible to not initialize colorama on import, that swap sys.stdout ?

Carreau:
PS: this happen only if stdout/err are not tty in which case colorama logic is to (not (not wrap)) the streams, and strip ansi sequences.

davidhalter:
This has been solved by #725, thanks @Carreau!

PR #725, opened 2016-05-17
Initialize colorama lazily

Carreau:
Will prevent colorama from wrapping stderr/out if not in debug mode,
which leads to some errors.

--- 

A bit of a hack, but does the job., and I suppose we can expect undefined behavior when debugging is enabled...

### DIFF
--- a/jedi/debug.py
+++ b/jedi/debug.py
@@ -4,6 +4,10 @@ import os
 import time
 
 def _lazy_colorama_init():
+    """lazily init colorama if necessary, not to screw up stdout is debug not enabled.
+
+    This version of the function does nothing.
+    """
     pass
 
 _inited=False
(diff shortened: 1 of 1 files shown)


Carreau:
would close #724.

davidhalter:
Thanks, I think that makes a lot of sense.

BTW: I'm working on completion for now, which I haven't been doing for over a year. I think the current changes that I'm doing will benefit ipython a lot.

Carreau:
Thanks !

We had some bug in the IPython completion layer that was matching IPython API and Jedi one, so had to revert for now. I should rework on completion soon.

davidhalter:
@Carreau: #740 will give you a new release with a hopefully much better interpreter completion!

PR #726, opened 2016-05-30
fix source_to_unicode py3 compatibility

nakamuray:
At original code, `source_to_unicode()` converting `source` from bytes to str using `str(source)`.
On python3, it generate `b'...'` (like `repr(source)`) which is not what we expect and following if condition always fail.
As a result, `source_to_unicode()` always fail to detect encoding and fallback to iso-8859-1.

It may cause this issue: davidhalter/jedi-vim#566

### DIFF
--- a/jedi/common.py
+++ b/jedi/common.py
@@ -127,7 +127,7 @@ def source_to_unicode(source, encoding=None):
             # UTF-8 byte-order mark
             return 'utf-8'
 
-        first_two_lines = re.match(r'(?:[^\n]*\n){0,2}', str(source)).group(0)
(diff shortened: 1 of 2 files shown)


davidhalter:
Wow, the last few months used to be a bit messy. I just didn't really look at this PR. Sorry!

It looks good and I see the problem. What do you think about not even converting to unicode?

Because just using byte regex works as well:

    re.match(br'(?:[^\n]*\n){0,2}', b'a\nb\nc\n').group(0)

I think this is cleaner and more performant, because we don't even need to cast. (Your test case obviously wouldn't change.)

davidhalter:
I'm merging, but only to change it afterwards to my proposed solution.

PR #728, opened 2016-06-04
Encoding isn't set for imported modules

sadovnychyi:
https://github.com/davidhalter/jedi/blob/master/jedi/evaluate/imports.py#L445

Here the forced encoding isn't applied and therefore we cannot correctly recognize non-ascii doc strings in imported modules. 

Can we pass the value presented in Script's `encoding` argument somehow? Basically every call to source_to_unicode should be with default set to argument passed in Script.

https://github.com/sadovnychyi/autocomplete-python/issues/210


sadovnychyi:
Updated. It wasn't tested so I haven't touched any.

And it seems like it's better to keep that BOM checking in the beginning, but I'm not sure.

davidhalter:
Thanks!

PR #734, opened 2016-06-19
Add vim editor plugin deoplete-jedi

kiryph:


davidhalter:
Can you please also add it to the documentation? Thanks!

PR #727, opened 2016-06-01
Fix failure with ``module_path``

thet:
Fix failure when ``module_path`` wasn't available in ``jedi.evaluate.imports.Importer._do_import``.

PR #736, opened 2016-06-27
module_path needs to be initialized for the None check

blueyed:
Fixes https://github.com/davidhalter/jedi-vim/issues/587.

### DIFF
--- a/jedi/evaluate/imports.py
+++ b/jedi/evaluate/imports.py
@@ -304,6 +304,7 @@ class Importer(object):
                 return []
             else:
                 debug.dbg('search_module %s in paths %s', module_name, paths)
+                module_path = None
                 for path in paths:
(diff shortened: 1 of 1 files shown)


davidhalter:
Hmm this has been an issue quite a few times before:

#727 #624, just to name two of them.

I have still not received a test for it. I'm actually wondering why that happens.

So unless there's no test I'm not willing to merge. Feel free to add one and I'll reopen :)

blueyed:
Fair enough, and sorry for not checking for duplicates.
In https://github.com/davidhalter/jedi-vim/issues/587#issue-162476876 there seems to be a manual test case using flask.

davidhalter:
Yes, I still have to make a real test case out of it. I haven't come across this error myself.

Issue #737, opened 2016-06-27
Fix tests (on Travis)


blueyed:
It looks like tests are failing currently (e.g. https://travis-ci.org/davidhalter/jedi/builds/140589895).
This should be fixed.
Any insights in this already?

davidhalter:
They tests fixed AFAIK (it was an issue with travis), but only in the dev branch.

The latest branch (`away`) will hopefully replace master very soon. It might take a week or two to even reach dev, but I'm confident that it will soon happen.

PR #738, opened 2016-06-28
Merge .travis.yml from away branch

blueyed:
Fixes https://github.com/davidhalter/jedi/issues/737.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: python
+python: 3.5
+sudo: false
 env:
     - TOXENV=py26
     - TOXENV=py27
     - TOXENV=py32
     - TOXENV=py33
     - TOXENV=py34
+    - TOXENV=py35
     - TOXENV=pypy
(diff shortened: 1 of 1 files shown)


davidhalter:
Did you cherrypick it?

Hmm it seems like that the INTERNALERROR still seems to stay, see https://travis-ci.org/davidhalter/jedi/jobs/140765185.

I wouldn't invest a lot of time into this anyway, I have a branch ready that I want to push to master as soon as Interpreter completion works. This probably means two weeks. It's been a while.

blueyed:
I've merge it manually, but does not appear to be enough.
"away" is failing anyway (I've thought it would pass): https://travis-ci.org/davidhalter/jedi/branches
Should be more like "dev" then probably, if the .travis.yml is the issue itself.

I'd like to get this fixed, since it's bad for PRs in general if the target branch fails by itself.

davidhalter:
Yes, I understand!

PR #743, opened 2016-07-11
Add support for nested namespaces

scribu:
Fix for the problem described here: https://github.com/davidhalter/jedi/pull/624#issuecomment-231817572


davidhalter:
Awesome! Thanks for getting into this :)

There's one little thing that I'm still not sure about: What about the comment you mentioned in #624? Should it be removed?

scribu:
I removed the outdate comment. However, I just noticed that my fix breaks a bunch of other tests. :(

davidhalter:
It's not easy, I know ;)

davidhalter:
I have fixed it by now. I figured that it's a pretty hard thing to do.

Your findings however helped me a lot. The comment you mentioned was key: I had to get the "parent" path of the module. That's at least how Python does it.

Thanks again for the test! I cherry-picked it.

scribu:
Happy to help. :)

Issue #730, opened 2016-06-07
Error when completing


nmiculinic:
Jedi throws error when completing following code

```
import tensorflow as tf
tf.<activate autocomplete>
```

I'm using python 3.5 and tensorflow 0.9.0rc0. Jedi version is 0.9.0. 
Error which I get is following:

```
Traceback (most recent call last):
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/_compatibility.py", line 24, in find_module_py33
    loader = importlib.find_loader(string)
  File "/usr/lib/python3.5/importlib/__init__.py", line 87, in find_loader
    raise ValueError('{}.__loader__ is None'.format(name))
ValueError: google.__loader__ is None
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/lpp/.atom/packages/autocomplete-python/lib/completion.py", line 340, in watch
    self._process_request(self._input.readline())
  File "/home/lpp/.atom/packages/autocomplete-python/lib/completion.py", line 331, in _process_request
    request.get('prefix', '')))
  File "/home/lpp/.atom/packages/autocomplete-python/lib/completion.py", line 142, in _serialize_completions
    description = completion.docstring()
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/api/classes.py", line 449, in docstring
    followed = self._follow_statements_imports()
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/api/classes.py", line 485, in _follow_statements_imports
    return i.follow()
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/evaluate/imports.py", line 94, in follow
    types = importer.follow()
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/evaluate/cache.py", line 41, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/home/lpp/.atom/packages/autocomplete-python/lib/jedi/evaluate/imports.py", line 252, in follow
    return self._do_import(self.import_path, self.sys_path_with_modifications())
[...]

Issue #627, opened 2015-10-01
Jedi could not find modules in eggs


immerrr:
I stumbled upon this several days ago when tried to look up a freshly installed module that came in a wheel package.  The problem seems to be that pre33-module finder uses `imp` module that does not support eggs:

https://github.com/davidhalter/jedi/blob/fbe26ab64a75753eb4f3baa268e8bfd3b4310a7c/jedi/_compatibility.py#L57

It could be improved by reimplementing module lookup mechanism using ``pkgutil.ImpLoader`` for regular modules and ``zipimport.zipimporter`` for eggs.  And during reimplementation we can make it work on custom lists, not necessarily `sys.path`, so the `sys.path` swap trick will become unnecessary, too.

davidhalter:
Yeah, should definitely be improved. Thanks for the report.

davidhalter:
How is that related to your latest pull request?

immerrr:
It is not related, at least directly. It will help to further separate `Evaluator.sys_path` and the `sys.path`  as you won't need to do the `sys.path, oldpath = sys_path, sys.path` trick. But other than that I think they are pretty independent.

davidhalter:
I was thinking about the egg-links.

immerrr:
egg-links to me really seem like a NIH version of .pth links implemented for "pkg_resources".

immerrr:
Ah, indeed, I have forgotten about that. Egg files are added to path with .pth files installed by setuptools/pip, so any improvement in module loading will be useless until .pth files are handled.

sadovnychyi:
@immerrr maybe you could try to test my PR?

https://github.com/davidhalter/jedi/pull/745

davidhalter:
This should have been fixed by #745. If it didn't get fixed, let me know!

PR #745, opened 2016-07-18
Try to correctly import ZIP and EGG files

sadovnychyi:
As discussed in https://github.com/davidhalter/jedi/issues/730.

As a workaround of reading source files inside of ZIP files we can try to use `get_source` method of the "loader". There's already "loader" implementation for py3, so only small changes there, but it would require more re-writing for py2. I've started it, but then moved to py3 and didn't come back to py2 yet.

I'm still not 100% sure that we can correctly understand everything inside of ZIP files which are packages (with `__init__` files) but single-file packages like `six.py` seems to work perfectly. 

Related [PEP 302](https://www.python.org/dev/peps/pep-0302/).


jedi/_compatibility.py
```diff
@@ -35,30 +56,63 @@ def find_module_py33(string, path=None):
...
-            module_file = open(loader.path, 'rb')
+            module_file = DummyFile(loader, string, loader.path)
         except AttributeError:
             module_path = string
             module_file = None
         finally:
             is_package = False
 
+    if hasattr(loader, 'archive'):
+      module_path = loader.archive
```
sadovnychyi: We need this for `parser.utils.load_parser` to correctly read the last mod time of the file.

jedi/evaluate/imports.py
```diff
@@ -337,12 +337,15 @@ def _do_import(self, import_path, sys_path):
...
+            if module_path.endswith(('.zip', '.egg')):
+                source = module_file.loader.get_source(module_name)
+            else:
+                module_path = get_init_path(module_path)
         elif module_file:
             source = module_file.read()
             module_file.close()
 
-        if module_file is None and not module_path.endswith('.py'):
+        if module_file is None and not module_path.endswith(('.py', '.zip', '.egg')):
```
sadovnychyi: I don't really like this special cases. Any ideas how to avoid them?

sadovnychyi:
As per tests, now it works in both py2 and py3. Still need to make sure that nothing is broken though, and code may need some refactoring. Especially the `DummyFile` part.

jedi/_compatibility.py
```diff
@@ -18,6 +19,19 @@
 is_py26 = not is_py3 and sys.version_info[1] < 7
 
 
+class DummyFile(object):
```
sadovnychyi: New instance of `DummyFile` is getting created on each import lookup, even if have done it before. Is it a problem?

davidhalter:
Would this PR affect #627 in any way? Because we can load zipped modules. Aren't eggs sometimes zipped?

jedi/evaluate/imports.py
davidhalter: Hmm, I fully agree. But nothing comes to my mind at the moment.

davidhalter:
Looks very good, thanks! For 2.6 you can easily just remove support for zip imports. I don't think this is an issue at all. What I want though is that it doesn't raise an exception. This should also be reflected in the tests that they just return no completions for zip imports.

BTW: The 2.7 tests have failed on travis, because of an issue of mine.

sadovnychyi:
2.6.9 seems to work if we change `get_filename` -> `_get_filename` and all other methods are there, despite what specified in the docs. So I see two options here:
  1. Check `is_py26` inside of `find_module_pre_py33` and return ImportError above all importer related code.
  2. Make it work with (latest?) version of py26 by forcing it to use `_get_filename` if we are at py26.

I've implemented first option with typical `ImportError` without trying any zipimports.

About EGGs – as I know – they are mostly zipped, and there's no big difference between `.egg` or `.zip` module so we can import both. I've manually tested it with `easy_install six` which installs egg and then trying Jedi – it's works. That was my original issue why I started this PR.

Let me know if you want me to refactor something here or add more tests.

jedi/_compatibility.py
davidhalter: No, that's fine. There's also import caching that won't let you import the same thing over and over again.

davidhalter:
It looks really good!

I think we should go with option 2: `Make it work with (latest?) version of py26 by forcing it to use _get_filename if we are at py26.`

Eggs are supported in Python 2.6 as far as I know and should therefore also be completable. What I don't get is why you say that there's only a `_get_filename`. I personally see in `2.6.9` and `2.6.1` that there's an `pkgutil.ImpLoader.get_filename`. Or are you talking about something else?

jedi/_compatibility.py
```diff
@@ -35,30 +48,67 @@ def find_module_py33(string, path=None):
...
+        result = imp.find_module(string, path)
+        if result is not None:
+            module_file, module_path, description = result
+            module_type = description[2]
+            return module_file, module_path, module_type is imp.PKG_DIRECTORY
+    except ImportError:
+        if is_py26:
+            raise
+    if path is None:
+        path = sys.path
```
davidhalter: I think that should actually be `path = evaluator.sys_path`. Or even better just give the command the sys path.

jedi/_compatibility.py
```diff
@@ -35,30 +48,67 @@ def find_module_py33(string, path=None):
...
+    for item in path:
+        loader = pkgutil.get_importer(item)
```
davidhalter: Doesn't this logic also cover the above cases? Or do you really need both of them? Isn't the normal import logic a part of the PEP 302 import hooks?

sadovnychyi:
For zip-packages it returns an instance of `zipimport.zipimporter` and it does not have `get_filename` attribute. And according to the docs `get_source` shouldn't (but it is) be there.

```python
Python 2.6.9 (unknown, Jun 26 2016, 00:32:33) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.31)] on darwin
>>> import zipimport
>>> dir(zipimport.zipimporter)
['__class__', '__delattr__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '_files', '_get_filename', 'archive', 'find_module', 'get_code', 'get_data', 'get_source', 'is_package', 'load_module', 'prefix']
```

jedi/_compatibility.py
sadovnychyi: Yeah, we can remove it and everything *should* work. I thought to leave it there to be safe.

jedi/_compatibility.py
sadovnychyi: What do you mean by "give the command the sys path"?

jedi/_compatibility.py
sadovnychyi: Looks like some completions are missing after I removed it (as per travis). Ideas?

davidhalter:
Just checked `2.6.1`. It doesn't have `zipimport.zipimporter._get_filename`. Can we just ignore it if it's not there and infer imports if they are there?

jedi/_compatibility.py
davidhalter: You don't need the evaluator object. You just need to pass down `evaluator.sys_path` as a parameter.

davidhalter: I just realized that this is not necessary after all, because we overwrite `sy.path` before anyways. You can just ignore my comment here.

sadovnychyi:
Well, if you don't mind double try/except with continue – then yeah. Updated.

Still need to figure out whether we should keep that old block of code with imp or not. Some tests seems to fail without it.

davidhalter:
Hmm. I don't know why it's not working. If `pkgutil.get_importer` is not working the way we expect it, that's fine for now. Just stick with the old solution. You may as well debug it, but it's not that important (or can be done later).

sadovnychyi:
Reverted to use imp. `pkgutil` does have a wrapper around `imp` called `ImpLoader` but it does not work for ZIP/EGGs anyway.

davidhalter:
Awesome, thanks so much!

PR #746, opened 2016-07-19
Added CWD of vim to sys.path for completion modules of opened project.

ANtlord:
This commit added CWD of Vim to sys.path. It allows complete modules of current opened project.


davidhalter:
I'm against this. It's the editors job to provide a good sys.path.

PR #751, opened 2016-07-31
Fixed Windows slashes problem.

bstaint:
Fixed Windows slashes problem.

### DIFF
--- a/jedi/evaluate/compiled/__init__.py
+++ b/jedi/evaluate/compiled/__init__.py
@@ -343,7 +343,9 @@ def dotted_from_fs_path(fs_path, sys_path):
     # - Window
     # X:\path\to\lib-dynload/datetime.pyd => datetime
     module_path = fs_path[len(path):].lstrip(os.path.sep).lstrip('/')
-    return _path_re.sub('', module_path).replace(os.path.sep, '.')
+    # - Window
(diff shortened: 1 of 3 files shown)


jedi/evaluate/compiled/__init__.py
```diff
@@ -339,7 +339,11 @@ def dotted_from_fs_path(fs_path, sys_path):
     for s in sys_path:
         if (fs_path.startswith(s) and len(path) < len(s)):
             path = s
-    return _path_re.sub('', fs_path[len(path):].lstrip(os.path.sep)).replace(os.path.sep, '.')
+
+    # - Window
+    # X:\path\to\lib-dynload/datetime.pyd => datetime
+    module_path = fs_path[len(path):].lstrip(os.path.sep).lstrip('/')
```
davidhalter: I think it should be:

    return _path_re.sub('', module_path).replace(os.path.sep, '.').replace('/', '.')

Because what if there are multiple `slashes, like `X:\path\to\something/foo/bar.py`?

bstaint: imp.find_module will only return one slashes before filename.

davidhalter: And how do you know that?

davidhalter:
Thanks man!

PR #763, opened 2016-08-18
allow access to jedi.evaluate.representation.Instance.names_dict

mfripp:
A user is unable to run code like `n = jedi.names("lst = list()")[0].defined_names()` (see https://stackoverflow.com/questions/37686311/python-jedi-how-to-retrieve-methods-of-instances). This seems to be because `jedi.evaluate.representation.Instance.__getattr__()` blocks evaluation of `.names_dict`. This change allows the code to run.

### DIFF
--- a/jedi/evaluate/representation.py
+++ b/jedi/evaluate/representation.py
@@ -244,7 +244,7 @@ class Instance(use_metaclass(CachedMetaClass, Executed)):
 
     def __getattr__(self, name):
         if name not in ['start_pos', 'end_pos', 'get_imports', 'type',
-                        'doc', 'raw_doc']:
+                        'doc', 'raw_doc', 'names_dict']:
             raise AttributeError("Instance %s: Don't touch this (%s)!"
                                  % (self, name))
         return getattr(self.base, name)
--- a/test/test_evaluate/test_representation.py
+++ b/test/test_evaluate/test_representation.py
@@ -1,6 +1,6 @@
 from textwrap import dedent
 
-from jedi import Script
+from jedi import Script, names
 
 
 def get_definition_and_evaluator(source):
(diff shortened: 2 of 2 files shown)


denfromufa:
Can you add a test for this bug?

mfripp:
I've added a test for this bug fix. However, I'm not sure if this is considered a "correct" use of Jedi (i.e., a real bug, vs. misuse of the library). In particular, I can't find an equally simple fix to make this code work: `jedi.names('x = []')[0].defined_names()`  (assigning `[]` to `x` instead of `list()`), because jedi.classes.iterable.Array doesn't have a `.names_dict` attribute (it has a `names_dicts()` method, but I can't see how to use that or think of a clean way to access it from `jedi.api.classes.Definition.defined_names()`.

Yevgnen:
Any news ?

davidhalter:
Names_dicts are getting removed by the latest changes. We'll see if this is even an issue by then.

davidhalter:
Can you describe your use case a bit better? I'd be happy to add this functionality (I'm the creator of Jedi). Do you want to list all attributes of an instance (including super classes) or do you want a list of all the attributes of an instance that are defined in just that class?

In general it's very unclear what we want to do with defined_names. Currently in case of a module it only lists names that are defined in that modules. However there are other attributes that are also available in a module (like `__name__`, `__file__`). Would you guys want to see those attributes as well? In case of instances it's even more questionable what you would want. I'm also open to add a parameter for that.

davidhalter:
I'd happily accept a pull request for this, but since names_dicts doesn't exist anymore and not all questions have been cleared, I'm closing. Still, I think we should be able to add something like this. Just not sure how we define it.

PR #780, opened 2016-09-23
Makes CallSignature.call_signature()

Alexey-T:
It is same result as in docstring, ie method .doc, ie property. So it is with "self" (for methods) -i dont like it, but no other way, easy.

### DIFF
--- a/test/test_api/test_call_signatures.py
+++ b/test/test_api/test_call_signatures.py
@@ -27,7 +27,10 @@ def assert_signature_string(source, line, column, expected_string):
     assert len(signatures) <= 1
 
     if signatures:
-        assert signatures[0].call_signature == expected_string
+        assert signatures[0].call_signature(4000) == expected_string
+    else:
+        assert False, \
+            'There are no signatures, but `%s` expected.' % expected_string
 
 
 class TestCallSignatures(TestCase):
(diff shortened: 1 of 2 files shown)


davidhalter:
Tests? I'm not sure about the naming yet, but agree that we should have something like this.

Alexey-T:
Tests... well, tested on working Cudatext plugin. not sure how to make tests

Alexey-T:
Added one test

Alexey-T:
Was wrong text, now its OK

test/test_api/test_call_signatures.py
```diff
@@ -21,6 +21,15 @@ def assert_signature(source, expected_name, expected_index=0, line=None, column=
         return signatures[0]
 
 
+def assert_signature_string(source, line, column, expected_string):
+    signatures = Script(source, line, column).call_signatures()
+
+    assert len(signatures) <= 1
+
+    if signatures:
+        assert signatures[0].call_signature == expected_string
```
davidhalter: There should be an 

    else:
        assert expected_string is None

there.

The test below is probably running into this.

Alexey-T:
So the test is fixed, you agree?

Alexey-T:
it is fixed.

davidhalter:
Sorry for taking so long on this, but I didn't like what we came up with as the API together. It took a long time to develop another one, because the old internals didn't support what we were doing at the time.

Now there's `[signature.to_string() for signature in definition.get_signatures()]. Thanks for your patience.

PR #782, opened 2016-10-02
support for nested namespace packages

### LINKED ISSUE #122: Namespace packages are not supported

sbidoul:
fixes #122

### DIFF
--- a/jedi/evaluate/representation.py
+++ b/jedi/evaluate/representation.py
@@ -876,24 +876,39 @@ class ModuleWrapper(use_metaclass(CachedMetaClass, tree.Module, Wrapper)):
         is a list of paths (strings).
         Raises an AttributeError if the module is not a package.
         """
+        def ns_path(name, init_path):
+            res = []
+            while True:
+                if os.path.isfile(init_path):
+                    with open(init_path, 'rb') as f:
+                        content = common.source_to_unicode(f.read())
+                        options = ('declare_namespace(__name__)', 'extend_path(__path__')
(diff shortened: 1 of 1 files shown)


gustavovalverde:
Hi @sbidoul, as you've been actively working on this matters, including your commit on Odoo https://github.com/odoo/odoo/commit/8bf06adc8d78c1bff8dc80ce53d6646e3ad22fff#diff-1c2d6cde5b9dec890dc33408b4e82005, have you managed to make Jedi work with Odoo?

I've landed this PR because I'm using VS Code, but there's been no way to make Intellisense find methods inside odoo/addons.

sbidoul:
@gustavovalverde I've not pursued this work on jedi. On Odoo 10+ where addons are pip installed and in the odoo.addons namespace, jedi should work out of the box. It should be even better on 11 with python3 namespaces.

PR #783, opened 2016-10-02
sometime namespace packages got no parents

sbidoul:
Hi,

I got a strange situation with nested namespace packages. say I have the following modules:
- ns1.ns2.mod1
- ns1.ns2.mod2 in another path

The first time I do a <leader>d on ns1.ns2.mod1 it works.
The second time I do a <leader>d on ns1.ns2.mod2 it fails because the ns2 has no parent_module.

This PR resolves this although I have no real understanding on the root cause. 

refs #122


davidhalter:
> This PR resolves this although I have no real understanding on the root cause. 

This is what you need to fix it. It's not worth anything without a test. It might have been fixed in the dev branch also.

sbidoul:
Closing this for now as the fix is incorrect. It looks likes this has to do with caching, perhaps related to #784

PR #787, opened 2016-10-14
sys_path: prepend/prefer egg-link files

blueyed:
With `pip install -e` the generated .egg-link file gets preferred over
any normally installed distribution, and `pip uninstall` will first
remove the egg-link before the normal package.

I am not really confident if this is correct, but it just helped with resolving `src/django-filter/django_filters/filters.py` to the version installed with `pip install -e`.


davidhalter:
Totally fine with this. I still have only understood eggs partially, so happy if you care for them. :)

PR #788, opened 2016-10-14
tox: use posargs and test by default for py.test command

blueyed:
Otherwise it will collect tests from other dirs also by default.


davidhalter:
This would exclude doctests, though? Right?

blueyed:
Right, amended.

Issue #761, opened 2016-08-16
Regression with goto_definitions going to / staying at the child class (bisected)


blueyed:
On "dev" and "master" jedi-vim fails to go to the definition of `RequestFactory` when trying to go there from line 5 (cursor being on `RequestFactory`, using `jedi#goto()` ([code](https://github.com/davidhalter/jedi-vim/blob/99132ec54560de15f84f656539dad3094865b222/jedi_vim.py#L233-L237)).

I've bisected this in Jedi to the following commit:

27f05de3b7c64e06814602589455657b6028c77f is the first bad commit
commit 27f05de3b7c64e06814602589455657b6028c77f
Author: Dave Halter davidhalter88@gmail.com
Date:   Mon Jun 13 18:21:17 2016 +0200

```
Fix a few more issues, mostly with the fast parser.
```

:040000 040000 de6f62be0a20a24b26e67b0243debfa7659832bc 7b30878d3e423b9a3fa3ebb963238e391aa1767c M    jedi
:040000 040000 461af1f7c2d08d9217665774c4897f26bccf805c dd0a7bc10ece812044f77d5492854d16fccdb7cc M    test

``` python
class RequestFactory(object):
    pass


class Client(RequestFactory):
    pass
```

I hope this is reproducible, otherwise I can provide more information and/or a test case.

PR #790, opened 2016-10-14
Fix goto_definitions being invoked on a parent class

blueyed:
When invoking `goto_definitions` on `RequestFactory` in line 5, it would
jump to `Client` after 27f05de:

``` python
class RequestFactory(object):
    pass

class Client(RequestFactory):
    pass
```

Fixes https://github.com/davidhalter/jedi/issues/761.

@davidhalter 
This needs a test, and I am not sure if this fix is correct and makes sense.
Please advice me where and how to add it.

### DIFF
--- a/jedi/evaluate/__init__.py
+++ b/jedi/evaluate/__init__.py
@@ -446,7 +446,9 @@ class Evaluator(object):
         def_ = name.get_definition()
         is_simple_name = name.parent.type not in ('power', 'trailer')
         if is_simple_name:
-            if name.parent.type in ('file_input', 'classdef', 'funcdef'):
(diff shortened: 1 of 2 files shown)


davidhalter:
The fix looks ok. (Well it doesn't, but that's my fault, because it's my spaghetti code. There's a lot of improvement and better abstractions coming to the evaluation so that other people than me can actually work on it).

Just add a test and I'm very happy with it! As long as all tests are passing ;)

blueyed:
Where should the test go?

davidhalter:
I think `test/completion/classes.py`.

blueyed:
Added the test.

PR #799, opened 2016-11-08
fix source code read issue.

Cologler:
fix the source file read mode issue.
if use 'r' mode, this may raise a `UnicodeDecodeError`.


davidhalter:
Thanks!

PR #807, opened 2016-11-15
docstrings: fix "Sphinx param with type" pattern

### LINKED ISSUE #806: Sphinx "param with type" regex requires description to appear on the same line
I like to format my docstrings with the parameter descriptions on a new line, like so:

    :param type foo:
        A parameter named foo.

The Sphinx documention implies that this is valid, as in [this example](http://www.sphinx-doc.org/en/1.4.8/domains.html#directive-js:function).

However, the [regex](https://github.com/davidhalter/jedi/blob/be2a97cd36ce1b7d6f587ef122efc3433b1e8ae2/jedi/evaluate/docstrings.py#L31) that Jedi uses to parse such docstrings ends with `[^\n]+`, which requires there to be characters after the closing colon on the first line, or it won't match. It looks like this may not have been intentional: it could just be from copying the same fragment from the two nearby regular expressions, where the `+` is necessary to ensure that the regular expression won't match an with an empty capture. In this case, though, a `*` looks like it would be more appropriate.

sruggier:
Previously, the pattern only matched if the parameter description
followed on the same line, like so:

    :param type foo: A param named foo.

However, it's also valid for the parameter description to be wrapped
onto the next line, like so:

    :param type foo:
        A param named foo.

This change updates the pattern to match the second example as well.

Fixes #806.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -42,6 +42,7 @@ Cristi Burcă (@scribu)
 bstaint (@bstaint)
 Mathias Rav (@Mortal) <rav@cs.au.dk>
 Daniel Fiterman (@dfit99) <fitermandaniel2@gmail.com>
+Simon Ruggier (@sruggier)
 
 
 Note: (@user) means a github user name.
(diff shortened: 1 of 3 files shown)


sruggier:
I'm going to update this to follow the contributing guidelines in a couple of minutes.

sruggier:
Ok, I've rebased on dev, added a test, and added myself to the AUTHORS file.

davidhalter:
Sorry I had actually reviewed this, but not submitted the review. Therefore you probably didn't/couldn't see it. There's just this one minor issue. :)

Sorry again.

sruggier:
Thanks! I've moved/rewritten the test into `test/completion/docstring.py`.

PR #796, opened 2016-11-07
Added google style docstr support + numpy docstr fixes

Erotemic:
The purpose of this PR is to address issue #504 properly and improve upon the original attempt submitted in PR #617. It took a bit more time than expected, but I think I have everything working and tested. 

In addition to enhancing jedi with ability to  and return values from google style docstrings, I also included several fixes for numpy style docstrings. 

A summary of what is changed is as follows:

    * [ENH] jedi can now extract type hints for parameters and return values from google style docstrings
    * [ENH] added tests for new google style parsing
    * [ENH] added doctest examples to several functions in jedi/evaluate/docstrings.py
    * [FIX] numpy return type hints are now correctly parsed and used
    * [FIX] jedi can correctly extracts type hints for alternative-type parameters from numpy docstrings
    * [FIX] fixed an exception in python2.7 for numpy set-of-values parameters. 

----

Here is a breif description of how the new google style parsing was implemented: 

To add support for the google style docstrings I attempted to mirror what was happening for numpy style docstrings. I created a file called `jedi/evaluate/docscrape_google.py` to work in a simlar way to `numpydoc.docscrape`. The main work in this file is done by  `split_google_docblocks` which separates the docstring into a list of sections represented as a list of tuples. Each tuple is a pair of section names and unparsed section lines. This split is made primarily using indentation information and a list of known google-style tags. Header aliases are taken into account so Arguments and Args section headers all resolve back to Args in the returned list. 

The publicly facing functions are `parse_google_args` and `parse_google_returns` which get the appropriate Args / Returns / Yields section and then calls either `parse_google_retblock` or `parse_google_argblock` to grab a list of type hints represented as dictionaries. Back in `jedi/evaluate/docstrings.py` I use this new functionality to augment the behavior or `find_return_types` and  `_search_param_in_docstr`. These functions will now try to use google style docstrings if parsing Epydoc/Sphinx docstrings fails. 

When first looking into `jedi/evaluate/docstrings.py` I was having a hard time getting a feel for the information flow of the module. My style of coding is very integrated with an IPython terminal. I ended up creating several example doctests just as a byproduct of needing to define valid input arguments in IPython. 

----

In regards to the numpy docstring changes: 

[...]

PR #818, opened 2016-12-17
PR: Return types for CompiledObjects using docstrings.py

bcolsen:
This pull request in conjunction with PR: #796 for numpydoc returns in docstrings solves issue #372 for numpy complied objects. This makes completions functions numpy.array to return an numpy.ndarray object resulting in returns like this:

    >>> import jedi; jedi.Script('import numpy as np; np.array([1,2,3]).a').completions()
    [<Completion: all>,
     <Completion: any>,
     <Completion: argmax>,
     <Completion: argmin>,
     <Completion: argpartition>,
     <Completion: argsort>,
     <Completion: astype>]

I have not yet written a test since I need to import numpy to do such. A standard python library with complied objects and sphinx documentation would be a better test.

### DIFF
--- a/jedi/evaluate/compiled/__init__.py
+++ b/jedi/evaluate/compiled/__init__.py
@@ -78,6 +78,13 @@ class CompiledObject(Base):
     @property
     def doc(self):
         return inspect.getdoc(self.obj) or ''
+        
+    @property
+    def raw_doc(self):
+        try:
+            return unicode(self.doc)
+        except NameError: # python 3
+            return self.doc
 
     @property
     def params(self):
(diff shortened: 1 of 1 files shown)


jedi/evaluate/compiled/__init__.py
```diff
@@ -196,9 +203,12 @@ def name(self):
         return FakeName(name, self)
 
     def _execute_function(self, params):
+        from jedi.evaluate import docstrings
```
bcolsen: This is necessary due to cyclical importing of CompliedObject in analysis.py which is import by docstrings.py through iteratable.py

ccordoba12:
@bcolsen, could you update #796 and this one to work with the new 0.10.0?

bcolsen:
@ccordoba12, I can look at getting the spyder jedi patch to work for both Jedi 0.9.0 and 0.10.0.

If needed, I could work on #796 but I'm not familiar with the googledocs stuff.  I'm also going to wait on this PR until #796 goes through.

ccordoba12:
> @ccordoba12, I can look at getting the spyder jedi patch to work for both Jedi 0.9.0 and 0.10.0

Yes, that'd be great, thanks!

bcolsen:
Closing this because I submitted PR: #868 that includes an updated version of this code.

PR #823, opened 2016-12-28
parser: add grammar3.6

### LINKED ISSUE #820: 3.6 Grammar Is Missing
When invoking Jedi via API under Python 3.6, an explicit request for `grammar3.6.txt` is made:
```
 File "preview\jedi\api\__init__.py", line 130, in __init__
    self._grammar = load_grammar(version='%s.%s' % sys.version_info[:2])
```
Since there's no such file, Jedi crashes, and autocompletion in the editor breaks.

fsouza:
Fix #820.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,25 @@
 language: python
-python: 3.5
 sudo: false
-env:
-    - TOXENV=py26
-    - TOXENV=py27
-    - TOXENV=py33
-    - TOXENV=py34
-    - TOXENV=py35
-    - TOXENV=pypy
-    - TOXENV=cov
-    - TOXENV=sith
+python:
+  - 2.6
+  - 2.7
+  - 3.3
+  - 3.4
+  - 3.5
+  - 3.6
+  - pypy
 matrix:
-    allow_failures:
-        - env: TOXENV=cov
-        - env: TOXENV=sith
-        - env: TOXENV=pypy
-python: 3.5
+  allow_failures:
+    - python: pypy
+    - env: TOXENV=cov
+    - env: TOXENV=sith
+  include:
+    - python: 3.5
+      env: TOXENV=cov
+    - python: 3.5
+      env: TOXENV=sith
 install:
-    - pip install --quiet tox
+    - pip install --quiet tox-travis
 script:
     - tox
 after_script:
--- a/jedi/parser/grammar3.6.txt
+++ b/jedi/parser/grammar3.6.txt
@@ -1,0 +1,161 @@
+# Grammar for Python
+
+# Note:  Changing the grammar specified in this file will most likely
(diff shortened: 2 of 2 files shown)


davidhalter:
Please add Python 3.6 to the tested versions in travis. otherwise this doesn't help us very much (it's not tested)

 https://docs.travis-ci.com/user/languages/python/

fsouza:
I tried to work on that and it looks like tox is not straightforward. I'm going to figure this out later today, but any special reason for not using Travis's standard multi-version approach?

While trying to set it up, I also noticed that the tests are failing on Python 3.6, so I'm going to take a look at that too.

fsouza:
Hey @davidhalter, I applied the needed changes in the grammar, so the grammar-related tests now pass. I also took a totally "ungood" path with travis: since I couldn't have Python3.5 on Python 3.6 box, nor Python 3.6 on Python 3.5 box, I'm assembling the build matrix manually, which sucks.

Back to the tests, it looks like there were some changes in Python internals and some pep0484_typing related tests are failing (when using ``typing.Union`` in the docstring).

I'm not entirely familiar with that, but I'm willing to give it a try. Anyway, if you have any suggestions, I'd be happy to accept them :)

.travis.yml
```diff
@@ -7,15 +7,21 @@ env:
...
+    - TOXENV=py36
     - TOXENV=pypy
     - TOXENV=cov
     - TOXENV=sith
 matrix:
-    allow_failures:
-        - env: TOXENV=cov
-        - env: TOXENV=sith
-        - env: TOXENV=pypy
-python: 3.5
```
davidhalter: Why did you change this and the include/exclude?

fsouza: Because "python: 3.5" && TOXENV=py36 fails with InterpreterNotFound: https://travis-ci.org/davidhalter/jedi/jobs/187838411.

Likewise, "python: 3.6" && TOXENV=py35 fails with the same error: https://travis-ci.org/davidhalter/jedi/jobs/187838832.

davidhalter: Ok. Thank you for that information ;)

davidhalter:
Are you sure that you need to exclude 3.5? Now that one fails with `ERROR: InterpreterNotFound: python3.5`

davidhalter:
Not sure if it's possible the same way we do it now. Maybe we need to switch to something like https://github.com/twisted/klein/blob/master/.travis.yml and specify all the stuff in the matrix.

davidhalter:
Or we just use https://github.com/ryanhiebert/tox-travis

It looks easier and more straight forward.

fsouza:
@davidhalter tox-travis looks interesting! I'll give it a try, thank you!

davidhalter:
This is what I mean! Thank you very much for improving it. I like!

PR #829, opened 2017-01-15
[RFC] Do not use sys.version_info with _get_venv_sitepackages

blueyed:
Those might differ, e.g. when Neovim's host is Python 3.6, but the
virtualenv is 3.5.2.

### DIFF
--- a/jedi/evaluate/sys_path.py
+++ b/jedi/evaluate/sys_path.py
@@ -55,7 +55,8 @@ def _get_venv_path_dirs(venv):
     venv = os.path.abspath(venv)
     sitedir = _get_venv_sitepackages(venv)
     sys_path = []
-    addsitedir(sys_path, sitedir)
+    if sitedir:
+        addsitedir(sys_path, sitedir)
     return sys_path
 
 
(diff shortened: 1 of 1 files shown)


blueyed:
There should be some warning being logged in that case probably.
Would it be possible for tools like jedi-vim then to get notified about it / display it?

davidhalter:
I actually really don't want to add this, because it leads to extremely weird segfaults in some situations (like using a library like QT with the wrong Python version). I'm sorry, but it would be adding additional bugs to Jedi (that are even more annoying). 

#385 is still highly relevant and I plan to work on that ASAP. However it might still take a while :( It's just way to much work such a library.

JakobGM:
@davidhalter Would you consider adding this behind an experimental flag? 

Unfortunately, this makes jedi-vim unusable with Python 3.6, at the moment. And this silent 'bug' took me way too long to debug.

As @blueyed suggested, some sort of logging mechanism would make it possible to alert the user of the fact that autocompletions don't work for non-Python3.5 versions.

Thanks!

blueyed:
> makes jedi-vim unusable with Python 3.6

Not really.. it depends on the version(s) you are using.
I can only recommend Neovim again, which allows for more flexibility with regard to the Python version(s) you are using/mixing.

But I agree that it makes it more difficult to figure out, especially since there is no feedback.  IIRC it took me quite a while myself back when I've created this PR.

JakobGM:
@blueyed Yeah, unusable is probably the wrong word here. For the moment, I have applied the patch manually, and will see how these segfaults will affect my normal workflow (if at all?).

Will do the research on NeoVim as a suitable permanent replacement as well. Thanks!

hbutau:
I have been trying neovim with  python3.6 and jedi-vim just doesnt work.

PR #832, opened 2017-01-19
Improve some documentation about name_with_symbols, name and completion.

Carreau:
It was unclear to me and lead to (some) bugs in IPython. 

In particular it was unclear to me of what `complete` is the suffix. And in particular when doing case insensitive completion how to calculate how many symbols to delete (and then what to insert). 

Also I cannot reproduce the example with `list()` and `.append()`, I don't know if I misunderstood it.

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -432,22 +432,35 @@ class Completion(BaseDefinition):
         """
         Return the rest of the word, e.g. completing ``isinstance``::
 
-            # <-- Cursor is here
+            isinstan# <-- Cursor is here
 
         would return the string 'ce'. It also adds additional stuff, depending
         on your `settings.py`.
+
+        Assuming the following function definition::
+
+            def foo(param=0):
+                pass
+
+        completing ``foo(par`` would give a ``Completion`` which `complete`
+        would be `am=`
+
+
         """
         return self._complete(True)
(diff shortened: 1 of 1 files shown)


jedi/api/classes.py
```diff
@@ -432,10 +432,20 @@ def complete(self):
...
         would return the string 'ce'. It also adds additional stuff, depending
         on your `settings.py`.
+
+        Assuming the following function definition:
+
+            def foo(param=0):
+                pass
+
+        completing ``foo(par`` would give a ``Completion`` which `complete`
+        would be `m=`
```
davidhalter: not `am=`?

jedi/api/classes.py
```diff
@@ -445,9 +455,18 @@ def name_with_symbols(self):
         Similar to :attr:`name`, but like :attr:`name`
         returns also the symbols, for example::
 
-            list()
+            list()| <-- cursor is here
+
+        Would return ``.append`` and others (which means it adds a dot).
```
davidhalter: This feature was actually removed a while ago to simplify and generalize completions. I think you can safely remove the above 3 lines.

Carreau:
Done, and done.

davidhalter:
Thanks mate!

PR #834, opened 2017-01-21
Fix keywords detected as modules

Carreau:
Also improve a tiny-bit an error message.


davidhalter:
Thanks! I think it's cleaner to move the `api_type = 'keyword' into `jedi.api.keywords.KeywordName`, though.

Carreau:
> Thanks! I think it's cleaner to move the api_type = 'keyword' intojedi.api.keywords.KeywordName`, though.

Ah ! Sure ! Done.

PR #840, opened 2017-01-27
Fix uninitialized variable reference

ccope:
If the length of paths is 0, module_path was undefined.


ccope:
This is the same issue as #736

ccope:
So, I can easily reproduce the issue, but I'm not sure what the best solution is here. I think it happens when trying to find modules that aren't in the PYTHONPATH.

davidhalter:
Pretty sure it's not as easy as that. If possible, please try to create a test case first. It will be very easy for me to solve the issue if I have a proper test case.

Issue #846, opened 2017-02-04
Installing jedi downloads 24MB


takluyver:
There appears to be a random .wav file in the docs folder of the sdist tarball with nearly 2 hours of audio.

davidhalter:
Thx. I will fix this.

takluyver:
Thanks. Heads up: PyPI does not let you replace a file that was uploaded once, so fixing it will probably require a new version number.

davidhalter:
I have just uploaded the wheel for Python 3. Let me know if somebody really needs the sources. I might need to release a new version in that case.

takluyver:
Thanks, that works for me.

You may find downstream packagers asking for an sdist - I've put a number of wheel-only packages on PyPI (using flit), and there's generally some pushback. I'll let them make that request, though.

PR #847, opened 2017-02-04
Build universal wheels

takluyver:
This tells wheel that the packages are compatible with Python 2 and 3.

The wheel currently on PyPI is only tagged 'py2', so installing on Python 3 uses the source tarball (which is why I noticed #846).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal=1


davidhalter:
Is there a way to tell the wheel that it's not compatible with Python 2.5?

takluyver:
In `setup.py` you can pass an argument like `python_requires='>=2.6'`. It needs fairly new versions of pip/setuptools to understand this, however, and if someone's still on Python 2.5, it's a good bet that they have old versions of packaging tools.

davidhalter:
Thanks man!

PR #848, opened 2017-02-04
Complete dotted names properly in decorators

Mortal:
Outside decorators, a dotted name is parsed by the grammar as
```
stmt -> test -> power -> atom trailer -> (NAME) ('.' NAME)
```
where the first `NAME` is an `atom` and the second `NAME` is a `trailer`.

Thus, testing if the most recent variable is a `trailer` and the most recent node is a `'.'` is almost always enough for Jedi to properly complete dotted names.

However, the grammar for decorators is more restricted and doesn't allow arbitrary atoms and trailers; instead, a dotted name in a decorator is
```
decorator -> '@' dotted_name -> '@' (NAME '.' NAME),
```
meaning the most recent variable will be `dotted_name`, not `trailer`.

Besides in decorators, the `dotted_name` variable is only used in import statements which are handled previously in `_get_context_completions`, so checking for `dotted_name` in this arm of the `if` only covers decorators and not inadvertently anything else.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -40,6 +40,7 @@ Guido van Rossum (@gvanrossum) <guido@python.org>
 Dmytro Sadovnychyi (@sadovnychyi) <jedi@dmit.ro>
 Cristi Burcă (@scribu)
 bstaint (@bstaint)
+Mathias Rav (@Mortal) <rav@cs.au.dk>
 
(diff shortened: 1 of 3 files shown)


davidhalter:
Thanks a lot! I have moved the tests to test/completion/decorators.py`.

Issue #850, opened 2017-02-06
Fast, Non-lazy Completions.type


Carreau:
Hey there !

For once not a bug, but a request for improvement. 
With current IPython integration, I need to greedily compute the `.type` of all completions.
For the terminal version I could it lazily, notebook I can't. In some case (`matplotlib.` ) this can be gargantuesqually slow (don't know if it's an english adverb, but anyway), and AFAICT, jedi is reparsing – or not reusing some available context.

I do understand it can be hard to "Guess" if the context have changed.

So is there a way to tell jedi (like using a context manager): Thinks are not going to change underneath you, can you compute all these things at once ?

Right now it seem like an LRU cache in `mixed.py` does the trick ( line 120ish)

```
@functools.lru_cache()
def parse(grammar, path):
    with open(path) as f:
        source = f.read()
    source = common.source_to_unicode(source)
    return FastParser(grammar, source, path)
```

Though I understand that FastParser is _also_ made to recompute only what's necessary using DiffParser. But FastParser is supposed to have CachedFastParser as a metaclass so should it be cached .. but is not... Anyway getting late let me know if things make sens

davidhalter:
The parser structure/calling is really bad. I know that and I will be working on that. I think you're on to something that I have to look into first.

At least over the lifetime of an Evaluator there is no need to re-parse anything (or even open a file).

davidhalter:
We have fixed this one afaik.

PR #851, opened 2017-02-06
Performance improvement.

Carreau:
Followup #850. It's late but I push my work to resume it tomorrow and let you know I'm working on it.

```
 ~/dev/ipython master "!!" $ time python jedi_slow.py # with patch
        2.20 real         1.91 user         0.20 sys
 ~/dev/ipython master "!!" $ time python jedi_slow.py # without.
       21.48 real        21.05 user         0.30 sys
 ~/dev/ipython master "!!" $ cat jedi_slow.py
import matplotlib
import jedi
cp = jedi.Interpreter('matplotlib.', [{'matplotlib':matplotlib}], line=1, column=len('matplotlib.')).completions()

for c in cp:
   a = c.type
```

Also [stay away as far as you can from difflib from stdlib if you can](http://carreau.github.io/posts/08-Dear-DiffLib.html). Also because it has some between O(n^2) and O(n^3) behavior in some weird input. I need to finish writing my difflib2 package, and publish the benchmarks. So I'm sure I can write a small standalone differ that would speed the diffparser of Jedi.

(I have to recheck some microbenchmark)

### DIFF
--- a/jedi/evaluate/compiled/mixed.py
+++ b/jedi/evaluate/compiled/mixed.py
@@ -3,6 +3,7 @@ Used only for REPL Completion.
 """
 
 import inspect
+import functools
 import os
 
 from jedi import common
@@ -11,6 +12,9 @@ from jedi.evaluate import compiled
 from jedi.cache import underscore_memoization
 from jedi.evaluate import imports
 from jedi.evaluate.context import Context
+
+
+lru_cache = getattr(functools, 'lru_cache', lambda:(lambda x:x))
 
 
 class MixedObject(object):
@@ -115,6 +119,7 @@ class MixedObjectFilter(compiled.CompiledObjectFilter):
         #return MixedName(self._evaluator, self._compiled_object, name)
 
 
+@lru_cache()
 def parse(grammar, path):
     with open(path) as f:
         source = f.read()
--- a/jedi/parser/diff.py
+++ b/jedi/parser/diff.py
@@ -14,7 +14,7 @@ from jedi import settings
 from jedi.common import splitlines
 from jedi.parser import ParserWithRecovery
 from jedi.parser.tree import EndMarker
-from jedi.parser.utils import parser_cache
+from jedi.parser.utils import parser_cache, ParserCacheItem, save_parser, load_parser, ParserPickling
 from jedi import debug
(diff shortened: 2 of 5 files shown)


Carreau:
Ok, here is my take on speeding things up. 

----


Some performance enhancement.

- Make sure the `CachedFastParser` are cached on disk
- Change the pickle serialisation on disk is optimized **on write**,
this does not change the pickling version, it just make it more
efficient (unless things have already been written)
- Do not trust the `SequenceMatcher` to be efficient if the two files
are the same, just take a shortcut and return previous completer.
- Do some argument caching in the core of the parser loop.

With this the 2 remaining method that could be made faster and are
taking most of the time are:
    - the `splitlines` one, but hard to be made much faster
    - the `pickle.load` one, if we come up with our own serialisation.

With this, on my machine I have in some case a ~15 fold increase in
speed. Leading to things like `matplotlib.<tab>` becoming reactive
enough to use.


--- 

I've couple that on IPython to compute completion types  at most for the first `0.4` seconds of the spend on the completion generator. So it hopefully should make things reasonable even on slowish machines. 

I'm not ready to write a custom Parser serializer for now.

Carreau:
Ok, I have no clue about this last failure. 

I'll re-dig into this later, or try to find out which part of this make it fail.

Carreau:
Rebase on top of #866, and tests seem to pass with it.

davidhalter:
@carreau I finally had time to review this one. I like many of the ideas, but at the same time there's some stuff that doesn't make sense or is simply not complete (the lru stuff for example is only Python2).

Would it be possible for you to create different commits for the different sets of changes to allow me to cherry-pick from it? Would be really nice :)

Carreau:
Done as #890 #891 #892 #893 #894

davidhalter:
Thanks man. I didnt' even want you to do so much work :) Simple commits in this branch would have been enough. But thanks anyways!

Carreau:
> Thanks man. I didn't even want you to do so much work :) Simple commits in this branch would have been enough. But thanks anyways!

It was not harder, once you have all the commits you juste `rebase -i` and push to a different branch. :-)

davidhalter:
Closing this one in favor of the other 6 or so pull requests.

PR #852, opened 2017-02-06
Added support for implicit namespace packages (PEP 420)

dfit99:
Hi, I added support for implicit namespace packages as specified by PEP 420. These changes will work for Python versions 3.4 and above. The tests I added are essentially copy and pastes of the existing namespace packages tests, but with the __ init__.py files removed from the test directories. 

In _compatibility.py I added a new function ,`find_module_py34`, that makes use of `importlib.PathFinder.find_spec` to get the module spec.  If the module's origin is 'namespace', I to return an `ImplicitNSInfo` object which contains useful information about the namespace package (the package name and the list of paths in which this package lives in). `find_module_py34` also checks to see if it took in a `ImplicitNSInfo` object (when the parent module is an implicit namespace package).

In imports.py I check if module_path is of type `ImplicitNSInfo`. If so, I return  an `ImplicitNSWrapper`, which extends `ModuleContext`. The creational logic for making an `ImplicitNSWrapper` entails passing in a dummy parsed module and the corresponding `ImplicitNSInfo` object. 

In the `ImplicitNSWrapper` class, I  overrode `py__path__` to return the `ImplicitNSInfo` object in which `ImplicitNSWrapper` wraps around. I also overrode `_sub_modules_dict` in order to account for the multiple locations in which this package can live in. Finally, I overrode  `py__package__` which will get the package name from the wrapped `ImplicitNSInfo` object.


davidhalter:
That's awesome. I'll try to review it ASAP. I was close to adding support for this myself.

jedi/evaluate/representation.py
```diff
@@ -557,3 +558,33 @@ def _sub_modules_dict(self):
...
+    def _sub_modules_dict(self):
+        names = {}
+
+        paths = self.implicit_ns_info.paths
+        file_names = chain.from_iterable(os.listdir(path) for path in paths)
+        mods = [file_name.rpartition('.')[0] if '.' in file_name else file_name for file_name in file_names]
+
+        for name in mods:
+            names[name] = imports.SubModuleName(self, name)
+        return names
```
davidhalter: I think this whole class should be a separate context. If you really need part of the Module class (i.e. some methods of it), I suggest we refactor and create an AbstractModule to reuse code. If you don't have `module` as a parameter to this namespace, there will also be no need for parsing empty files in `imports.py`. The contexts are flexible enough to make them bend to whatever you want to do.

Please also call it `ImplicitNamespace` or `ImplicitNamespaceContext`, that's clearer.

`get_filters` would look like this:

```
get_filters(self, search_global, until_position=None, origin_scope=None):
    yield DictFilter(self._sub_modules_dict())
```

There's no need to do all the other module stuff (like star imports) as well.

jedi/_compatibility.py
```diff
@@ -34,8 +34,32 @@ def close(self):
...
+    implicit_namespace_pkg = False
+    spec = None
+    loader = None
+
+    if isinstance(path, ImplicitNSInfo):
+        old_dotted_path = path.name
+        path = path.paths
+    spec = importlib.machinery.PathFinder.find_spec(string, path)
+    if hasattr(spec, 'origin'):
+        origin = spec.origin
```
davidhalter: Why do you need origin? It's not used anywhere as far as I can see?

jedi/_compatibility.py
```diff
@@ -34,8 +34,32 @@ def close(self):
...
+        implicit_namespace_pkg = origin == 'namespace'
```
davidhalter: Why is this a string? Could be True or am I wrong?

dfit99: Here I am just assigning `implicit_namespace_pkg` to the result of the  evaluated expression `origin == 'namespace'`
 It is the same thing as saying:
`implicit_namespace_pkg  = True if origin == 'namespace' else False
` 
I'll refactor this to make it clearer and I'll change `implicit_namespace_pkg` to `is_implicit_namespace_pkg` to make it more explicit that it is a boolean value.

jedi/_compatibility.py
dfit99: The origin is needed to determine whether or not we are dealing with implicit namespace packages. If `origin == 'namespace'` we can safely assume we are dealing with implicit namespace packages.

jedi/evaluate/representation.py
dfit99: Thank you for reviewing this so quickly. I agree with your suggestions here.  I'll go ahead and refactor this.

jedi/_compatibility.py
davidhalter: LOL no you don't have to. I have misread this. Don't ask me how that happened. I read `implicit_namespace_pkg = origin = 'namespace'`, which just clearly didn't make any sense. Haha, sorry my bad.

jedi/_compatibility.py
davidhalter: Also my bad. (See comment below)

jedi/evaluate/representation.py
dfit99: I refactored the code, it is now called `ImplicitNamespaceContext` and does not inherit from `ModuleContext` anymore. I've also added an `ImplicitNSName` class (which extends `AbstractNameDefinition`) that infers to nothing to prevent Jedi from raising exceptions. Let me know what you think!

jedi/evaluate/representation.py
```diff
@@ -557,3 +558,65 @@ def _sub_modules_dict(self):
...
+
+
+class ImplicitNamespaceContext(use_metaclass(CachedMetaClass, context.TreeContext)):
+    """
+    Provides support for implicit namespace packages
+    """
+    api_type = 'module'
+    parent_context = None
+
+    def __init__(self, evaluator, implicit_ns_info=None):
```
davidhalter: I think the only thing left to complain about for me is the nesting of those classes. I would prefer if you just copied the name and paths of the ImplicitNSInfo class over to this one. It's just way more readable IMO. But otherwise, really good job!

dfit99: Thanks! I refactored the code a bit. I got rid of the nested classes. ImplicitNamespaceContext has no more references to ImplicitNSInfo. Let me know if there is anything else I can do.

davidhalter:
Thanks man! This is a really good contribution. Appreciate it!! Happy to see more of this stuff :)

PR #853, opened 2017-02-07
Missing assert

Carreau:
Either this is a missing assert, or it should have been an assignment.

That's the test which annoy me on #851 that I don't understand, and fail anyway locally as well.

### DIFF
--- a/test/test_parser/test_old_fast_parser.py
+++ b/test/test_parser/test_old_fast_parser.py
@@ -257,7 +257,7 @@ def test_string_literals():
     """)
 
     script = jedi.Script(dedent(source))
-    script._get_module().tree_node.end_pos == (6, 0)
+    assert script._get_module().tree_node.end_pos == (6, 0)
     assert script.completions()
 
 


Carreau:
Actually that's super weird, here is what I've got here:

```python
import jedi

source = """
x = ur'''

def foo():
    pass
"""

script = jedi.Script(source)
print([(p,p.end_pos) for p in script._get_module().tree_node.children])
```

```
[(<ErrorNode: x = ur@2,0>, (2, 6)),
 (<ErrorLeaf: errortoken:"'''\n\ndef foo():\n    pass\n", (2, 6))>, (6, 0)),
 (<EndMarker: >, (5, 9))
]
```

Which seem to be wrong as parser assume the last token is the last position and uses this for valid ranges.

davidhalter:
WTF your example above is super weird. I'm really wondering why the tests are not failing. Can you add a new issue with the exact content of the above?

Carreau:
Thanks !

> WTF your example above is super weird. I'm really wondering why the tests are not failing. Can you add a new issue with the exact content of the above?

Yes I will.

PR #857, opened 2017-02-09
Catch FileNotFoundError when opening file cache

andylee-ncc:
After running jedi a long time, a cache file might get deleted, causing a crash when it is opened. This commit handles the FileNotFoundError gracefully.

### DIFF
--- a/jedi/parser/utils.py
+++ b/jedi/parser/utils.py
@@ -10,6 +10,7 @@ import pickle
 
 from jedi import settings
 from jedi import debug
+from jedi._compatibility import FileNotFoundError
 
 
 def underscore_memoization(func):
@@ -127,12 +128,15 @@ class ParserPickling(object):
             # the pickle file is outdated
             return None
 
-        with open(self._get_hashed_path(path), 'rb') as f:
-            try:
-                gc.disable()
-                parser_cache_item = pickle.load(f)
-            finally:
-                gc.enable()
+        try:
+            with open(self._get_hashed_path(path), 'rb') as f:
+                try:
+                    gc.disable()
+                    parser_cache_item = pickle.load(f)
+                finally:
+                    gc.enable()
+        except FileNotFoundError:
+            return None
 
         debug.dbg('pickle loaded: %s', path)
         parser_cache[path] = parser_cache_item
--- a/test/test_cache.py
+++ b/test/test_cache.py
@@ -3,6 +3,7 @@ Test all things related to the ``jedi.cache`` module.
 """
 
 import time
+from os import unlink
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


davidhalter:
I wonder why this doesn't fail in the tests. `FileNotFoundError` is part of Python 3 and not 2.

I suggest you add a test here (for this exact issue).

andylee-ncc:
I cannot get it to happen consistently. I am using macOS. It might be macOS cleaning the cache.

I note that you put the cache file in `~/Library/Cache`. According to [dev docs](https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html),

> Note that the system may delete the Caches/ directory to free up disk space, so your app must be > able to re-create or download these files as needed.

jedi/parser/utils.py
```diff
@@ -11,6 +11,7 @@
 from jedi import settings
 from jedi import debug
 
+FileNotFoundError = getattr(__builtins__, 'FileNotFoundError', IOError) # -- from cwallenpoole on SO
```
davidhalter: Can you put that into `_compatibility.py`?

davidhalter:
I think you're right and it makes sense to add this. I just require a test case for such things. If you don't create one, I will just leave it open until someone does.

andylee-ncc:
Would you consider it a valid test if I just simulated the deletion of a cache file by an external process, instead of actually triggering it?

davidhalter:
I guess so, if it's not overly broad.

andylee-ncc:
Added test in `test_cache.py`, squashed commits, changed base to origin/dev.

davidhalter:
Thanks man!

andylee-ncc:
Your welcome.

I'd like to ask a favor. I just noticed I that I messed up my email in one of the commits. I see that the dev branch just merged my commits, and given it is relatively easy to fix this at the moment, could you please redo the merge, with my corrected commits? I've force-pushed the commits with the fixed email. (I wouldn't be asking if this wasn't near the head of the dev branch.)

davidhalter:
The problem is that I have already a bit of code on top of the dev branch (and I have also done a more complicated merge). Is it really that important? Because it's only like one commit.

andylee-ncc:
Well, I don't like leaking that particular email to the public. If you publish the complicated merge I could try to rework the tree. If I succeed then you could diff the top with what you have to make sure the contents didn't change. Or I could try this <https://help.github.com/articles/changing-author-info/> and post steps that you can inspect and then run to reproduce the name-change.

davidhalter:
I have updated the current dev branch. Please pull from it, change what you need and I will force push. But please do it very soon.

andylee-ncc:
OK, I have reworked the dev branch at <https://github.com/andylee-ncc/jedi/tree/dev>.

Thank you for your patience with this.

davidhalter:
Welcome ;-)

PR #864, opened 2017-02-11
readme: Add maralla/completor.vim

polyzen:
https://github.com/maralla/completor.vim

https://github.com/maralla/completor.vim#python


Carreau:
Sweet.

+1

Issue #855, opened 2017-02-08
End Marker not at the end. 


Carreau:
```
$ git describe --tags; cat foo.py ; python foo.py
v0.10.0-13-g8131f197
print('=======')
import jedi

source = """
x = ur'''

def foo():
    pass
"""

nsource = '\n'.join([l.rstrip() for l in source.splitlines()])+'\n'
assert source == nsource , "Make sure no surprise with trailing space."

script = jedi.Script(source)
for p in script._get_module().tree_node.children:
    print(p, 'end:', p.end_pos)
print('=======')
=======
<ErrorNode: x = ur@2,0> end: (2, 6)
<ErrorLeaf: errortoken:"'''\n\ndef foo():\n    pass\n", (2, 6))> end: (6, 0)
<EndMarker: > end: (5, 9)
=======
```

As you can see the end marker `(5,9)`, is before `(6,0)`. I've kinda bisect it to when you introduced the Binary tree to speed up the localisation of that, but it seem to have been an earlier latent bug.

Carreau:
Or even simpler:

```
source = """'''
"""
``` 

That is to say quote-quote-quote-newline is enough.

Carreau:
Or, so so far I've tracked it down to an inconsistency between `source_token` in `jedi.parser.tokenize` and LeafWithNewLines (`tree.py`) the tokenizer is already "wrong" (or right, depending on which side you want to fix):

```
list(jedi.parser.tokenize.source_tokens('"""\n'))
Out[4]:
[TokenInfo(type=ERRORTOKEN, string='"""\n', start=(1, 0), prefix=''),
 TokenInfo(type=ENDMARKER, string='', start=(1, 4), prefix='')]
```

PR #866, opened 2017-02-13
Make sure error token set the new_line flag when necessary

Carreau:
Should solve #855

As far as I can tell, this should be a step toward the right thing to do. I'm going to assume it's likely not totally right and will make other test fail.

### DIFF
--- a/jedi/parser/tokenize.py
+++ b/jedi/parser/tokenize.py
@@ -233,6 +233,8 @@ def generate_tokens(readline, use_exact_op_types=False):
         if not line:
             if contstr:
                 yield TokenInfo(ERRORTOKEN, contstr, contstr_start, prefix)
+                if contstr.endswith('\n'):
+                    new_line = True
             break
 
         lnum += 1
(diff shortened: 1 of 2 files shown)


Carreau:
Well, strangely test appear to to be happy, and it has been working locally. 
I'm still going to attempt writing a few more tests.

jedi/parser/tokenize.py
```diff
@@ -233,6 +233,8 @@ def generate_tokens(readline, use_exact_op_types=False):
         if not line:
             if contstr:
                 yield TokenInfo(ERRORTOKEN, contstr, contstr_start, prefix)
+                if contstr.endswith('\n'):
+                    new_line = True
```
davidhalter: I like! Can you add a test for this one?

Carreau:
> I like! Can you add a test for this one

Yes, thanks for reminding me. 
Done.

davidhalter:
Thanks man and sorry for taking such a long time for a patch that's only a few LOC. :)

Carreau:
it's ok, no hurry. I have plenty other things on my plate as well, so I can keep myself busy :-)

PR #867, opened 2017-02-13
Use start and end position in repr for simpler debugging.

Carreau:


jedi/parser/tree.py
```diff
@@ -317,7 +317,7 @@ def nodes_to_execute(self, last_added=False):
 
     @utf8_repr
     def __repr__(self):
-        return "<%s: %s>" % (type(self).__name__, self.value)
+        return "<%s: %s %s-%s>" % (type(self).__name__, self.value, self.start_pos, self.end_pos)
```
davidhalter: Hmm, I would prefer if this was just the start position. It's harder for me when debugging, otherwise (When enabling Jedi's internal debugging).

Carreau:
> just the start position....
> Also it's usually pretty easy to calculate the end in your head.

Sure, I got confused with position when debugging the newline bug. Start only if fine for me and done.

Carreau:
Thanks !

PR #868, opened 2017-02-14
Return types from numpydocs and compiled objects return types from docstrings

bcolsen:
In this PR I have updated the docstrings.py from @erotemic in PR: #796 so that it merges against the changes that have been made to the current dev branch since the original commits. I have also removed the google docstrings from this PR as I don't have the experience necessary to test these docstrings. However, It should be quite simple for @erotemic to add this functionality as an addition PR. With this split there are less code changes, hopefully making this PR easier to merge.

This pull request solves issue #372 for numpy complied objects. 

Numpy functions such as numpy.array return an numpy.ndarray object resulting in completions like this:

> import jedi; jedi.Script('import numpy as np; np.array([1,2,3]).a').completions()
[<Completion: all>,
 <Completion: any>,
 <Completion: argmax>,
 <Completion: argmin>,
 <Completion: argpartition>,
 <Completion: argsort>,
 <Completion: astype>]

I have not yet written a test since I need to import numpydocs and numpy to do such. 
A standard python library with complied objects and sphinx documentation would be a better test.

### DIFF
--- a/jedi/evaluate/docstrings.py
+++ b/jedi/evaluate/docstrings.py
@@ -1,11 +1,12 @@
 """
 Docstrings are another source of information for functions and classes.
 :mod:`jedi.evaluate.dynamic` tries to find all executions of functions, while
-the docstring parsing is much easier. There are two different types of
+the docstring parsing is much easier. There are three different types of
 docstrings that |jedi| understands:
 
 - `Sphinx <http://sphinx-doc.org/markup/desc.html#info-field-lists>`_
 - `Epydoc <http://epydoc.sourceforge.net/manual-fields.html>`_
+- `Numpydoc <https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt>`_
 
 For example, the sphinx annotation ``:type foo: str`` clearly states that the
 type of ``foo`` is ``str``.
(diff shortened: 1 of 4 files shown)


bcolsen:
I forgot to add the testing file to this PR.
I'll add the tests with the google docstrings test removed.

Is there a way to make a complied object to test the compiled object type return?

ccordoba12:
@davidhalter, could you take a look at this one please?

We're maintaining a big patch in Spyder to make this work, and we would really like to drop it in favor of using an upstream release.

davidhalter:
Seriously I'm sorry. I just don't have the time to review this. There were some other priorities (including real life) that made it hard to review this. It's really the next thing on my list once I have time.

I'll try ASAP.

ccordoba12:
Thanks for the answer, and good to know this one is high in your TODO list.

davidhalter:
Sorry again for taking so long. I finally had time to review #796 and have tried to reach the author. If we don't get that working in the next few weeks, we'll just merge this one. I hope this is OK for you.

ccordoba12:
Thanks David, that's fine for us.

ccordoba12:
@bcolsen, could you update this PR in the meantime? It has some conflicts with master.

bcolsen:
@ccordoba12, yeah I'll try a rebase on the dev branch.

bcolsen:
Well that was more difficult than I thought it would be. 

There were some breaking changes in docstrings.py, but I fixed it.

ccordoba12:
@davidhalter, now that the original author of #796 has confirmed that he doesn't have time to update his work, I think we can proceed with this one.

jedi/evaluate/docstrings.py
```diff
@@ -55,14 +59,49 @@ def _search_param_in_numpydocstr(docstr, param_str):
...
-                    return [p_type]
+                return _expand_typestr(p_type)
         return []
 
+    def _search_return_in_numpydocstr(docstr):
+        r"""
+        Search `docstr` (in numpydoc format) for type(-s) of `param_str`.
+        """
+        doc = NumpyDocString(docstr)
+        returns = doc._parsed_data['Returns']
```
davidhalter: I just realized again that we're using a non-public API here. Can we please have a try/except around it and catch both key error and

```
try:
    # This is a non-public API. If it ever changes we should be prepared and return gracefully.
    returns = doc._parsed_data['Returns']
    returns += doc._parsed_data['Yields']
except (KeyError, AttributeError):
    return []
```

Please also change this for the param version of this function.

jedi/evaluate/docstrings.py
```diff
@@ -55,14 +59,49 @@ def _search_param_in_numpydocstr(docstr, param_str):
...
+        returns += doc._parsed_data['Yields']
+        found = []
+        for p_name, p_type, p_descr in returns:
+            if not p_type:
+                p_type = p_name
+                p_name = ''
+
+            m = re.match('([^,]+(,[^,]+)*?)$', p_type)
```
davidhalter: WTF does this do?

jedi/evaluate/docstrings.py
```diff
@@ -55,14 +59,49 @@ def _search_param_in_numpydocstr(docstr, param_str):
...
+    # Check if alternative types are specified
+    if re.search('\\bor\\b', p_type):
+        types = [t.strip() for t in p_type.split('or')]
+    # Check if type has a set of valid literal values
+    elif p_type.startswith('{'):
+        if not is_py3:
+            # python2 does not support literal set evals
+            # workaround this by using lists instead
+            p_type = p_type.replace('{', '[').replace('}', ']')
+        types = set(type(x).__name__ for x in literal_eval(p_type))
```
davidhalter: Jedi has it's own ways of parsing stuff. No need for this. I'm not sure what this does anyway, but I'm pretty sure that it's not the way how you should do things.

bcolsen:
@davidhalter Thank you for looking into this.  

I have read your comments and did more research into what this code is doing since I have been mostly porting it over up until now. 

I have already started to simplify the code and write my own tests.

jedi/evaluate/docstrings.py
bcolsen: Nothing I think. I removed it and it still works.

It seems to be copied from the exiting numpydocs params function:
`m = re.match('([^,]+(,[^,]+)*?)(,[ ]*optional)?$', p_type)`
which seems like a complicated way of removing the `, optional` from the parameter type string.

jedi/evaluate/docstrings.py
bcolsen: Numpydoc allows types to be a set of literals like:
```
order : {'C', 'F', 'A'}
```
It works like this:
```
>>> from ast import literal_eval
>>> set(type(x).__name__ for x in literal_eval("['x', [], 2.3]"))
{'float', 'list', 'str'}
```

This code in already in docstrings.py:
https://github.com/davidhalter/jedi/blob/dev/jedi/evaluate/docstrings.py#L60

All this being said I'm sure that there is already a way to do this in jedi

davidhalter:
Let me know when you're ready!

bcolsen:
It's ready.  I reduced and rewrote the code using `yields` to return multiple types instead of lists.

The `_expand_typestr` function is likely the worst but it is tested and works to parse the numpy type strings for multiple type.

I made a bit of a mess in tox.ini it seems that numpydocs doesn't support py2.6 or py3.3 although the next version of jedi might drop those versions as well.

I also can't seem to make a test for yields.  Jedi returns a generator object.

davidhalter:
@bcolson Good work! Thanks. I think it's pretty cool that we have this now within Jedi.

I made one minor change: literal_eval should not be used. It leads to potential exceptions. I instead used jedi's parser (parso) to parse it and work with it. I hope that's fine. This also gets rid of the set literals hack (even though I think that was not needed in the first place, because they are not supported in 2.6 but are in 2.7. However since numpydocs doesn't work in 2.6 you can just ignore 2.6.).

> I also can't seem to make a test for yields. Jedi returns a generator object.

Hmm if you're still interested, that might be true, but you can just use `next` or a for loop in test code to see what's in there.

bcolsen:
Thanks for the review!  It is cool to be completing Numpy now.

The parser code for the literal is much better than trusting `eval`.

>Hmm if you're still interested, that might be true, but you can just use next or a for loop in test code to see what's in there.

Of course, I wasn't thinking right about that one.  The tests work either way but I'll give it a shot with `next`.

Issue #875, opened 2017-02-23
Minimal Crash 


Carreau:
While debuggin #874 

```
import pandas as pd 
s = pd.Series() 
import jedi
jedi.Interpreter('s.',[ {'s':s}], line=1, column=2).completions()[0].type
```

Should trigger crash in 2 position that use != (replace by `is not`), then `==` (fake.py line 135). Not sure it's the right fix, but I takes notes before forgetting.

Carreau:
Once these are fixed, with is/is not:

```
jedi.Interpreter('s.dat',[ {'s':s}], line=1, column=5).completions()[0].type
```

fails with a 3 stage error Attribute/Key/Value error https://gist.github.com/Carreau/4ed0987c3671d9cb92cf3b8a29542619

davidhalter:
Is this fixed now?

Carreau:
Sorry I have been away and offline for about 10 days, and need to catch up with things. I've seen you did a lot of refactor and need to reinvestigate all of that. It's on my todo list.

davidhalter:
It's working again on the virtualenv branch.

Issue #874, opened 2017-02-21
Print warnings make it hard to use


kernc:
I'm not sure this is jedi bug as the whole stack is: Python < IPython < p**t**python < jedi, but it only manifests since upgrading to jedi 0.10.0.

Environment: jedi==0.10.0, ptpython==0.36, ipython==5.2.2, pandas==0.19.0+479 (git).

Typing the following in the prompt:
```py
>>> import pandas as pd
>>> s = pd.Series()
>>> s.
```
Now, as soon as the last period is pressed, the following FutureWarning is printed at cursor location:
```py
/home/user/venv/lib/python3.4/site-packages/jedi/evaluate/compiled/__init__.py:328: FutureWarning: is_time_series is deprecated. Please use Series.index.is_all_dates
  getattr(obj, name)
```
Which makes the whole prompt input buffer hard to read:
```py

>>> import pandas as pd
>>> s = pd.Series()
>>> s./home/user/venv/lib/python3.4/site-packages/jedi/evaluate/compiled/__init__.py:328: FutureWarning: is_time_series is deprecated. Please use Series.index.is_all_dates
  getattr(obj, name)
|  ← cursor position
```
This repeats _every_ time `s` is followed by a `.`.

Downgrading to jedi 0.9.0 works.

davidhalter:
@Carreau Did you see this as well?

Carreau:
Yes/No. I can trigger an error as well in pure IPython, but it does not mess up the layout. 

Digging in I can trigger an error:

```
is_instance = self._is_instance or fake.is_class_instance(obj)
        # ``dir`` doesn't include the type names.
        if not inspect.ismodule(obj) and obj != type and not is_instance:
            for filter in create(self._evaluator, type).get_filters():
                names += filter.values()
        return names

```

the `ob != type` trigger a custom `__eq__` of pandas crash some things. replacing by `is not` seem worse.

davidhalter:
> replacing by is not seem worse.

How do you mean that?

Carreau:
> How do you mean that?

I get more errors with IPython. Digging in. It's due to IPython trying to get the completions  `types` and many of them error.

kernc:
Can no longer reproduce on Python 3.5, jedi 0.10.2.

davidhalter:
Well the problem is probably that in general that warnings can be triggered by certain actions within Jedi. I'm thinking about wrapping certain parts of Jedi to avoid those warnings.

[...]

PR #876, opened 2017-02-23
'fix a couple of error locations'

Carreau:
That seem to avoid #875 I don't have much time to look into it. with this I don't get the behavior of #874 nor the crash I observed while debugging.


kernc:
I still reproduce #874 with this. :cry:

davidhalter:
I think this all makes sense and reduces code execution issues.

PR #881, opened 2017-03-01
Only resolve names for actual modules in usages

micbou:
Consider the following files:
```python
# file1.py
abc = 3
```
```python
# file2.py
from file1 import abc
abc
```
When calling `usages` on the second `abc` in file2.py, Jedi will return these definitions :

Definition | File | Cursor position
--- | --- | ---
\<Definition instance int> | None | (None, None)
\<Definition abc = 3> | file1.py | (2, 0)
\<Definition instance abc>| file2.py | (2, 10)
\<Definition abc> | file2.py | (3, 0)

Obviously, the first definition shouldn't be returned. AFAIK, this happens because Jedi assumes that `abc` on the import line is of type `module` in [the `resolve_names` function](https://github.com/davidhalter/jedi/blob/master/jedi/api/usages.py#L12-L22) which is not the case. This is fixed by checking that the underlying types (contexts) are actual modules.

Added a test.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/881)

### DIFF
--- a/jedi/api/usages.py
+++ b/jedi/api/usages.py
@@ -14,8 +14,9 @@ def usages(evaluator, definition_names, mods):
             if name.api_type == 'module':
                 found = False
                 for context in name.infer():
-                    found = True
-                    yield context.name
+                    if isinstance(context, ModuleContext):
+                        found = True
+                        yield context.name
                 if not found:
                     yield name
             else:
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks man!

And sorry for usages being pretty much worthless in general. I hope I can improve that situation in the future.

PR #891, opened 2017-03-14
Optimize Pickle on write.

Carreau:
Optimising is slower on write, but gains across multiple
reload, accumulate.
We could provide a "reoptimize all jedi function that load all pickled
file, optimize and re-write." but is it worth it ?

No need to bump the version number as reading old pickle files will
still work.

Bonus: on next pickler version upgrade old cached files will be
regenerated once and be faster !

### DIFF
--- a/jedi/parser/utils.py
+++ b/jedi/parser/utils.py
@@ -7,6 +7,7 @@ import hashlib
 import gc
 import shutil
 import pickle
+import pickletools
 
 from jedi import settings
 from jedi import debug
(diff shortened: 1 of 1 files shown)


davidhalter:
For this one I would actually want a speed comparison with old versus new.

Doesn't need to be something crazy, but a file > 1000 LOC save and load tested (separately).

I started doing this in my own performance tests, to be able to tell if my changes actually make a difference.

Carreau:
Do you mean like see the results, or commit it ? If so, where in the codebase ?

screen shot 2017-03-14 at 18 03 05

Size of dump on this particular file in 0.76 of original, I go from ~20ms to ~15ms on `pathlib` the total gain dump vs repetitive load is worth it (in this case) after about 170 loads if I read things correctly, and we take into account the ~800ms optimize pass.

I'll assume disk access is free and file loads disk->memory is peanuts.

davidhalter:
That's exactly what I wanted. How much slower is writing it (as a factor)? Estimate is good enough.

Carreau:
`pathlib` dumps: 53.6ms, optimized: +810ms

Carreau:
Ouch matplotlib.axes._axes if way worth though:

dumps: 141ms optimize + 2500 ms

I'll probably have to run a couple more test to see the average cost/benefit.

PR #892, opened 2017-03-14
Make better use of Parser Caching.

Carreau:
As far as I can tell, this should make a better use of existing parser
for modules. This is the core of what get the inference of types for
some completion on my machine from ~20s to ~2s. We are still caching on
disk  and it should be possible to cache more agressively on other
locations.


davidhalter:
@Carreau I have completely refactored the parser API. It looks a lot different and I think also a lot of things are easier to understand. Would it be possible for you to review it? Especially the cache stuff. I'd like any form of feedback. These changes are not really relevant anymore, since most of the code your changing doesn't exist anymore - or in a completely different fashion.

PR #894, opened 2017-03-14
Add lru cache on parser(grammar, source)

Carreau:
This is the second part of the patch that makes `matplotlib.<tab>` go
from ~20s to 2s on my machine. It allow to reuse existing parser without
having to load them from disk.

--- 

I hardcoded a simple LRU cache for both python 2 and 3 which is limited to getting items. Seem more than necessary for this small usecase.

### DIFF
--- a/jedi/evaluate/compiled/mixed.py
+++ b/jedi/evaluate/compiled/mixed.py
@@ -12,6 +12,8 @@ from jedi.cache import underscore_memoization
 from jedi.evaluate import imports
 from jedi.evaluate.context import Context
 
+from collections import OrderedDict
+
 
 class MixedObject(object):
     """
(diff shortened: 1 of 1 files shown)


Carreau:
Well, seem like I'll have to write my own ordered dict on 2.6 :-)

Carreau:
Ok, let's try a custom minimal implementation of ordered dict. 
I did not put it in `_compatibility` as it is not a full featured implementation. 
Let me know if this is acceptable.

Carreau:
well, my implementation is wrong apparently.... let me figure that out.

davidhalter:
Don't invest too much time. There's a chance that we already have the caching infrastructure in place within Jedi.

Carreau:
> Don't invest too much time. There's a chance that we already have the caching infrastructure in place within Jedi.

I found `memoize`, but it have no upper bound on how much is cached (AFAICT).

davidhalter:
I have added my solution, I think it's how Jedi always caches. There's no upper bound, because recursion detection is typically the thing that stops Jedi from going crazy. I think this is as good as it gets for now. Happy to investigate if it you're running into cache issues.

PR #898, opened 2017-03-20
Implement signal for stopping Jedi execution

andylee-ncc:
I am implementing a code analysis tool that uses Jedi. I was having trouble with runaway Jedi execution, and also general responsiveness. This PR solves these problems.

# Commit message:
Implement `jedi.evaluate.recursion.stop_execution_signal_queue`. This
allows a controller thread to stop Jedi execution, for improved
responsiveness.

### DIFF
--- a/jedi/evaluate/recursion.py
+++ b/jedi/evaluate/recursion.py
@@ -11,7 +11,7 @@ from contextlib import contextmanager
 
 from jedi import debug
 from jedi import settings
-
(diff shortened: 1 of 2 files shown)


davidhalter:
For now this is not going to happen. Maybe in a distant future where Jedi has it's own async implementation or where we have in general more experience with async implementations.

I think if you create an async implementation of Jedi (which has been done before a couple of times for different editors), you have probably 99% of your problems solved.

PR #899, opened 2017-03-22
Fix `api.usages` so it finds cross-module usages

andylee-ncc:
Jedi could not find certain cross-module usages. The problem was in how it compared contexts in `api.usages`.

I've included a test that fails without my included fix.

### DIFF
--- a/jedi/api/usages.py
+++ b/jedi/api/usages.py
@@ -5,6 +5,10 @@ from jedi.evaluate.filters import TreeNameDefinition
 from jedi.evaluate.representation import ModuleContext
 
 
+def compare_contexts(c1, c2):
+    return c1 == c2 or (c1[1] == c2[1] and c1[0].tree_node == c2[0].tree_node)
+
+
 def usages(evaluator, definition_names, mods):
     """
     :param definitions: list of Name
@@ -40,7 +44,9 @@ def usages(evaluator, definition_names, mods):
             for name_node in m.tree_node.used_names.get(search_name, []):
                 context = evaluator.create_context(m, name_node)
                 result = evaluator.goto(context, name_node)
-                if [c for c in compare_array(result) if c in compare_definitions]:
+                if any(compare_contexts(c1, c2)
+                       for c1 in compare_array(result)
+                       for c2 in compare_definitions):
                     name = TreeNameDefinition(context, name_node)
                     definition_names.add(name)
                     # Previous definitions might be imports, so include them
--- a/test/test_api/import_tree_for_usages/__init__.py
+++ b/test/test_api/import_tree_for_usages/__init__.py
@@ -1,0 +1,4 @@
+"""
+An import tree, for testing usages.
+"""
+
--- a/test/test_api/import_tree_for_usages/a.py
+++ b/test/test_api/import_tree_for_usages/a.py
@@ -1,0 +1,4 @@
+from . import b
+
+def foo():
+    b.bar()
--- a/test/test_api/import_tree_for_usages/b.py
+++ b/test/test_api/import_tree_for_usages/b.py
@@ -1,0 +1,2 @@
+def bar():
+    pass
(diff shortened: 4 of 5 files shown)


jedi/api/usages.py
```diff
@@ -39,8 +44,14 @@ def compare_array(definition_names):
         if isinstance(m, ModuleContext):
             for name_node in m.tree_node.used_names.get(search_name, []):
                 context = evaluator.create_context(m, name_node)
-                result = evaluator.goto(context, name_node)
-                if [c for c in compare_array(result) if c in compare_definitions]:
+                try:
+                    result = evaluator.goto(context, name_node)
+                except (NotImplementedError, RecursionError) as err:
+                    logging.getLogger(__name__).error(err)
+                    continue
```
davidhalter: No.

andylee-ncc: Yeah, I didn't mean to add that. I'll remove it.

BTW, I added that part a long time ago because the usages function was running into errors and then giving up. Catching the exception and continuing the loop allowed jedi to find other usages that didn't raise errors. I logged the error because I didn't want the errors to go unnoticed. What is an alternative way you would accept?

andylee-ncc:
*Reposting some parts because I don't know where github put my previous comment (and yours)*

I've removed the try/except part. Sorry, I didn't mean to add that to the PR. I've also made a minor indexing fix to the test code.

BTW the try/except was because `goto` was raising errors and stopping`usages` from returning any other items it successfully found. I added logging so that the errors would not be hidden. What alternative would you suggest to solve this?

**Note:** The old stuff is at [fix-api.usages-bad](https://github.com/andylee-ncc/jedi/tree/fix-api.usages-bad).

davidhalter:
> What alternative would you suggest to solve this?

Fixing the actual bug. Catching NotImplementedError is like the worst error that I've ever heard.

andylee-ncc:
I see. I (mistakenly) thought you just hadn't finished this part of Jedi, and I didn't want the not implemented part to block the user from seeing all the other usages that is *was* able to find.

By your comment, I take it that Jedi should not be raising `NotImplementedError`, and when it does it means an abstractmethod was called by accident. In that case, I wonder why you are not using ABCMeta, which would catch this error earlier? Is it for compatibility reasons?

davidhalter:
> I (mistakenly) thought you just hadn't finished this part of Jedi

Maybe it even is, but that doesn't matter. I think open source libraries should strive for perfection and not work some of the time. If something is wrong we should fix it.

> I wonder why you are not using ABCMeta, which would catch this error earlier? Is it for compatibility reasons?

I'm not really using ABCMeta, because it would make some things slower. Using `NotImplementedError` can also be very useful in case you have never really finished something - like an else statement.

> Also, would you be OK with catching RecursionError?

No. I'm not ok with catching any exceptions if it's nested. These are actual bugs that you are hiding by removing `RecursionError`. It's not like this should ever happen. And if it does and it's fine, we should either increase the stack or change certain limits so that it does not happen again.

I get a lot of pull requests here that just "fix" an error by catching e.g. `AttributeError`. Now this makes debugging harder by a factor of 10. I think most Python people don't really understand that having the power of catching pretty much all exceptions does not mean that you should do it. In complex programs this always leads to extremely bad code. Abstractions are there for a reason. Catching an error that was thrown two levels deeper just doesn't make any sense 99% of the time.

davidhalter:
Ok. This looks good. I think it's a good solution for now. I still think that usages is completely and utterly broken and we have to rewrite it, but it's a good fix for now.

Would it be possible to refactor the tests a bit into a form of `jedi/test/completion/usages.py`? That would be awesome, because until now all the usages tests are there.

andylee-ncc:
In order to refactor it into the form of `jedi/test/completion/usages.py`, I think I would need to change `jedi.Script.usages` to include the code from `usages_with_additional_modules` from my test code. Otherwise the integration test would ignore the explicitly included module. I'm not sure you want that though.

davidhalter:
Ok. I see!

PR #902, opened 2017-04-05
Fix path for grammar files in MANIFEST

flacerdk:
The `grammar*.txt` files were recently moved to `jedi/parser/python`, but that change wasn't reflected in `MANIFEST.in`, so they weren't being copied over to the package. This change fixes that. (I also added the `include_package_data` option, which seems to be necessary for `python setup.py sdist`, but as I understand it isn't used by `bdist`.)


davidhalter:
Thanks man!! I hate python packaging.

PR #912, opened 2017-04-20
Fix : Jedi do not complete numpy arrays in dictionary

Carreau:
Fix ipython/ipython#10468

@stefanv does that fix your issue ? You should "just" have to install Jedi from the `dev` branch.


stefanv:
That does the trick!

davidhalter:
Thanks! That makes sense.

PR #914, opened 2017-04-21
Use stacklevel in warnings or filters don't work.

Carreau:
In particular with the right stacklevel IPython will display the warning
if code is directly entered by the user. Without this info it does not.

Use the opportunity to add in the warning since when things are
deprecated. This leads to one less lookup of information for the user.

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -93,10 +93,10 @@ class Script(object):
                  encoding='utf-8', source_path=None, source_encoding=None,
                  sys_path=None):
         if source_path is not None:
-            warnings.warn("Use path instead of source_path.", DeprecationWarning)
+            warnings.warn("Deprecated since version 0.7. Use path instead of source_path.", DeprecationWarning, stacklevel=2)
             path = source_path
         if source_encoding is not None:
-            warnings.warn("Use encoding instead of source_encoding.", DeprecationWarning)
+            warnings.warn("Deprecated since version 0.8. Use encoding instead of source_encoding.", DeprecationWarning, stacklevel=2)
             encoding = source_encoding
 
         self._orig_path = path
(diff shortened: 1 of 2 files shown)


davidhalter:
Hmm, I just don't really understand the Python documentation regarding `stacklevel`. Can you elaborate a bit more?

Carreau:
Sure. It's basically which stack frame should trigger the warning. 

```
def foo():
      warnings.warn(stacklevel=2)

foo() #<- this will trigger the warning and is "deprecated"

def qux():
    warnings.warn(stacklevel=3)

qux() #<- won't trigger

def bar():
    qux()

bar() # <- will trigger warning
```

The use if for decorators and other things to mark the code triggering the deprecation as not the current location of warn, but N frames up. It's almost always a stack value of 2 which is the "right" thing to do, but usually you want to couple that with warning filters. 

Does that make some sens, or do you want a longer explanation ?

davidhalter:
So your basic idea is that with this change you can actually implement warning filters?

Carreau:
Yes, and these filters are enabled on things like IPython: 

screen shot 2017-04-25 at 14 15 56

though if I were to define:
```python
# bar.py
def baz():
    from foo import will_warn
    will_warn()
```

The interactive usage of `baz` (`from bar import baz; baz()`) will not trigger. The filter should triggers deprecation warnings for warnings emitted by user-level code. But in order to know wether a warning is triggered by user-level code the filter needs to know the correct stacklevel – which is usually 2. 

Technically that can be used by many other things and on my todo list is to write an article that explains python warnings.

cc @mpacer who talked with me about warnings and might be able to better explain this.

davidhalter:
Ok I just tested it non-interactively as well, it really looks like it makes more sense.

PR #922, opened 2017-05-06
Fix #919 by preventing unecessary call to repr() from within inspect.

### LINKED ISSUE #919: Slow completion for existing object with long `__repr__` method
Hi,

For completion with existing objects it seems that the `__repr__()` method is evaluated for an error message within the inspect module which is afterward caught. This makes the completion as slow as `__repr__()` while this evaluation in unnecessary for the user. This method can bee rather slow with big pandas objects or custom class.

This as been seen here :
[qtconsole issue #90](https://github.com/jupyter/qtconsole/issues/90)

A minimal code to reproduce it :
```python
import jedi
class Bugger(object):
    def __init__(self, size=10000):
        """Create bigg data."""
        self.big_data = [list(range(i)) for i in range(size)]

    def easy_method(self):
        """Method that should be really fast."""
        return self.big_data[-1][-1]

    def __repr__(self):
        output = ""
        for nested in self.big_data:
            for elem in nested:
                output += str(elem)
            output += '\n'
        return output

test = Bugger()
jedi.Interpreter('test.ea', [locals()]).completions()
```

And a more convenient working version :
```python
import jedi, pdb, inspect
class Bugger(object):
    def __init__(self, size=10):
        """Create bigg data."""
        self.big_data = [list(range(i)) for i in range(size)]

    def easy_method(self):
        """Method that should be really fast."""
        return self.big_data[-1][-1]

    def __repr__(self):
        frames = inspect.getouterframes(inspect.currentframe())
        for frame in frames:
            print(frame)
        pdb.set_trace()
        output = ""
        for nested in self.big_data:
            for elem in nested:
                output += str(elem)
            output += '\n'
        return output

test = Bugger()
jedi.Interpreter('test.ea', [locals()]).completions()
```

Jedi seems to end here through jedi/evaluate/compiled/mixed.py, line 121.
Maybe making checks before the calling `inspect.getsourcefile()` could fix it, but not I don't know jedi enough to claim it.

ElieGouzien:
Hi, this should fix the performance issue #919.
Please let me know if you want me to change anything.

PS : works great with IPython 6+.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -43,6 +43,7 @@ bstaint (@bstaint)
 Mathias Rav (@Mortal) <rav@cs.au.dk>
 Daniel Fiterman (@dfit99) <fitermandaniel2@gmail.com>
 Simon Ruggier (@sruggier)
+Élie Gouzien (@ElieGouzien)
 
 
 Note: (@user) means a github user name.
--- a/jedi/evaluate/compiled/mixed.py
+++ b/jedi/evaluate/compiled/mixed.py
@@ -116,8 +116,21 @@ def _load_module(evaluator, path, python_object):
     return module
 
 
+def source_findable(python_object):
+    """Check if inspect.getfile has a chance to find the source."""
+    return (inspect.ismodule(python_object) or
+            inspect.isclass(python_object) or
+            inspect.ismethod(python_object) or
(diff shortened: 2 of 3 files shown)


davidhalter:
I like! Thanks man. Good test.

PR #941, opened 2017-07-12
Fix keyword docstring

micbou:
Jedi raises a `NotImplementedError` exception when asking for the docstring of any Python keyword. For instance, running this code with latest dev branch
```python
import jedi

jedi.Script('assert').completions()[0].docstring()
```
outputs
```python
  File "jedi\jedi\cache.py", line 119, in wrapper
    return dct[key]
KeyError: ((), frozenset({('fast', False)}))

  File "test.py", line 10, in <module>
    jedi.Script('assert').completions()[ 0 ].docstring()
  File "jedi\jedi\api\classes.py", line 472, in docstring
    return super(Completion, self).docstring(raw=raw, fast=fast)
  File "jedi\jedi\api\classes.py", line 248, in docstring
    return _Help(self._name).docstring(fast=fast, raw=raw)
  File "jedi\jedi\api\classes.py", line 723, in docstring
    for context in self._get_contexts(fast=fast):
  File "jedi\jedi\cache.py", line 121, in wrapper
  File "jedi\jedi\api\classes.py", line 713, in _get_contexts
    return self._name.infer()
  File "jedi\jedi\evaluate\filters.py", line 21, in infer
    raise NotImplementedError
NotImplementedError
```

This is fixed by implementing the `infer` method in `KeywordName` and by replacing the `docstr` property (which is used nowhere) with the `py__doc__` method in `Keyword`. `imitate_pydoc` only accepts a string so we make sure that we pass one and not a `KeywordName` object. Also, we strip the trailing `\n` characters from the docs.

I am not completely sure about the implementation so tell me if things should be done differently and I'll update the PR. In particular, I am wondering if `Keyword` should be a subclass of `Context` since they have `api_type` and `py__doc__` in common.

Added a test and replaced `raw_doc` by `docstring()` in the docstring tests as the former is deprecated in favor of the latter according to the docs.

### DIFF
--- a/jedi/api/keywords.py
+++ b/jedi/api/keywords.py
@@ -76,11 +76,15 @@ class KeywordName(AbstractNameDefinition):
     api_type = 'keyword'
 
     def __init__(self, evaluator, name):
+        self.evaluator = evaluator
         self.string_name = name
         self.parent_context = evaluator.BUILTINS
 
     def eval(self):
         return set()
+
+    def infer(self):
+        return [Keyword(self.evaluator, self.string_name, (0, 0))]
 
 
 class Keyword(object):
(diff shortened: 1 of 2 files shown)


micbou:
>  If you're interested in diving into the details of if Keywords could be Contexts, I would very much appreciate it (if it's not possible, please also tell me).

I'll look into it.

Thanks for merging.

PR #942, opened 2017-07-13
Fix instance docstring

micbou:
Jedi always returns an empty docstring for an instance. For example, the following code:
```python
import jedi

source = """
class TestClass():
    '''Docstring of `TestClass`.'''
tc = TestClass()
tc"""

definitions = jedi.Script(source).goto_definitions()
print(definitions[0].docstring())
```
returns nothing while it should output ``Docstring of `TestClass`.``. It's a regression introduced by commit 7ca62578e11031ae09a6ad79fd74089de25ed46d. 

This PR fixes the regression and add two docstring tests: one with a class and another with an instance.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/942)

### DIFF
--- a/jedi/evaluate/instance.py
+++ b/jedi/evaluate/instance.py
@@ -197,6 +197,11 @@ class CompiledInstance(AbstractInstanceContext):
 
 
 class TreeInstance(AbstractInstanceContext):
+    def __init__(self, evaluator, parent_context, class_context, var_args):
+        super(TreeInstance, self).__init__(evaluator, parent_context,
+                                           class_context, var_args)
+        self.tree_node = class_context.tree_node
+
     @property
     def name(self):
         return filters.ContextName(self, self.class_context.name.tree_name)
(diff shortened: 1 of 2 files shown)


jedi/evaluate/instance.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, evaluator, parent_context, class_context, var_args):
         super(AbstractInstanceContext, self).__init__(evaluator, parent_context)
         # Generated instances are classes that are just generated by self
         # (No var_args) used.
+        self.tree_node = class_context.tree_node
```
davidhalter: Hmm, thanks for the test and fix. However I'm not sure I like this here. If anything I would have probably moved that assignment to `jedi.evaluate.instance.TreeInstance`. Would that make sense for you?

micbou: Yes, I moved it to `TreeInstance`.

davidhalter:
Thanks man, good stuff!

Issue #946, opened 2017-07-24
Support for python2 type comments


ilinum:
Currently, jedi only supports type hints for functions using python3 syntax.

It would be great to support type comments, which is widely used with python2.

```py
def foo(s):
    # type: (str) -> bool
    return s is not None and s.startswith('hello')
```

PR #947, opened 2017-07-25
Function annotations as comments, for type support in Python 2.

ldanilek:
Addressing issue 946 (https://github.com/davidhalter/jedi/issues/946).
Enable one-line function type annotations in comments, per https://www.python.org/dev/peps/pep-0484/.

These annotations determine the types of variables which are parameters or return values of functions, allowing better code navigation to instance methods.

### DIFF
--- a/jedi/evaluate/pep0484.py
+++ b/jedi/evaluate/pep0484.py
@@ -83,6 +83,50 @@ def _fix_forward_reference(context, node):
 @memoize_default()
 def infer_param(execution_context, param):
     annotation = param.annotation
+    if annotation is None:
+        # If no Python 3-style annotation, look for a Python 2-style comment
+        # annotation.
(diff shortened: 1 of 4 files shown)


jedi/evaluate/pep0484.py
```diff
@@ -82,7 +82,58 @@ def _fix_forward_reference(context, node):
 
 @memoize_default()
 def infer_param(execution_context, param):
```
davidhalter: Hmm, I think you're doing too much work here for no gain. I think this should be done with the parser. The parser can do a lot of cool stuff. By parsing something like `parse('foo(%s): pass' % regex_string)`, you could really easy validate a lot of things that you assume here or test with complicated param matching.

Wilfred:
Now that #1002 is merged, I think this can be closed.

PR #948, opened 2017-07-26
Fix get_line_code method.

gcmt:
Problem: `get_line_code` returns the wrong line of code.

Fix: make sure line numbers start at 0 when used as list indices. Also, make sure to prevent negative indices.

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -391,9 +391,9 @@ class BaseDefinition(object):
         path = self._name.get_root_context().py__file__()
         lines = parser_cache[path].lines
 
-        line_nr = self._name.start_pos[0]
-        start_line_nr = line_nr - before
-        return ''.join(lines[start_line_nr:line_nr + after + 1])
+        start = self._name.start_pos[0] - 1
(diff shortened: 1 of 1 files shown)


jedi/api/classes.py
```diff
@@ -391,9 +391,9 @@ def get_line_code(self, before=0, after=0):
         path = self._name.get_root_context().py__file__()
         lines = parser_cache[path].lines
 
-        line_nr = self._name.start_pos[0]
-        start_line_nr = line_nr - before
-        return ''.join(lines[start_line_nr:line_nr + after + 1])
+        start = self._name.start_pos[0] - 1
+        start = 0 if start - before < 0 else start - before
+        return ''.join(lines[start:start + after + 1])
```
davidhalter: I think there's a small mistake here. It should be more like 

```
middle = self._name.start_pos[0] - 1
start = max(middle - before, 0)
''.join(lines[start:middle + after + 1])
```

I think to avoid this from happening again, write some tests. If you did it would have likely fixed this in the beginning :)

PR #965, opened 2017-10-01
Add support to pyenv virtual env

lamenezes:
Since I started using pyenv + virtualenv jedi-vim stopped autocompleting any libs in my virtualenv. After some time analyzing the issue I discovered jedi doesn't work correctly with pyenv.

When running `:JediDebugInfo` I get the following (some lines hidden/altered for brevity):

```                                                                                                                                                                                                                       
 - sys.version: `3.5.3 (default, Jan 19 2017, 14:11:04), [GCC 6.3.0 20170118]`                                                                                                                                                                
 ...                                                                                                                                                                   
 - version: 0.10.2                                                                                                                                                                                                                            
 - sys_path:                                                                                                                                                                                                                                  
    - `~/.pyenv/versions/3.6.1/envs/jedi/lib/python3.5/site-packages`                                                                                                                                                                
    - `~/.pyenv/versions/3.6.1/envs/jedi/lib/python3.5/site-packages`
 ...
```

But the actual python path is different:

```
    - `~/.pyenv/versions/3.6.1/envs/jedi/lib/python3.6/site-packages`
```

That happens because jedi is using the system version (`3.5.3`) and not the pyenv version (`3.6.1`) when building virtualenv's sys paths. This PR extracts the pyenv version from the `PYENV_VIRTUALENV` and `PYENV_VERSION` environment variables.

It's important to note that this PR doesn't add _full_ pyenv support. This happens because there other ways to set the python version using pyenv. Such as creating a local `.python_version` file containing the desired version. In fact I believe it was because of this the PR #242 used subprocess to get the version from `pyenv version` command instead of using env vars.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -44,6 +44,7 @@ Mathias Rav (@Mortal) <rav@cs.au.dk>
 Daniel Fiterman (@dfit99) <fitermandaniel2@gmail.com>
 Simon Ruggier (@sruggier)
 Élie Gouzien (@ElieGouzien)
+Luiz Menezes (@lamenezes) <luiz.menezesf@gmail.com>
 
 
 Note: (@user) means a github user name.
--- a/jedi/evaluate/__init__.py
+++ b/jedi/evaluate/__init__.py
@@ -84,6 +84,7 @@ from jedi.evaluate.context import ClassContext, FunctionContext, \
 from jedi.evaluate.context.iterable import CompForContext
 from jedi.evaluate.syntax_tree import eval_trailer, eval_expr_stmt, \
     eval_node, check_tuple_assignments
+from jedi.evaluate.sys_path import get_python_version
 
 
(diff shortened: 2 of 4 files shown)


davidhalter:
I'm actually working on virtualenv's right now. I'll keep this one open.

The problem with your solution is basically that it uses another Python version for the virtualenv than it was created with. This leads potentially to segfaults in binary modules.

My approach is to just move all of that stuff out to a subprocess.

davidhalter:
Hi @lamenezes 
The current master branch works very well with virtualenvs. There is a function called `jedi.api.environment.get_default_environment`. Could you check if it does the correct thing for pyenv?

The old logic doesn't exist anymore. There's really a lot of new things.

davidhalter:
Closing, because the code that you modified doesn't exist anymore. Let me know if it still doesn't work ad we find a way. I hope that these virtualenvs just work, now.

PR #971, opened 2017-10-23
FIX: install on python 3.7

tacaswell:
https://github.com/python/cpython/pull/46 /
https://bugs.python.org/issue29463 move the module comment into the
AST node and hence out of the tree which means the 2nd entry in the
tree is now the import rather than the `__version__` string.

Adds nightly on travis.

Please let me know if you need anything else from me!

PS: I :heart: jedi


tacaswell:
I don't know tox well enough to quickly convince it to install the parso master branch for the nightly entry in the build matrix :disappointed:

davidhalter:
Thanks, I'll wait for the next parso release (to get the 3.7 tests to pass). I'll merge then.

davidhalter:
parso 0.11.1 has been released. Can you just remove the nightly and use 3.7-dev and remove it from the allowed-failures? Thanks :)

tacaswell:
Fixed, but it looks like there are a couple of real failures :(

davidhalter:
Thanks! I think now we would just need to figure out why the fails happen on travis. That's a bit more complicated and it would probably involve me installing Python nightlies. I'm too lazy for that. Maybe we should just leave it in the allowed-failures?!

The failures don't really matter IMO. I think we should just get it to pass (don't mind if it's one more commit, I will just squash).

tacaswell:
I have a super hacky bash script I use to build master branch version of everything in the scipy stack, this is the chunk that I think is useful to you

```bash
TARGET_ENV=bleeding

OSPATH=~/source/other_source/

pushd $OSPATH/cpython/
git pull
git clean -xfd
./configure
make -j 8
./python -m venv --copies --clear ~/.virtualenvs/$TARGET_ENV
popd

source ~/.virtualenvs/$TARGET_ENV/bin/activate

echo $PATH


pull_build () {
    git pull
    git clean -xfd
    pip install -ve .
}


master_build () {
    git checkout master
    git pull
    git clean -xfd
    pip install -v .
}


pushd $OSPATH/parso
master_build
popd
pushd $OSPATH/jedi
pull_build
popd
```

davidhalter:
What does that help with? Sorry I don't really understand that. Thanks for the changes though!

tacaswell:
boot-straps a virtual env with the python master branch in it.  I guess it is more compelling if you are installing 15 things....

carlwgeorge:
I'm getting build errors on Fedora now, and I think it's due to the fact that the [AST change was reverted in 3.7.0b5](https://github.com/python/cpython/pull/7121).  Can this PR be reverted to match?

hroncok:
Similar change in pytest: https://github.com/pytest-dev/pytest/pull/3531

carlwgeorge:
On second thought, just reverting this would undo the classifiers and travis changes as well.  I'll pop a new pull request to just undo the necessary parts.

PR #978, opened 2017-10-28
Docstring for evaluate_call_of_leaf

langsamer:
Hi David,

I removed the TODO about the optional parameter 'cut_own_trailer' because it really is needed. In its place I wrote an explanation when this parameter must be set to True.  (... which I hope is correct and clear)
The two commits in thist PR are identical in content; they stem from my difficulty grasping how forking and upstream branches work.  (Even that lesson made the sprint worth while for me ;-) )

Cheers,
Malte


davidhalter:
I just squash merged it :)

Thanks for your efforts Malte!

PR #981, opened 2017-10-30
Fix dead link to Monty Python Jedi.

techdragon:
Hopefully this is close enough to the original video.


davidhalter:
Haha, thanks for catching this one!

Issue #974, opened 2017-10-26
PEP 526 - Ahead of time annotations


InfernoZeus:
It seems that jedi doesn't properly support PEP 526 annotations when they're used ahead of time. This example should demonstrate the problem:

```python
    test_string: str = "Hello, world!"
    test_string.<tab> # Completes as you would expect: .capitalize, etc

    char: str
    for char in test_string:
        char.<tab> # Does not complete properly
```

I suspect it's closely related to #933.

Relevant section of PEP 526:

> Only single assignment targets and single right hand side values are allowed. In addition, one cannot annotate variables used in a for or with statement; they can be annotated ahead of time, in a similar manner to tuple unpacking:
>```python
>a: int
>for a in my_iter:
>    ...
>
>f: MyFile
>with myfunc() as f:
>    ...
>```

davidhalter:
I'm pretty sure we don't support this use case, yet.

johannesmik:
Your testcase worked for me, when I tested it with Jedi 0.11 and Python 3.5:

```
import jedi

source = """
teststring = "Hello, World!"
char : str
for char in teststring:
    char."""

jedi_script = jedi.Script(source=source)
print("Completions:", jedi_script.completions())
print("Jedi Version:", jedi.__version__)
```
 got the following output:

```
Completions: [<Completion: capitalize>, <Completion: casefold>, <Completion: center>, <Completion: count>, <Completion: encode>, ....... ]
Jedi Version: 0.11.0
```

I also got it working with Vim and YouCompleteMe (Ycm), after updating Ycm.

johannesmik:
Yet, it looks like that Jedi is ignoring the "ahead of time annotation", and tries to infer the type in a different way.

For example:
```python
somelist = [1, 2, 3, "A", "A"]

element : int
for element in somelist[0:3]:
    element.<TAB> # Gives both int and string suggestions (bit_length, capitalize, ...)
```

davidhalter:
We were discussing this issue at the pycon.de sprints and I guess @johannesmik and me figured out what the problem was. There's a pull request incoming.

InfernoZeus:
@davidhalter @johannesmik Thanks to both of you! Really impressed by the quick fix 😃

davidhalter:
Now you might be less impressed on how long it took to merge the request. I'm still trying to cleanup some things. Not really sure it works already.

davidhalter:
[...]

PR #982, opened 2017-10-31
Including ahead of time annotations 

johannesmik:
In response to issue #974 we worked on including ahead of time annotations together at the Pycon.de sprint.

While the 5 tests I defined all pass, it seems like the code breaks some existing tests. I got **100+ failures** running `py.test test/test_integration.py`. Could this be a python version issue?


davidhalter:
Ok. I think the first step we have to take is to remove the issue where `get_root_context().tree_node` is None. Just skip it in that case. This basically means that it's a builtin (something like `str`, `dict`, etc). Once that's done let's look at the remaining tests.

davidhalter:
I just did it for you, hehe :)

Thanks for being part of the sprints, it was great fun!

Issue #985, opened 2017-11-01
Support all async constructs of python 3.6


robinro:
So far not (fully) working:
* [ ] async generators:
```
async def foo():
    yield 1
    await asyncio.sleep(0)
    yield 2
```
  * `foo()` should be async-generator
  * `for x in foo()` should give NO generator
  * `async for x in foo()`: should make x an int
  * ensure support for `asend`, see [PEP 525](https://www.python.org/dev/peps/pep-0525/#asynchronous-generator-object)
* [ ] async comprehension (list, set, dict, generator) `[x async for x in foo()]`
* [ ] `async with` context managers

Maybe more, see [PEP 492](https://www.python.org/dev/peps/pep-0492/)

robinro:
First step: #986.
I will probably look into getting more of this done, but I'm also happy if someone else implements it :)

davidhalter:
Thanks for your work again. I think I've gotten all of the remaining constructs working. Please test and let me know if there's still issues. It's on master.

PR #986, opened 2017-11-01
[WIP] Support async/await syntax

robinro:
Improve support for async/await syntax.

This is still work in progress. 

TODO:
* [x] support for coroutines (async def): `async def a(): pass`
* [x] support for calling coroutines (await): `a` is function, `a()` is coroutine, `await a()` is return type
* [x] rough support for async generators, only working usecase so far:
```
async def foo():
    yield 1
a = [x async for x in foo()]
```
* [ ] correctly deal with async generators (return awaitable/async_generator_asend instead of values)
* [ ] figure out how to parse async/await types from python < 3.5/3.6, the `types` module only has the right types in 3.5 and 3.6; or at least give a better error than the current `ImportError`
* [ ] refactor coroutine/asyncgenerator to reuse code of generator/...; maybe move `Coroutine` to `function.py`

Fully supporting all async syntax is outside the scope of this PR. I opened an issue to track the progress: #985

### DIFF
--- a/jedi/evaluate/base_context.py
+++ b/jedi/evaluate/base_context.py
@@ -63,10 +63,13 @@ class Context(BaseContext):
         arguments = ValuesArguments([ContextSet(value) for value in value_list])
         return self.execute(arguments)
 
-    def iterate(self, contextualized_node=None):
+    def iterate(self, contextualized_node=None, is_async=False):
         debug.dbg('iterate')
         try:
-            iter_method = self.py__iter__
+            if is_async:
+                iter_method = self.py__aiter__
+            else:
+                iter_method = self.py__iter__
         except AttributeError:
             if contextualized_node is not None:
                 from jedi.evaluate import analysis
@@ -241,9 +244,9 @@ class ContextSet(BaseContextSet):
     def py__class__(self):
         return ContextSet.from_iterable(c.py__class__() for c in self._set)
 
-    def iterate(self, contextualized_node=None):
+    def iterate(self, contextualized_node=None, is_async=False):
         from jedi.evaluate.lazy_context import get_merged_lazy_context
-        type_iters = [c.iterate(contextualized_node) for c in self._set]
+        type_iters = [c.iterate(contextualized_node, is_async=is_async) for c in self._set]
         for lazy_contexts in zip_longest(*type_iters):
             yield get_merged_lazy_context(
                 [l for l in lazy_contexts if l is not None]
--- a/jedi/evaluate/context/function.py
+++ b/jedi/evaluate/context/function.py
@@ -63,11 +63,19 @@ class FunctionContext(use_metaclass(CachedMetaClass, TreeContext)):
         """
         Created to be used by inheritance.
         """
-        yield_exprs = get_yield_exprs(self.evaluator, self.tree_node)
-        if yield_exprs:
-            return ContextSet(iterable.Generator(self.evaluator, function_execution))
+        is_coroutine = self.tree_node.parent.type == 'async_stmt'
+        is_generator = bool(get_yield_exprs(self.evaluator, self.tree_node))
+
+        if is_coroutine:
+            if is_generator:
+                return ContextSet(iterable.AsyncGenerator(self.evaluator, function_execution))
+            else:
+                return ContextSet(iterable.Coroutine(self.evaluator, function_execution))
         else:
-            return function_execution.get_return_values()
+            if is_generator:
+                return ContextSet(iterable.Generator(self.evaluator, function_execution))
+            else:
+                return function_execution.get_return_values()
 
     def get_function_execution(self, arguments=None):
         if arguments is None:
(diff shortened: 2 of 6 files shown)


robinro:
To work this needs davidhalter/parso#19

jedi/evaluate/context/iterable.py
```diff
@@ -54,6 +67,83 @@ def name(self):
...
+
+    @property
+    def name(self):
+        return compiled.CompiledContextName(self, 'coroutine')
+
+
+class Coroutine(CoroutineMixin, Context):
+    def __init__(self, evaluator, func_execution_context):
+        if not HAS_COROUTINE:
+            raise ImportError("Need python3.5 to support coroutines.")
```
davidhalter: I don't think we need to check for this. At the moment jedi is using the same version for autocompletion in the parser and type inference. It should therefor never be an issue.

davidhalter: I'm actually working to get cross-Python support working for a lot of things. However that is an entirely different issue that you shouldn't care about. Just use the mindset that you're writing this for Python 3.6 and I'll help you to get the tests to pass. Just get the tests working for 3.6 :)

jedi/evaluate/base_context.py
```diff
@@ -63,10 +63,13 @@ def execute_evaluated(self, *value_list):
         arguments = ValuesArguments([ContextSet(value) for value in value_list])
         return self.execute(arguments)
 
-    def iterate(self, contextualized_node=None):
+    def iterate(self, contextualized_node=None, is_async=False):
         debug.dbg('iterate')
         try:
-            iter_method = self.py__iter__
+            if is_async:
+                iter_method = self.py__aiter__
```
davidhalter: Niiiice :) I like!

jedi/evaluate/context/function.py
```diff
@@ -63,11 +63,19 @@ def infer_function_execution(self, function_execution):
...
         """
-        yield_exprs = get_yield_exprs(self.evaluator, self.tree_node)
-        if yield_exprs:
-            return ContextSet(iterable.Generator(self.evaluator, function_execution))
+        is_coroutine = self.tree_node.parent.type == 'async_stmt'
+        is_generator = bool(get_yield_exprs(self.evaluator, self.tree_node))
+
+        if is_coroutine:
+            if is_generator:
+                return ContextSet(iterable.AsyncGenerator(self.evaluator, function_execution))
```
davidhalter: I think here we have to check if it's Python 3.5. This basically means that you check for `if evaluator.python_version == (3, 5): return ContextSet()` (+ add a warning).

If we don't do that we end up with an AsyncGenerator in 3.5 that doesn't really work in that version.

jedi/evaluate/context/iterable.py
robinro: Should I remove all version checks / ImportErrors or how do we proceed?

robinro:
For the async generator I'm not sure how to proceed. Somehow the yield needs to return an "awaitable" object that is either awaited explicitly `await my_async_gen().__anext__()` or implicitly `[res async for res in my_async_gen()]`, `async for res in my_async_gen(): print(res)`

robinro:
Maybe it's better to split of the (done) coroutine part and spend more time on the async generator to get it right?

jedi/evaluate/context/iterable.py
davidhalter: Yes just remove them. :)

davidhalter:
> Maybe it's better to split of the (done) coroutine part and spend more time on the async generator to get it right?

I don't care. We can do it either way. I think one of the problems we have is that we call execute_await and we're not searching for `__await__` methods there. This would certainly be possible, but it's a little bit of additional effort. In a first step we should probably rename `execute_await` to `py__await__` call that one on the objects created by `__aiter__`

And now that I've written that I'm actually not sure if that makes a lot of sense. I have really not understood all of PEP 525. I just think that we need a better way of defining awaitables. `py__await__` would probably be the way to go.

In case of the generator I think we just need to create a `AsyncGeneratorNext` object or something like the `async_generator_asend` object below that has a `py__await__` method.

```
>>> async def genfunc():
...     yield 1
...     yield 2
... 
>>> 
>>> gen = genfunc()
>>> 
>>> 
>>> assert gen.__aiter__() is gen
>>> x = gen.__anext__()
>>> x
<async_generator_asend object at 0x7f5a92b58808>
```

Do you understand what I'm proposing? We can also talk about it on skype or something if that's better.

davidhalter:
I merged it locally! Thanks for all the work. It's good stuff. The things I fixed had mostly to do with different Python versions and a better abstraction to mix builtin objects with Contexts.

I think async/await should now be completable, pretty much. One thing that is probably still a bit buggy is `async with`, but that can also be fixed at a later point. Most people will probably not use that a lot anyway.

The other thing I did not do (yet), is to implement `__anext__` properly. This doesn't really matter for most autocompletion related things, but it's a bit ugly and might make some autocompletion results worse. Then again this is not properly implemented for normal iterators either and nobody has complained, yet. :)

PR #1002, opened 2017-11-24
Support mypy annotations using Python 2 comment syntax

### LINKED ISSUE #946: Support for python2 type comments

Wilfred:
This allows us to use mypy annotations for completion in Python 2.

This is based on #947, rebased on the latest master. I've also done my best to resolve @davidhalter's feedback on that PR. I've replaced the regexp parsing with a call to `parso.parse`, and moved the tests as suggested.

I haven't resolved the comment on `_fix_forward_reference` -- I'm not sure what's wrong with that function that needs fixing.

Closes #946.

### DIFF
--- a/jedi/evaluate/pep0484.py
+++ b/jedi/evaluate/pep0484.py
@@ -22,7 +22,7 @@ x support for type hint comments for functions, `# type: (int, str) -> int`.
 import os
 import re
 
-from parso import ParserSyntaxError
+from parso import ParserSyntaxError, parse
 from parso.python import tree
 
 from jedi._compatibility import unicode, force_unicode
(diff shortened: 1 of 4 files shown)


davidhalter:
Thanks man! There's quite a few issues in the tests. Can you maybe rebase on master and try to fix them.

Sorry for not replying sooner. But I was pretty busy :/ The general approach definitely makes sense and I'd be happy to review the remaining things again once we're sure that it's working :)

Wilfred:
Rebased. Almost all of the tests are passing, except Python 2.7 on JEDI_TEST_ENVIRONMENT=35 and 36:

```python
ValueError: malformed node or string: b"u'str'"
```

It's not clear to me what's going on here.

jedi/evaluate/pep0484.py
```diff
@@ -80,9 +80,66 @@ def _fix_forward_reference(context, node):
...
+def _split_mypy_param_declaration(decl_text):
+    """
+    Split decl_text on commas, but group generic expressions
+    together.
+
+    For example, given "foo, Bar[baz, biz]" we return
+    ['foo', 'Bar[baz, biz]'].
+
+    """
+    node = parse(decl_text).children[0]
```
davidhalter: Here we need to handle the error cases. Please add a test with a comment that is invalid. You can either just catch the parser error and return `[]` or use error_recovery=True and then handle all the issues.

jedi/evaluate/pep0484.py
```diff
@@ -102,7 +159,26 @@ def py__annotations__(funcdef):
...
+            return NO_CONTEXTS
+
+        match = re.match(r"^#\s*type:\s*\([^#]*\)\s*->\s*([^#]*)", comment)
+        if not match:
+            return NO_CONTEXTS
+
+        annotation = tree.String(
+            repr(str(match.group(1).strip())),
+            node.start_pos)
+        annotation.parent = node.parent
```
davidhalter: IMO we shoud not actually need to create a string object. We probably have to create a function similar to ` _evaluate_for_annotation` that can just take a string argument and use that to resolve comment annotations.

I know that it's already used once, so it's not your fault and you don't have to change it. I'll merge anyway if you don't do it in the next few days. If you don't do it I will :)

test/completion/pep0484_comments.py
```diff
@@ -103,3 +103,67 @@ class Employee:
...
+class Cat(object):
+    def __init__(self, age, friends, name):
+        # type: (int, List[Dog], str) -> None
+        self.age = age
+        self.friends = friends
+        self.name = name
+
+cat = Cat(UNKNOWN_NAME4, UNKNOWN_NAME5, UNKNOWN_NAME6)
+#? str()
+cat.name
```
davidhalter: Great set of tests! Thanks!

Wilfred:
Thanks for the review! :)

I've added tests for syntax errors in type comments. These seem to just pass, so it looks like parse errors are already handled somewhere else in jedi.

I've also made some minor updates to the docs to say that function annotations are now supported.

davidhalter:
I merged a little bit before you did the changes and then did some of my changes that I thought were necessary.

I will close because the changes have been merged (You force pushed a bit after). I think I have written a bit of a different test than you, but that shouldn't matter.

Thanks again for your effort. Good job! :) Happy to merge more of your contributions :)

Wilfred:
Thank you :)

PR #1004, opened 2017-11-29
Improve IntegrationTestCase.__repr__

blueyed:
Having the path (together with the line only) makes it easy to go to the
actual test.


davidhalter:
Makes sense!

Issue #995, opened 2017-11-07
"Goto" prefers local module instead of global one


blueyed:
Given the following project setup, in a virtualenv where stripe is installed (`pip install stripe`):

```
.
├── proj
│   └── app
│       ├── __init__.py
│       ├── models.py
│       └── stripe
│           └── __init__.py
```

With `models.py`:
```
import stripe

stripe.error
```

Jedi-vim's `goto` will use the project's module instead of the global one.

It behaves as if `from . import stripe` would be used instead of `import stripe`.

Python 3.6.3, latest parso and jedi(-vim).

The same applies when moving the `__init__.py` from `app` to `proj`.

blueyed:
This fixes it:

```
diff --git i/jedi/evaluate/imports.py w/jedi/evaluate/imports.py
index ecf656b1..89dd833f 100644
--- i/jedi/evaluate/imports.py
+++ w/jedi/evaluate/imports.py
@@ -266,7 +266,7 @@ def sys_path_with_modifications(self):
             # Since we know nothing about the call location of the sys.path,
             # it's a possibility that the current directory is the origin of
             # the Python execution.
-            sys_path_mod.insert(0, os.path.dirname(self.file_path))
+            sys_path_mod.append(os.path.dirname(self.file_path))
 
         return in_path + sys_path_mod
 ```

blueyed:
The patch does not cause any test failures, but would need one.

@davidhalter 
I guess it's easier for you to come up with one?
Should go to `test/completion/imports.py` probably, wrapped in some temporary dir changing?! (going to `pkg`, where the upper `mod1` should get preferred then)

kiddick:
Very annoying behavior; It need to be fixed

blueyed:
@kiddick 
Agreed.
The fix is there - just needs a test I guess.
Feel free to chime in, otherwise we have to wait for help from @davidhalter.

kiddick:
May be it is wirh to make PR with described changes?

blueyed:
Done: https://github.com/davidhalter/jedi/pull/1011.

davidhalter:
Just regarding the test: It's probably not that easy to write one. You will probably have to create the whole tree that you described above within Jedi and then test from there.

PR #1011, opened 2017-12-19
sys_path_with_modifications: append local file

blueyed:
This fixes "goto" preferring a local module instead of a global one.

Fixes https://github.com/davidhalter/jedi/issues/995.


davidhalter:
I think this is fine since it doesn't break any tests. Note however that in the virtualenv branch there are a lot of changes for sys path coming. So without a test this one might fail again.

blueyed:
Yeah, see https://github.com/davidhalter/jedi/issues/995#issuecomment-347848687.
Some advice would be nice, or just adding it yourself if you get the chance.

blueyed:
Thanks for merging already!

PR #1017, opened 2018-01-04
Drop support for EOL Python 2.6 and 3.3

### LINKED ISSUE #1016: Drop support for EOL Python 2.6 and 3.3?
It looks like significant effort is going into maintaining support for EOL Python versions. When they reach EOL that means they're no longer receiving security updates.

image

https://en.wikipedia.org/wiki/CPython#Version_history

Here's the pip installs for jedi from PyPI for the last month (via `pypinfo --percent --pip --markdown jedi pyversion`).

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 3.6            |   44.4% |        155,231 |
| 3.5            |   27.5% |         96,124 |
| 3.4            |   22.2% |         77,788 |
| 2.7            |    5.2% |         18,142 |
| 3.7            |    0.4% |          1,559 |
| 3.3            |    0.2% |            800 |
| 2.6            |    0.0% |             30 |
| 3.2            |    0.0% |              5 |

From these numbers, it looks like 2.6 and 3.3 can be safely dropped.

This would allow the code to be modernised and ease the maintenance burden quite a bit.

Thoughts?

hugovk:
Fixes #1016.

Also add `python_requires` in setup.py to help pip, and some other code cleanup.


davidhalter:
Wow. This is awesome. I'll be thinking about even dropping 3.3 then.

Since most of the changes are related to Python 2.6, would you be able to easily separate those? I'm 100% sure to drop Python 2.6 support and most of the things that are changing, are changing because of Python 2.6.

hugovk:
Yes, should be fine to separate the 2.6 things. Do you want those in their own PR, or just their own commit?

davidhalter:
PR would be nice. You can probably just cherry-pick the commits above (and maybe do one or two minor changes).

If possible also make a PR for 3.3. That can also depend on the other branch, but I will just wait with merging that for another few months. Thanks a lot!

hugovk:
Done, I've split this into #1018 and #1019.

PR #1019, opened 2018-01-05
[postponed] Drop support for EOL Python (2.6 and) 3.3

### LINKED ISSUE #1016: Drop support for EOL Python 2.6 and 3.3?

hugovk:
Edit: rebased on master, so there's just the one commit needed here.


hugovk:
Travis CI failure is a temporary network issue, it passed on my fork: https://travis-ci.org/hugovk/jedi/builds/325357381.

hugovk:
Rebased on master.

tox.ini
```diff
@@ -30,30 +29,20 @@ commands =
...
-    numpydoc
     {[testenv]deps}
 [testenv:py36]
 deps =
-    numpydoc
     {[testenv]deps}
 [testenv:cov]
 deps =
     coverage
-    numpydoc
```
blueyed: Why is numpydoc removed here (and above)?

hugovk: Oops, that'll be a mistake from rebasing. I'll revert it!

davidhalter: Thanks for finding this one!

davidhalter:
Thanks for the rebase. I merged it finally. This means that the next version will not support Python 3.3.

PR #1020, opened 2018-01-05
http -> https

hugovk:


davidhalter:
Looks good and thanks for also checking the URLs :)

PR #1022, opened 2018-01-05
Made 'l' a string + added warning

MohamedAlFahim:
One of the helper methods is missing, so be extra careful.


davidhalter:
Thanks!

By the way the refactoring piece of code hasn't been in use for a very long time. It's still a bit broken.

Thanks for the review anyway, but some of the code is probably not going to be reused. The rest of the code however could happily use some changes :)

PR #1026, opened 2018-01-12
Fix AttributeError when incorrectly using generators with variable-length arguments

micbou:
Jedi raises the exception
```python
 File "jedi\evaluate\arguments.py", line 200, in get_calling_nodes
    return [arguments.argument_node or arguments.trailer]
AttributeError: 'ValuesArguments' object has no attribute 'argument_node'
```
when getting the definition of `function_var_args` in the following code
```python
def decorator_var_args(function, *args):
    return function(*args)

@decorator_var_args
def function_var_args(param):
    return param
```
This error was initially reported in https://github.com/vheon/JediHTTP/issues/51 and reduced to the code above.

While I don't fully understand the involved code in Jedi, I think the proposed fix makes sense (defining `argument_node` and `trailer` properties in the base class `AbstractArguments`) as `arguments` can be any object derived from `AbstractArguments` (`ValuesArguments` in that case) and not necessarily `TreeArguments`.

Added a test that fails without the fix.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1026)

### DIFF
--- a/jedi/evaluate/arguments.py
+++ b/jedi/evaluate/arguments.py
@@ -29,6 +29,8 @@ def try_iter_content(types, depth=0):
 
 class AbstractArguments(object):
     context = None
+    argument_node = None
+    trailer = None
 
     def eval_argument_clinic(self, parameters):
         """Uses a list with argument clinic information (see PEP 436)."""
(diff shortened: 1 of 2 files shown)


davidhalter:
Wouldn't it be a better solution if you just returned `[]` (instead of `break`) on line 177? I think that is the issue.

micbou:
Right. I updated the PR. Just one thing. [There is another place where the `argument_node` and `trailer` properties are used](https://github.com/davidhalter/jedi/blob/master/jedi/evaluate/context/instance.py#L421-L427). Should we do something about it? It's going to raise the same error if `_var_args` is not a `TreeArguments` object.

davidhalter:
That is a very good point. I guess you were right after all. You can probably just reverse to what you proposed first.

micbou:
Done.

davidhalter:
Thanks for the good research!

Issue #1037, opened 2018-02-01
Current master branch not usable with Python 2


TobiasRzepka:
Hello,
with the current branch of jedi/parso I get always a stacktrace on Python 2!!! I believe it's since the merge of the virtualenv and master branch.
Here is my code (already known from a previous post):
```
s = jedi.Script("max_attr_value = int(2) # Cast to int for spe", 1, 45, "cmd")
s.completions()
```
With the resulting stacktrace:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\jedi\api\__init__.py", line 155, in completions
    self._evaluator, self._get_module(), self._code_lines,
  File "C:\Python27\lib\site-packages\jedi\api\__init__.py", line 137, in _get_module
    name = dotted_path_in_sys_path(self._evaluator.get_sys_path(), self.path)
  File "C:\Python27\lib\site-packages\jedi\evaluate\__init__.py", line 123, in get_sys_path
    return self.project._get_sys_path(self, environment=self.environment)
  File "C:\Python27\lib\site-packages\jedi\evaluate\cache.py", line 40, in wrapper
    rv = function(obj, *args, **kwargs)
  File "C:\Python27\lib\site-packages\jedi\api\project.py", line 96, in _get_sys_path
    sys_path = list(self._get_base_sys_path(environment))
  File "C:\Python27\lib\site-packages\jedi\api\project.py", line 83, in _get_base_sys_path
    sys_path = environment.get_sys_path()
  File "C:\Python27\lib\site-packages\jedi\cache.py", line 118, in wrapper
    result = method(self, *args, **kwargs)
  File "C:\Python27\lib\site-packages\jedi\api\environment.py", line 73, in get_sys_path
    return self._get_subprocess().get_sys_path()
  File "C:\Python27\lib\site-packages\jedi\evaluate\compiled\subprocess\__init__.py", line 162, in get_sys_path
    return self._send(None, functions.get_sys_path, (), {})
  File "C:\Python27\lib\site-packages\jedi\evaluate\compiled\subprocess\__init__.py", line 201, in _send
    is_exception, traceback, result = pickle_load(self._process.stdout)
  File "C:\Python27\lib\site-packages\jedi\_compatibility.py", line 384, in pickle_load
    return pickle.load(file)
  File "C:\Python27\lib\pickle.py", line 1384, in load
    return Unpickler(file).load()
[...]

PR #1040, opened 2018-02-11
Fix for Python 2 on Windows, see #1037.

TobiasRzepka:
Hello Dave,
I've found the spot and here's the fix, tested in Windows/Linux and Python 2.7/3.x combination.


TobiasRzepka:
This one was good enough for the special case mentioned in #1037 but not for all other real world cases :( But I'm on it.

davidhalter:
Awesome, thanks. I've quickly looked at it and it looks like it breaks some tests on linux.

However I thin you might be on to something.

PR #1046, opened 2018-02-17
Fix for #1037 and Python 3 issues on Windows.

TobiasRzepka:
Hello David,
this time it's looking better. All tox test run fine (at least 2 are broken on Windows but that are other issues). Main problem is the shell which consumes all control characters (below 32) and expand \n to \r\n. Not good, if binary data is transported. After that, if found out, that Python on Windows don't throw EPIPE and EOF errors for pipes. So I reraise them with the correct type / error number.
On Python 3.3 it was another special case, where the flush throws sometimes an error, where normally at the pickle dump it is thrown. So I shifted the flush also to _compatibility.py.

Tobias

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,4 +1,4 @@
-Main Authors
+﻿Main Authors
 ============
 
 David Halter (@davidhalter) <davidhalter88@gmail.com>
@@ -48,5 +48,6 @@ Robin Roth (@robinro)
 Malte Plath (@langsamer)
 Anton Zub (@zabulazza)
 Maksim Novikov (@m-novikov) <mnovikov.work@gmail.com>
+Tobias Rzepka (@TobiasRzepka)
 
 Note: (@user) means a github user name.
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -47,6 +47,5 @@ Simon Ruggier (@sruggier)
 Robin Roth (@robinro)
 Malte Plath (@langsamer)
 Anton Zub (@zabulazza)
-Tobias Rzepka (@TobiasRzepka)
 
 Note: (@user) means a github user name.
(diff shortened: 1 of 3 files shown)


davidhalter:
Hi Tobias

Awesome. Linux tests seem to be passing, too. Good job! Can you add the long comment you made above to the compatibility file where you changed the code? I think that would clear it up for most other users.

I agree that the tests that are broken on Windows should be fixed separately. Maybe I can fix that once I have access to a Windows machine, or you yourself. Ideally we would run it on travis, but that's not possible, unfortunately.

~ Dave

TobiasRzepka:
Good idea!  I added the comments to _compatibility.py.
If I find a minute, I'll looking on the broken test on windows...

davidhalter:
@TobiasRzepka I intend to merge this. While I don't like that we have to do it this ugly, I think it's well done from your side.

I will try to test it on a Windows computer quite soon and then merge.

TobiasRzepka:
I won't take it personal, but what did you mean with "we have to do it this ugly"? The conversion with binascii.hexlify? Then I totally agree, but all other attempts failed somewhere, e.g. other unicode to x conversions, because pickle don't create real unicode but binary data. Most of the tests pass, but because your nice tox setup I found always some special cases, which couldn't handled else... Changing to sockets might solve the problem, raising new ones maybe.

davidhalter:
Haha, I just mean the binascii.hexlify part ;)

You did obviously a great job. I really mean that. Not many people have the patience to fix these things. So thank you!

davidhalter:
Very awesome. I just got it working on Windows. Looks really a lot better. I'm still looking into some other fails. However most tests are already passing.

Can I maybe ping you in the future if I have issues with Windows? That would be really nice! :)

TobiasRzepka:
Yes of course. But currently I'm very short on time.

PR #1054, opened 2018-03-03
 jedi/api/environment.py - fix syntax error in find_virtualenvs

ggilmore:
`set.append(x)` isn't a method that `set` offers. I think that you want `set.add(x)`?


davidhalter:
Obviously. Thank you for the fix. I should write a few tests for that.

PR #1057, opened 2018-03-07
Fix VS Code Python extension link

orn688:
Microsoft has taken over management of the VS Code Python extension, so the extension's URL has changed.


davidhalter:
Thanks for noting!

@DonJayamanne Let me know if this is not ok for you or if there's something wrong and I'll revert.

DonJayamanne:
It's all good

PR #1064, opened 2018-03-22
Specify executable extension to detect virtual environment on Windows

micbou:
Jedi is unable to detect a virtual environment on Windows because it checks for the existence of the `Scripts\python` path, which doesn't exist since the extension is missing. We need to check for `Scripts\python.exe` on this platform.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1064)


davidhalter:
You're right. Thanks a lot!

We probably just need to add tests for virtualenvs.

PR #1065, opened 2018-03-22
Fix paths from assignment test on Windows

micbou:
The `test_paths_from_assignment` test fails on Windows because `_paths_from_assignment` normalizes the paths through `os.path.abspath` and `os.path.abspath( '/foo/' )` returns `X:\foo\` on Windows where `X` can be any drive letter, not `/foo/`. The expected paths in the test should be normalized as well.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1065)

### DIFF
--- a/test/test_evaluate/test_sys_path.py
+++ b/test/test_evaluate/test_sys_path.py
@@ -15,9 +15,14 @@ def test_paths_from_assignment(Script):
         expr_stmt = script._module_node.children[0]
         return set(sys_path._paths_from_assignment(script._get_module(), expr_stmt))
 
-    assert paths('sys.path[0:0] = ["a"]') == {'/foo/a'}
-    assert paths('sys.path = ["b", 1, x + 3, y, "c"]') == {'/foo/b', '/foo/c'}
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks! Could you add a comment there to clarify that this is done for Windows? Otherwise I will be tempted to remove it again :)

micbou:
Of course. Done.

davidhalter:
Thanks dude!

Can I maybe ping you in the future if I have issues with Windows? That would be really nice! :)

micbou:
> Can I maybe ping you in the future if I have issues with Windows? That would be really nice! :)

No problem. Speaking of which, I am wondering if you would be interested in running automatically the tests on [AppVeyor](https://ci.appveyor.com). If you don't know, it's a CI service like Travis but for Windows. It's relatively easy to set up. You just need to put a configuration file `appveyor.yml` at the root of the project, connect to the AppVeyor website with your GitHub account, and add the repository. If you think that's a good idea, I can write the configuration file and send a PR when it's ready (with probably other changes to make the tests pass).

davidhalter:
Yeah, that would be very awesome!

PR #1066, opened 2018-03-23
Improve venv_and_pths test

micbou:
The `venv_and_pths` test expects the Python executable to be in `/usr/bin`. That obviously can't be the case on Windows but it's also not necessarily the case on other platforms. For instance on Travis, the Python executable can be located in `/opt/python/x.y/bin`. To find the real Python path, we run the interpreter and print the value of `sys.base_prefix` or `sys.real_prefix` if we are in a virtualenv.

The `os.system` call was replaced with `subprocess.call` since we don't need to run the command in a shell.

Other changes are differences in the `venv` directory structure between Windows and UNIX.

The test should now pass on Windows and on all Travis builds (except for `JEDI_TEST_ENVIRONMENT=27`) instead of being skipped in most of them.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1066)

### DIFF
--- a/test/test_evaluate/test_sys_path.py
+++ b/test/test_evaluate/test_sys_path.py
@@ -2,6 +2,7 @@ import os
 from glob import glob
 import sys
 import shutil
+import subprocess
 
 import pytest
 from jedi.evaluate import sys_path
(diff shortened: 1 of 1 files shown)


davidhalter:
Very awesome. I like the changes!

davidhalter:
I don't get why the Python core devs use different path schemas on Windows and linux, but that's a different discussion I guess :)

PR #1067, opened 2018-03-23
Properly raise broken pipe exception

micbou:
We need to pass an error message to the `IOError` exception in addition to the error number; otherwise, it is raised as a normal exception with the message `32` (`errno.EPIPE` being equal to `32`). See [the `EnvironmentError` docs](https://docs.python.org/2.7/library/exceptions.html#exceptions.EnvironmentError) (`IOError` is a subclass of `EnvironmentError` on Python 2) and [the `OSError` docs](https://docs.python.org/3.6/library/exceptions.html#OSError) (`IOError` is an alias of `OSError` on Python 3).


davidhalter:
Makes sense!

PR #1068, opened 2018-03-23
Add Appveyor configuration

micbou:
As discussed in PR https://github.com/davidhalter/jedi/pull/1065, here's the configuration file to run the tests on AppVeyor. PR https://github.com/davidhalter/jedi/pull/1066 and PR https://github.com/davidhalter/jedi/pull/1067 are required for the tests to pass. You can see [the build result on my account](https://ci.appveyor.com/project/micbou/jedi/build/1.0.52).

I had to change the `JEDI_TEST_ENVIRONMENT` environment variable to store the full Python path instead of just the version since there is no version in the Python executable name on Windows, which means that the environment cannot be deduced from the name. For the same reason, I changed the `test_versions` test to only check the current environment.

We may consider reducing the number of jobs since a full build takes about 50 minutes to complete (jobs are run sequentially on AppVeyor). We could for example only run the jobs `TOXENV=py27` and `TOXENV=py36`. We should be fine though if we enable [rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds).

I'll let you add the badge to the README. You can find the link in the project settings on the AppVeyor website.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1068)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,11 @@ python:
   - 3.6
 
 env:
-  - JEDI_TEST_ENVIRONMENT=27
-  - JEDI_TEST_ENVIRONMENT=33
-  - JEDI_TEST_ENVIRONMENT=34
-  - JEDI_TEST_ENVIRONMENT=35
-  - JEDI_TEST_ENVIRONMENT=36
+  - JEDI_TEST_ENVIRONMENT=/opt/python/2.7/bin/python
+  - JEDI_TEST_ENVIRONMENT=/opt/python/3.3/bin/python
+  - JEDI_TEST_ENVIRONMENT=/opt/python/3.4/bin/python
+  - JEDI_TEST_ENVIRONMENT=/opt/python/3.5/bin/python
+  - JEDI_TEST_ENVIRONMENT=/opt/python/3.6/bin/python
 
 matrix:
   allow_failures:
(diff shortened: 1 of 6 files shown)


davidhalter:
You can also just commit the small changes individually. I can squash in the end. Then I see what you change.

davidhalter:
I would actually recommend you not to touch the travis stuff too much. I have been debugging for a pretty long time until it worked.

I think your points about the paths are correct. We need to be able to supply paths like `c:\python3.5\python.exe` for Windows. But at the same time I don't think necessary to change anything for linux. It used to work great and the new API is just less generalized.

IMO we should just add a new variable like `JEDI_TEST_ENVIRONMENT_EXECUTABLE` that appveyor can use. Wouldn't that suffice? A change that I definitely do not want is the change to `tox.ini`. I'm using those pretty regularly.

micbou:
Okay, I added the `JEDI_TEST_ENVIRONMENT_EXECUTABLE` variable for AppVeyor as you suggested and disabled the `test_versions` test on Windows since I don't see a simple way to adapt this test. The `.travis.yml` and `tox.ini` files are now left unmodified.

davidhalter:
This is awesome! Thanks a lot.

I love that we now have a Windows CI!

davidhalter:
We are currently running all tests on Windows. I'm fine with this. Maybe we need to reduce it a bit in the future if it's annoying that it takes so long, but I don't really care at the moment.

This is really fucking awesome. It's something that I was always wishing for. :)

Where are Python versions typically installed on Windows? Is there a convention? It would be interesting to automatically detect virtualenvs in jedi.

davidhalter:
I added the badge now. AWESOMERINO.

micbou:
> Where are Python versions typically installed on Windows? Is there a convention? It would be interesting to automatically detect Python versions in jedi.

That depends on the version. Python 2.7 is installed by default to `C:\Python27` while Python 3 to `%USERPROFILE%\AppData\Local\Programs\Python\Python3*`. Probably the best way to detect versions would be to check for the `HKEY_LOCAL_MACHINE\SOFTWARE\Python\PythonCore\x.y\InstallPath` keys in the Windows registry, where `x.y` is the Python version. We could use [the `_winreg` module](https://docs.python.org/3/library/winreg.html) for that.

> I added the badge now. AWESOMERINO.

Nice!

micbou:
I suppose you want to extend the `find_python_environment` function to detect all the versions on Windows. Am I right?

davidhalter:
Exactly :)

micbou:
I'll work on it.

PR #1077, opened 2018-04-07
Find Python environments on Windows using the registry

micbou:
This PR extends the `get_python_environment` function to find Python on Windows when a string of the form `pythonX.Y` (where `X` and `Y` are the major and minor versions) is passed to that function. This is done by looking up the Windows registry.

With that change, we can now use the `JEDI_TEST_ENVIRONMENT` variable on AppVeyor.

Since there is no guarantee that the Python version may appear in the sys paths (especially on Windows), `test_versions` is changed to check that it is not empty and that the version from the Python environment is the expected version.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1077)


davidhalter:
Looks like a good improvement! I'll take a look at it tomorrow. It's probably the last thing that will make it into 0.12.0.

jedi/api/environment.py
```diff
@@ -163,12 +163,26 @@ def find_python_environments():
...
+# TODO: this function should probably return a list of environments since
+# multiple Python installations can be found on a system for the same version.
+def get_python_environment(python):
+    """
+    Return the first Python environment found for a given path or for a string
+    of the form 'pythonX.Y' where X and Y are the major and minor versions of
+    Python.
+    """
+    exe = find_executable(python)
+    if exe:
```
davidhalter: Don't we need to differentiate here as well for Windows? AFAIK it's different for Windows than OS X/Linux. This probably only matters for the few cases where people actually provide a path. I think for Windows it's just one less `dirname` (I haven't really looked at `PC/getpathp.c`).

jedi/api/environment.py
```diff
@@ -206,6 +220,33 @@ def _get_executable_path(path, safe=True):
...
+    # The winreg module is named _winreg on Python 2.
+    try:
+      import winreg
+    except ImportError:
+      import _winreg as winreg
+
+    # TODO: support Python Anaconda.
+    sub_keys = [
+      r'SOFTWARE\Python\PythonCore\{version}\InstallPath',
+      r'SOFTWARE\Wow6432Node\Python\PythonCore\{version}\InstallPath',
```
davidhalter: WTF is `Wow6432Node` lol. I also see it in CPython's source code. You don't have to explain, I'm just a bit surprised that this is in there.

davidhalter:
> Right. We now recursively look for Lib\os.py on Windows to find the Python prefix (like it's done in PC/getpathp.c). It won't always work since there are other rules so I added a note that we may consider deducing the prefix by running the Python interpreter and printing the value of sys.prefix.

I'm fine with both. If it doesn't work for people will report it and we can start fixing it. There's probably quite a lot of things we can improve on in the future.

davidhalter:
I'm merging, because the tests have passed and I like what you did.

I might get back do you if there are still open issues.

davidhalter:
The remaining issues are oddly enough relatively similar on Linux and Windows. I created two tests to test virtualenvs (so all other tests pass, just the new ones fail). 

I'll definitely try to fix the Linux part https://travis-ci.org/davidhalter/jedi/jobs/363909626

But if you have any ideas or just want to fix the Windows part, let me know. https://ci.appveyor.com/project/davidhalter/jedi/build/9

PR #1079, opened 2018-04-09
Do not cache unimportable compiled module

micbou:
Jedi attempts to cache a compiled module even if it failed to import it, resulting in the following error when trying to get the description of that module:
```
  File "script.py", line 26, in <module>
    completions[0].description
  File "jedi\jedi\api\classes.py", line 464, in description
    return Definition.description.__get__(self)
  File "jedi\jedi\api\classes.py", line 519, in description
    typ = self.type
  File "jedi\jedi\api\classes.py", line 146, in type
    for context in self._name.infer():
  File "jedi\jedi\evaluate\imports.py", line 172, in infer
    level=self._level,
  File "jedi\jedi\evaluate\imports.py", line 279, in follow
    return self._do_import(self.import_path, self.sys_path_with_modifications())
  File "jedi\jedi\evaluate\imports.py", line 367, in _do_import
    safe_module_name=True,
  File "jedi\jedi\evaluate\imports.py", line 503, in _load_module
    add_module(evaluator, module_name, module, safe=safe_module_name)
  File "jedi\jedi\evaluate\imports.py", line 515, in add_module
    evaluator.module_cache.add(module, module_name)
  File "jedi\jedi\evaluate\imports.py", line 40, in add
    path = module.py__file__()
AttributeError: 'NoneType' object has no attribute 'py__file__'
```
The issue can be reproduced by getting the description of the `QtBluetooth` module from [`PyQt5`](https://pypi.python.org/pypi/PyQt5) on Windows:
```python
import jedi
completions = jedi.Script('import PyQt5.QtBlueTooth').completions()
completions[0].description
```
The fix is to simply not cache the module in that case.

I'd like to add a test for this but it's tricky to do since this occurs with compiled modules and the reproducible case is specific to Windows. Any suggestions?

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1079)

### DIFF
--- a/jedi/evaluate/imports.py
+++ b/jedi/evaluate/imports.py
@@ -500,19 +500,21 @@ def _load_module(evaluator, path=None, code=None, sys_path=None,
             )
         else:
             module = compiled.load_module(evaluator, path=path, sys_path=sys_path)
-    add_module(evaluator, module_name, module, safe=safe_module_name)
+
+    if module_name is not None and module is not None:
+        add_module(evaluator, module_name, module, safe=safe_module_name)
+
     return module
 
 
 def add_module(evaluator, module_name, module, safe=False):
-    if module_name is not None:
-        if not safe and '.' not in module_name:
-            # We cannot add paths with dots, because that would collide with
-            # the sepatator dots for nested packages. Therefore we return
-            # `__main__` in ModuleWrapper.py__name__(), which is similar to
-            # Python behavior.
-            return
-        evaluator.module_cache.add(module, module_name)
+    if not safe and '.' not in module_name:
+        # We cannot add paths with dots, because that would collide with
+        # the sepatator dots for nested packages. Therefore we return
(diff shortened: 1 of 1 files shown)


davidhalter:
Good find. I think it's pretty obvious that this is a bug and testing it would be nice, but it's not like we'll just delete that code again, so I'm fine without a test.

IMO we shouldn't check this in the add_module function. Just check it in `_load_module`. I will then also move the `if module_name is not None` to a better place.

micbou:
Moved the two checks to `_load_module`.

davidhalter:
Thanks! I changed some things afterwards, but your fix remained.

davidhalter:
I also wrote a test. I just mocked compiled.load_module. If we want to do something better we can write a .pyc I guess that raises an error.

micbou:
Using a mock seems the right thing to do here. I'll try to remember about `monkeypatch` next time I found an issue with compiled modules.

PR #1080, opened 2018-04-09
Fix virtual environment tests

micbou:
There were three issues:
 - on Windows, prior to Python 3.6, there is no `activate` file but only `activate.bat` in the `Scripts` folder when creating a virtual environment;
 - on Travis, system Python 3.4 cannot properly create a virtual environment. We need to install the `python3.4-venv` package;
 - on Travis again, the pre-installed Python 3.6 is not owned by root which makes it unsafe to use. We need to install it manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,12 @@ env:
   - JEDI_TEST_ENVIRONMENT=35
   - JEDI_TEST_ENVIRONMENT=36
 
+addons:
+  apt:
+    packages:
+     # Required to properly create a virtual environment with system Python 3.4.
+     - python3.4-venv
+
 matrix:
   allow_failures:
     - python: pypy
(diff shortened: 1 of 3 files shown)


micbou:
The fix for the last point wasn't working with Python 3.6 and `JEDI_TEST_ENVIRONMENT=36` because in that case, there are two pre-installed Python 3.6 that are not owned by root: `/opt/python/3.6.3/bin/python` and `/opt/python/3.6/bin/python`, the former being used to create the virtual environment. So, instead of manually installing Python 3.6, we change the permissions of both Python.

micbou:
> How did you realize it was a chroot issue? Did you go into travis debug mode?

No, I put some `print` statements in the code until I found the issue (`os.getuid()` was returning `2000`, which is the user `travis`, instead of `0`). Maybe I'll try Travis debug mode next time.

PR #1082, opened 2018-04-13
Popen

davidhalter:
To fix the remaining issues with venvs/virtualenv I changed a few things:

1. Added the CREATE_NO_WINDOW flag for Windows
2. Look out for copies of installed Python versions to make it possible to safely execute those.
3. Some SamEnvironment magic was not really implemented the right way, you don't have to care.

@micbou Can you please review? :)

### DIFF
--- a/jedi/_compatibility.py
+++ b/jedi/_compatibility.py
@@ -508,4 +508,4 @@ class GeneralizedPopen(subprocess.Popen):
         creation_flags = 0
         if os.name == 'nt':
             creation_flags = subprocess.CREATE_NO_WINDOW
-        super(GeneralizedPopen, self).__init__(*args, creation_flags=creation_flags, **kwargs)
+        super(GeneralizedPopen, self).__init__(*args, creationflags=creation_flags, **kwargs)
--- a/jedi/_compatibility.py
+++ b/jedi/_compatibility.py
@@ -507,5 +507,10 @@ class GeneralizedPopen(subprocess.Popen):
     def __init__(self, *args, **kwargs):
         creation_flags = 0
         if os.name == 'nt':
-            creation_flags = subprocess.CREATE_NO_WINDOW
+            try:
+                # Was introduced in Python 3.7.
+                CREATE_NO_WINDOW = subprocess.CREATE_NO_WINDOW
+            except AttributeError:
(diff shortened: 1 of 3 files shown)


jedi/_compatibility.py
```diff
@@ -500,3 +501,16 @@ class Parameter(object):
         VAR_POSITIONAL = object()
         KEYWORD_ONLY = object()
         VAR_KEYWORD = object()
+
+
+class GeneralizedPopen(subprocess.Popen):
+    def __init__(self, *args, **kwargs):
```
micbou: The signature of `Popen` is `args` and a bunch of named parameters so it would be better to set the parameters of the constructor to `args, **kwargs`.

jedi/_compatibility.py
```diff
@@ -500,3 +501,16 @@ class Parameter(object):
...
+        creation_flags = 0
+        if os.name == 'nt':
+            try:
+                # Was introduced in Python 3.7.
+                CREATE_NO_WINDOW = subprocess.CREATE_NO_WINDOW
+            except AttributeError:
+                CREATE_NO_WINDOW = 0x08000000
+            creation_flags = CREATE_NO_WINDOW
```
micbou: We can directly do `kwargs[ 'creation_flags' ] = CREATE_NO_WINDOW` instead of using a variable.

jedi/api/environment.py
```diff
@@ -301,15 +326,31 @@ def _is_safe(executable_path):
...
+    if _is_unix_safe_simple(real_path):
+        return True
+
+    # Just check the list of known Python versions. If it's not in there,
+    # it's likely an attacker or some Python that was not properly
+    # installed in the system.
+    for environment in find_python_environments():
+        if environment.executable == real_path:
+            return True
+        else:
```
micbou: We don't need the `else:` since we always return in the `if` branch.

jedi/_compatibility.py
davidhalter: I disagree, I like the pattern of "add something"/"remove something" in these cases. :)

jedi/_compatibility.py
davidhalter: Thanks

davidhalter:
I fixed two of the issues you raised.

PR #1086, opened 2018-04-17
check_fs: handle FileNotFoundError

blueyed:
Ref: https://github.com/davidhalter/jedi-vim/pull/801

### DIFF
--- a/jedi/evaluate/imports.py
+++ b/jedi/evaluate/imports.py
@@ -17,7 +17,8 @@ from parso.python import tree
 from parso.tree import search_ancestor
 from parso import python_bytes_to_unicode
 
-from jedi._compatibility import unicode, ImplicitNSInfo, force_unicode
+from jedi._compatibility import (FileNotFoundError, ImplicitNSInfo,
+                                 force_unicode, unicode)
 from jedi import debug
 from jedi import settings
 from jedi.parser_utils import get_cached_code_lines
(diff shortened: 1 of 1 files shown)


jedi/evaluate/imports.py
```diff
@@ -17,7 +17,8 @@
 from parso.tree import search_ancestor
 from parso import python_bytes_to_unicode
 
-from jedi._compatibility import unicode, ImplicitNSInfo, force_unicode
+from jedi._compatibility import (FileNotFoundError, ImplicitNSInfo,
+                                 force_unicode, unicode)
```
blueyed: Not sure what style you prefer.
Would be nice to have some [isort](https://github.com/timothycrosley/isort) config for the project.

blueyed:
Where would a test go?

jedi/evaluate/imports.py
davidhalter: I don't really care and if I would I would just change it when I see it again. I think both are fine and there's more pressing issues than this :)

davidhalter:
Thanks!

jedi/evaluate/imports.py
blueyed: Sure, but those are little distractions that can be automated away.
If you do not care I suggest to use the default of isort then maybe?  But the diff is quite huge with that: http://ix.io/183Y (109 files changed, 481 insertions(+), 578 deletions(-)) :D
Maybe not enforce/push it, but add a config so that it could be adopted over time?  (it's easy to run isort on a buffer in Vim after all)

Using the following config you get http://ix.io/183Z (115 files changed, 534 insertions(+), 549 deletions(-)):
```
[isort]
known_first_party = jedi
default_section = THIRDPARTY
lines = 79
multi_line_output = 5
include_trailing_comma = 1
order_by_type = 1
combine_as_imports = 1
force_sort_within_sections = 1
use_parentheses = 1
```

`lines` could be longer there (as with jedi-vim now), but usually 79 is good.

Anyway, your call and probably not a good idea to apply the whole patch as-is, but I still think that having some config (if you do not like the defaults (which you do not know about ;D)) is a good first step.

davidhalter:
The test should probably go somewhere to the test_api folder. You can probably just create a new file test_settings.py.

PR #1088, opened 2018-04-18
Extend create_environment to accept an executable path

micbou:
There is currently no way to create an environment for a given executable path through the API. The `create_environment` function only accepts a path to a virtual environment. This PR extends that function to also accept executable paths.

@davidhalter I sent you an email about this but didn't get an answer. You probably missed it.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1088)

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -287,11 +287,15 @@ def get_system_environment(version):
 
 def create_environment(path, safe=True):
     """
-    Make it possible to create an environment by hand.
+    Make it possible to manually create an environment by specifying a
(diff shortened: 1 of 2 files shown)


davidhalter:
@micbou I saw it. I'm just quite busy at the moment . Will answer in a few days when I have some more time.

davidhalter:
This sounds reasonable. I'm usually a bit afraid to use two different semantics with one function (loading from a dir OR from an executable path), but here it seems quite reasonable. Thanks a lot for your work!

Issue #1107, opened 2018-04-27
Environment: improve getting sys.prefix


blueyed:
I've found the TODO with _get_python_prefix:
https://github.com/davidhalter/jedi/blob/91499565a987ff35b078960bdb966351ea976eab/jedi/api/environment.py#L249-L263

I think `Environment` could be changed to not take the `prefix` by default, but get it then together with the version then:

```diff
 jedi/api/environment.py | 23 ++++++++++++++++-------
 1 file changed, 16 insertions(+), 7 deletions(-)

diff --git i/jedi/api/environment.py w/jedi/api/environment.py
index 7e0a8507..7df415e0 100644
--- i/jedi/api/environment.py
+++ w/jedi/api/environment.py
@@ -73,7 +73,11 @@ def __init__(self, path, executable):
 
     def _get_version(self):
         try:
-            process = GeneralizedPopen([self.executable, '--version'], stdout=PIPE, stderr=PIPE)
+            process = GeneralizedPopen([
+                self.executable,
+                'import sys; print(sys.prefix); '
+                'print("%d.%d.%d" % sys.version_info[:3])'],
+                stdout=PIPE, stderr=PIPE)
             stdout, stderr = process.communicate()
             retcode = process.poll()
             if retcode:
@@ -81,12 +85,16 @@ def _get_version(self):
         except OSError:
             raise InvalidPythonEnvironment()
 
-        # Until Python 3.4 wthe version string is part of stderr, after that
-        # stdout.
-        output = stdout + stderr
-        match = re.match(br'Python (\d+)\.(\d+)\.(\d+)', output)
-        if match is None:
-            raise InvalidPythonEnvironment("--version not working")
+        output = stdout.splitlines()
+        if len(output) != 2:
+            raise InvalidPythonEnvironment(
+                "Could not get version information (stdout=%r, stderr=%r)" % (
+                    stdout, stderr))
+        match_version = re.match(r'(\d+)\.(\d+)\.(\d+)', output[1])
+        if match_version is None:
+            raise InvalidPythonEnvironment(
+                "Could not get version information (stdout=%r, stderr=%r)" % (
+                    stdout, stderr))
+
+        # TODO: save sys.prefix - does it work with the Windows registry handling?
 
         return _VersionInfo(*[int(m) for m in match.groups()])
 
```
Downside: `python --version` is slightly faster than `python -c 'print(__import__("sys").version_info)'`
[...]

PR #1108, opened 2018-04-27
Environment: improve/fix getting sys.prefix

blueyed:
Fixes https://github.com/davidhalter/jedi/issues/1107.


davidhalter:
I like the idea of fixing all of this.

I just think we should probably do it right. We should probably just start the subprocess and work with it instead of trying stuff with `-c`.

blueyed:
Let's move it into EvaluatorSubprocess then?

davidhalter:
I guess, yes. Probably to `_CompiledSubprocess`. There's already a `get_sys_path` function there that is being used, so we could use it as well for prefix and version.

blueyed:
Just a side-note: for jedi-vim it would be beneficial to get an Environment based on the exe only.

I will now look into addressing your comments / improve this.

jedi/api/environment.py
```diff
@@ -55,40 +55,53 @@ class Environment(_BaseEnvironment):
     should not create it directly. Please use create_environment or the other
     functions instead. It is then returned by that function.
     """
-    def __init__(self, path, executable):
-        self.path = os.path.abspath(path)
+    def __init__(self, executable):
+        self.path = None
```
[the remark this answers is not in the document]
davidhalter: I totally agree. The only problem is that the API is alrady out there. But we can probably just deprecate `self.path`.

jedi/api/environment.py
```diff
@@ -55,40 +55,53 @@ class Environment(_BaseEnvironment):
     should not create it directly. Please use create_environment or the other
     functions instead. It is then returned by that function.
     """
-    def __init__(self, path, executable):
-        self.path = os.path.abspath(path)
+    def __init__(self, executable):
```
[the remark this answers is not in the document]
davidhalter: Hmm, what's the advantage of specifying all those things? I would probably prefer simplicity. If we have to check it anyway for some cases, checking it for all cases probably makes sense. (It doesn't take a lot of time, what takes a lot of time is to start a Python interpreter).

davidhalter:
I agree. It's probably generally better for a lot of use cases to just specify an executable.

For virtualenvs it's sometimes more practical to specify a prefix, but that could just try to locate the exe and then progress like the exe mechanism.

blueyed:
Refactored, please review.

davidhalter:
That already looks way better. I'll have a closer look in the next few days.

jedi/api/environment.py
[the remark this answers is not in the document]
blueyed: Everythings gets retrieved in one go now.

jedi/api/environment.py
```diff
@@ -246,23 +242,6 @@ def find_system_environments():
             pass
```
davidhalter: _get_version was deleted in the super class, so we can remove it here safely as well. However SameEnvironment shouldn't execute anything (It's mostly there for REPL completions).

I think we should probably be doing something like this:

```
class Environment(object):
    def __init__(self, executable):
      self.executable = executable
      self._load_attributes()

   def _load_attributes(self):
      ... # subprocess stuff

class SameEnvironment():
   def __init__(self):
      super(SameEnvironment, self).__init__(sys.executable)

   def _load_attributes(self):
      self.version_info = sys.version_info
      self.path = sys.prefix
```

blueyed: I think it can be simpler - pushed 2bb32e5.

davidhalter:
```
E           jedi.api.environment.InvalidPythonEnvironment: Could not get version information (TypeError("get_subprocess() missing 1 required positional argument: 'version'",))
```

blueyed:
@davidhalter 
More context please.
(there is no version argument, is there?!  https://github.com/davidhalter/jedi/blob/2bb32e53093c1c7f110558bdef54182207a94d80/jedi/evaluate/compiled/subprocess/__init__.py#L32)

blueyed:
Oh, tests - I have not looked into fixing them yet.
Wanted to get your feedback on this in general first.

blueyed:
This is because of 282c6a2b.
There is a chicken-egg problem now: we use the subprocess to get the version, but that requires the version already (for the pickle protocol).
I think we need to start with the lowest protocol version initially (https://github.com/davidhalter/jedi/pull/1158).
/cc @micbou

jedi/api/environment.py
```diff
@@ -46,49 +44,46 @@ def _sha256(self):
...
+            raise InvalidPythonEnvironment(
+                "Could not get version information (%r)" % exc)
+
+        self.executable = info[0]
         """
-        The path to an environment, matches ``sys.prefix``.
+        The Python executable, matches ``sys.executable``.
         """
-        self.executable = os.path.abspath(executable)
+        self.path = info[1]
```
blueyed: @davidhalter 
Should we keep `path` here, or remove, or deprecate it instead of `self.prefix`?

micbou:
I see the problem. I would suggest to either revert PR https://github.com/davidhalter/jedi/pull/1152 or print `sys.version` and `sys.prefix` through the Python interpreter and parse the output.

blueyed:
@micbou 
Reverting is not necessary.
I think the server should pass its version to the client/subprocess, and then both would use `highest_pickle_protocol` with their own version and the other.

The internal protocol is not really clear to me, i.e could the instance be made available in the callback?
Then `_get_info` could set the pickle protocol on the server already and return what it sets.
(For now I've came across https://github.com/davidhalter/jedi/pull/1159 - which might simplify it a bit.)

davidhalter:
If it's too much of a hassle we can also just revert #1152. It's still possible to reintroduce it at a later stage. I think this PR is way more important.

Generally if you want to make the pickle version 4 work in certain cases don't create too much complexity for it. It's really not worth it. IMO it's almost too much code for the effect it has at the moment. Please don't add way more code to pick a pickle version than it uses now.

micbou:
> I think the server should pass its version to the client/subprocess, and then both would use highest_pickle_protocol with their own version and the other.

Except that the server also needs the version of the client. How are you going to pass that version from the client to the server?

blueyed:
@micbou 
It gets retrieved via the _get_info function, used when setting up the connection/Environment:
https://github.com/davidhalter/jedi/pull/1108/files#diff-c204f420c3d90e6c005f6354c65dc638R47

micbou:
@blueyed So the plan is to send `_get_info` with the lowest pickle protocol version (i.e. version 2), get the version of the client, compute the highest pickle protocol version, and use that version for all subsequent communications?

blueyed:
@micbou 
Yep.
Done in https://github.com/davidhalter/jedi/pull/1108/commits/9e66253e3a2aff6b9a96ebdd0c3b26639ecf4654.

blueyed:
Any idea about the `test_venv_and_pths` failures for all builds != `JEDI_TEST_ENVIRONMENT=27` ?

blueyed:
Are there plans to clean old environments already?
We should probably remember their starttime already, and when they are used?

Then maybe by default only the last X recently used should be kept running?  (to be checked when a new environment is created)

jedi/api/environment.py
davidhalter: Keep path for now. I think we should have a discussion about prefix in a separate PR or issue. I probably think we should deprecate it.

davidhalter:
I'm trying to pin down the remaining error. Have been searching for 2 hours, no luck so far. Will continue tomorrow while commuting.

davidhalter:
I fixed it in d06e55a it's on master, please rebase.

davidhalter:
You are probably finished, right? Everything else could go into a different PR?

blueyed:
Thanks!
Rebased.
I would still like to rename `Environment.path` to `Environment.prefix`.  We could keep `path` as a (deprecated) alias maybe then?
This could also be done in a separate PR though.

davidhalter:
I agree we should do that. I however have no idea why the tests still fail :-/ I cannot reproduce it locally anymore. I'll try to debug it on travis.

davidhalter:
Ok. I can actually reproduce it, but only like this: `pytest test_api/test_environment.py test_evaluate/test_sys_path.py`. If I just use `test_sys_path.py`, the test passes. I'm pretty sure it has to do something with caching, but I'm not yet sure why this didn't appear before.

blueyed:
This fixes it:

```diff
diff --git i/jedi/api/environment.py w/jedi/api/environment.py
index 06eb90ec..f6f47511 100644
--- i/jedi/api/environment.py
+++ w/jedi/api/environment.py
@@ -116,6 +116,12 @@ def get_sys_path(self):
         # site.py gets executed.
         return self._get_subprocess().get_sys_path()
 
+    def __del__(self):
+        try:
+            self._subprocess.kill()
+        except AttributeError:
+            pass
+
 
 class SameEnvironment(Environment):
     def __init__(self):
```

The AttributeError might come from `AttributeError: 'SameEnvironment' object has no attribute '_subprocess'`.

It makes sense in general I would say, but there is likely still an issue here that needs to be tracked down.

blueyed:
The problem is that `create_environment` picks up the previous subprocess.
`Environment` and `get_subprocess` should probably have an arg to use a new instance (and deleting the existing one).
I would assume that if I call `create_environment` it would be created after all, and not re-used.

`create_environment` is meant to be changed anyway (https://github.com/davidhalter/jedi/issues/1155).

I will push the `__del__` patch for now.

blueyed:
Only a temporary failure on Travis: https://travis-ci.org/davidhalter/jedi/jobs/402926601 - cannot restart it myself.

jedi/api/environment.py
```diff
@@ -32,7 +32,8 @@ class InvalidPythonEnvironment(Exception):
 class _BaseEnvironment(object):
     @memoize_method
     def get_grammar(self):
-        version_string = '%s.%s' % (self.version_info.major, self.version_info.minor)
```
davidhalter: BTW: I didn't mention this previously (now I do), but I think it's ok to have a line length of 100 characters occasionally (PEP8 thinks so too), I think these two examples are more readable if they are longer, especially the second one.

davidhalter:
Now the tests fail because of appveyor. I have removed the Python 3.3 tests for now. Planning to drop it anyway.

jedi/api/environment.py
blueyed: Ok, please add it to the config then.

davidhalter: To which config?

blueyed: `setup.cfg`:

```ini
[flake8]
max-line-length = 100
```

davidhalter:
Please comment on #1176.

PR #1111, opened 2018-04-27
Travis: remove TOXENV=cov from allowed failures

blueyed:


davidhalter:
I'm fine with that if it doesn't go back to failing all the time.

If you check https://travis-ci.org/davidhalter/jedi/builds/366495866, you can see that only the coverage job failed. I have no idea why. It seems to have fixed itself, but it's not like we changed anything (AFAIK). Do you have any idea? Regardless, I will probably just merge and we can try if it works for us.

blueyed:
Ok, let's investigate unexpected failures when this happens again then.

PR #1115, opened 2018-05-02
Fix namespace autocompletion error

m-novikov:
When using calling full_name on namespace autocompletion error is raised
```
...
     for n in reversed(module_context.py__name__().split('.')):
AttributeError: 'ImplicitNamespaceContext' object has no attribute 'py__name__'
```
Also while fixing it found out that full_name for namespaces was build incorrectly.

### DIFF
--- a/jedi/evaluate/context/namespace.py
+++ b/jedi/evaluate/context/namespace.py
@@ -3,24 +3,18 @@ from itertools import chain
 
 from jedi.evaluate.cache import evaluator_method_cache
 from jedi.evaluate import imports
-from jedi.evaluate.filters import DictFilter, AbstractNameDefinition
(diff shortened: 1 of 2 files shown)


m-novikov:
Why is ```py__path__``` defined as property and not as method? It seems really strange, because it goes through the trouble to mimic method behaviour.

davidhalter:
Thanks for the great pull request! You even removed code. That's awesome.

> Why is py__path__ defined as property and not as method?

Yeah, that's just really bad code. It probably started with something else and then we didn't clean up. At some points jedi uses properties in combination with AttributeErrors to make clear that a class doesn't have a function. I don't think it's a nice pattern, but I'm currently not sure what else to use.

In this case it was just plain stupid, had no use at all and even generated worse tracebacks, because of the lambda. Thanks for removing it :)

PR #1118, opened 2018-05-05
Fix _get_numpy_doc_string_cls: use cache

blueyed:
I've noticed that Jedi tries to import numpydoc a lot when using
jedi-vim's goto method in jedi_vim.py itself (via printing in Neovim's
VimPathFinder.find_spec).

This patch uses the cache before trying the import again and again.


davidhalter:
Makes definitely sense. Thanks!

PR #1119, opened 2018-05-10
Improve __repr__ for BaseDefinition and AbstractNameDefinition

blueyed:
TODO:

- [ ] tests

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -360,7 +360,11 @@ class BaseDefinition(object):
         return Definition(self._evaluator, context.name)
 
     def __repr__(self):
-        return "<%s %s>" % (type(self).__name__, self.description)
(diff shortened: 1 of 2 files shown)


jedi/api/classes.py
```diff
@@ -360,7 +360,12 @@ def parent(self):
         return Definition(self._evaluator, context.name)
 
     def __repr__(self):
-        return "<%s %s>" % (type(self).__name__, self.description)
+        return "%s(name=%r, full_name=%r, description=%r)" % (
+            self.__class__.__name__,
+            self._name,
+            self.full_name,
```
davidhalter: Why do we need name if we have full_name already? I agree that the keywords make some things a bit more clear, but using both feels like redundancy.

Also I think that we should stick with `<Definition ...>` instead of `Definition()`, because the definition cannot be created like that.

Issue #1123, opened 2018-05-17
Include output / more information with "has crashed" exception


blueyed:
Would help to diagnose issues like https://github.com/davidhalter/jedi-vim/issues/820#issuecomment-389810445 and https://github.com/davidhalter/jedi/issues/1113.

Likely the output is some traceback on stderr, and the exception message could be grabbed from there.
It might also be good to just include all of `stderr` in this case, even though it might be long.

PR #1124, opened 2018-05-17
Include stderr with "subprocess has crashed" exception

blueyed:
This does not add it to the other similar exception raised from `kill`,
since this should be something like "was killed already" anyway.

Fixes https://github.com/davidhalter/jedi/issues/1123.


blueyed:
Sadly no feedback from users.. :/

I've pushed a fixup (and rebased it).

It helped me already when working on improving the `Environment` a while ago, and I think it makes sense also without any user feedback - but should get squash-merged then.

davidhalter:
Makes sense. Thanks!

PR #1125, opened 2018-05-19
Travis: do not use --quiet with `pip install`

blueyed:
This drops any output including the version(s) that are installed.

Currently Python 3.3. builds appear to fail, because tox tries to install `wheel`?!

https://travis-ci.org/davidhalter/jedi/jobs/380136672#L577

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 sudo: true
+cache: false
 python:
   - 2.7
   - 3.3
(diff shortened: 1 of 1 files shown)


davidhalter:
I'm probably going to drop travis 3.3 tests. It's been quite annoying to deal with it on travis. appveyor still has some tests and I have some tests that I run before deploying so we should be fine.

I also think that I'm probably going to drop 3.3 in the next release.

davidhalter:
Dropped 3.3. I really think this is the more pragmatic way to go. Note that the environments for 3.3 are still tested. Thanks for your work though.

PR #1140, opened 2018-06-06
Fix flow analysis test

micbou:
There is [no `seekable` method for file objects on Python 2](https://docs.python.org/2/library/stdtypes.html#bltin-file-objects). Use `flush` instead which is available on both Python.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1140)


davidhalter:
Sorry, I was too lazy to fix that. Thanks for finding it!

PR #1143, opened 2018-06-15
Parse correct AST attribute for version

carlwgeorge:
Earlier development versions of Python 3.7 added the docstring field to AST nodes.  This was later reverted in Python 3.7.0b5.

https://bugs.python.org/issue29463
https://github.com/python/cpython/pull/7121

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@
 from setuptools import setup, find_packages
 
 import ast
-import sys
 
 __AUTHOR__ = 'David Halter'
 __AUTHOR_EMAIL__ = 'davidhalter88@gmail.com'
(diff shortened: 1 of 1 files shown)


davidhalter:
Great, thanks!

ilovezfs:
@davidhalter a new sdist is needed on PyPI as the current one fails to build with python 3.7 due to the issue that was fixed here.

davidhalter:
New release is coming very soon. I'm just trying to stabilize both jedi and parso.

davidhalter:
I did a new release, now.

ilovezfs:
Awesome! Thanks @davidhalter

PR #1145, opened 2018-06-16
Add support for DefaultDict on jedi_typing.py.

### LINKED ISSUE #1144: No completions for DefaultDict
```python
>>> import jedi
>>> source = '''
... from typing import DefaultDict
... d: DefaultDict[int, int]
... d.
... '''
>>> script = jedi.Script(source, 4, 2, 'foo.py')
>>> script.completions()
[]
```

While if I try with `Dict`...

```python

>>> source = '''
... from typing import Dict
... d: Dict[int, int]
... d.
... '''
>>> script = jedi.Script(source, 4, 2, 'foo.py')
>>> len(script.completions())
42
```

tarcisioe:
Fixes #1144.

### DIFF
--- a/jedi/evaluate/jedi_typing.py
+++ b/jedi/evaluate/jedi_typing.py
@@ -81,6 +81,9 @@ def factory(typing_name, indextypes):
     class Dict(MutableMapping, dict):
         pass
 
+    class DefaultDict(MutableMapping, dict):
+        pass
+
     dct = {
         "Sequence": Sequence,
         "MutableSequence": MutableSequence,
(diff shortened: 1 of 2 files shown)


davidhalter:
Can you add a test for it as well. Thanks for the fix!

tarcisioe:
I wasn't sure what to test, so I duplicated the tests that were done to `typing.Dict.` as well. I also didn't know how to support the instance variable `default_factory` that DefaultDicts have, because all other typing stubs only have methods and I didn't know how I could type a variable (no type annotations are used on the file, and it would be Python 3.6 only if I used variable annotations).

davidhalter:
This is good enough. Dict in typing.py is pretty much the same as DefaultDict anyway, because there is no way to figure out if a key is present in a dict. Thanks for the tests!

PR #1152, opened 2018-06-20
Use highest possible pickle protocol

micbou:
Currently, Jedi uses version 2 of the pickle protocol to be compatible with all supported versions of Python. However, we can be smarter and use the highest version of the pickle protocol that is compatible with the version of Python running Jedi itself and the one running the environment. Examples:
 - if both Jedi and the environment are running under Python 3.6, use version 4 of the protocol;
 - if Jedi is running under Python 3.6 but the environment is Python 3.3, use version 3;
 - if Jedi is running under Python 3.6 but the environment is Python 2.7, use version 2;
 - etc.

The reason for using the highest possible version of the pickle protocol is that performance is better with newer versions of the protocol as shown by the following results:

<table>
  <tr>    
    <th>Interpreter</th>
    <th colspan="4">Python 2.7</th>
    <th colspan="4">Python 3.6</th>
  </tr>
  <tr>    
    <th>Environment</th>
    <th colspan="2">Python 2.7</th>
    <th colspan="2">Python 3.6</th>
    <th colspan="2">Python 2.7</th>
    <th colspan="2">Python 3.6</th>
  </tr>
  <tr>
    <th rowspan="2">Pickle protocol</th>    
    <td>Before</td>
    <td>After</td>
    <td>Before</td>
    <td>After</td>
    <td>Before</td>
    <td>After</td>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>    
    <td>2</td>
    <td>2</td>
    <td>2</td>
    <td>2</td>
    <td>2</td>
    <td>2</td>
    <td>2</td>
    <td>4</td>
  </tr>
  <tr>    
    <th>os</th>
    <td>0.153s</td>
    <td>0.159s</td>
    <td>0.185s</td>
    <td>0.184s</td>
    <td>0.0978s</td>
    <td>0.105s</td>
    <td>0.074s</td>
    <td>0.0714s</td>
  </tr>
  <tr>
    <th>numpy</th>
    <td>0.271s</td>
    <td>0.26s</td>
    <td>0.262s</td>
    <td>0.259s</td>
    <td>0.206s</td>
    <td>0.198s</td>
    <td>0.188s</td>
    <td>0.173s</td>
  </tr>
  <tr> 
    <th>cv2</th>
    <td>0.964s</td>
    <td>0.949s</td>
    <td>1.08s</td>
    <td>1.07s</td>
    <td>0.513s</td>
    <td>0.513s</td>
    <td>0.465s</td>
    <td>0.308s</td>
  </tr>  
</table>

*Measurements obtained with [this script](https://gist.github.com/micbou/28eae3625de4820dac04b3b38ee0789b) when completing the `os`, `numpy`, and `cv2` ([opencv-python](https://pypi.org/project/opencv-python/)) modules on Ubuntu 18.04 64-bit.*

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1152)

### DIFF
--- a/jedi/_compatibility.py
+++ b/jedi/_compatibility.py
@@ -446,9 +446,6 @@ if sys.version_info[:2] == (3, 3):
     pickle.loads = loads
 
 
-_PICKLE_PROTOCOL = 2
-
-
 def pickle_load(file):
     try:
         if is_py3:
@@ -462,9 +459,9 @@ def pickle_load(file):
         raise
 
 
-def pickle_dump(data, file):
+def pickle_dump(data, file, protocol):
     try:
-        pickle.dump(data, file, protocol=_PICKLE_PROTOCOL)
+        pickle.dump(data, file, protocol)
         # On Python 3.3 flush throws sometimes an error even though the writing
         # operation should be completed.
         file.flush()
@@ -474,6 +471,20 @@ def pickle_dump(data, file):
         if sys.platform == 'win32':
             raise IOError(errno.EPIPE, "Broken pipe")
         raise
+
+
+# Determine the highest protocol version compatible for a given list of Python
+# versions.
+def highest_pickle_protocol(python_versions):
+    protocol = 4
+    for version in python_versions:
+        if version[0] == 2:
+            # The minimum protocol version for the versions of Python that we
+            # support (2.7 and 3.3+) is 2.
+            return 2
+        if version[1] < 4:
+            protocol = 3
+    return protocol
 
 
 try:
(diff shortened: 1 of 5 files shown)


davidhalter:
I'm not that happy with the added complexity, but I can understand the cause. Do you think it's actually worth it? I know it feels like quite a performance increase for cv2 on Python3.6, but it's also just that. I guess I'll leave the decision if I should merge to you.

blueyed:
Interesting!
`highest_pickle_protocol` should have some unit tests probably.

micbou:
> I'm not that happy with the added complexity, but I can understand the cause. Do you think it's actually worth it? I know it feels like quite a performance increase for cv2 on Python3.6, but it's also just that. I guess I'll leave the decision if I should merge to you.

I think it's worth. All modules with huge files are going to benefit from this and not only on Python 3.6 but all versions of Python starting from 3.3.

> highest_pickle_protocol should have some unit tests probably.

Added a bunch of unit tests.

davidhalter:
Fair enough.

PR #1156, opened 2018-06-24
Force unicode when listing module names

micbou:
`pkgutil.iter_modules` return the module name as `str` instead of `unicode` for egg packages on Python 2. This causes the following exception if Jedi is running on Python 3 and the environment is Python 2:
```python
  File "script.py", line 36, in <module>
    jedi.Script('from ', environment=environment).completions()
  File "jedi\api\__init__.py", line 174, in completions
    completions = completion.completions()
  File "jedi\api\completion.py", line 103, in completions
    return sorted(completions, key=lambda x: (x.name.startswith('__'),
  File "jedi\api\completion.py", line 36, in filter_names
    if string.startswith(like_name):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1156)

### DIFF
--- a/jedi/evaluate/compiled/subprocess/functions.py
+++ b/jedi/evaluate/compiled/subprocess/functions.py
@@ -69,7 +69,7 @@ def get_module_info(evaluator, sys_path=None, full_name=None, **kwargs):
 
 def list_module_names(evaluator, search_path):
     return [
-        name
+        force_unicode(name)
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks!

PR #1158, opened 2018-06-30
highest_pickle_protocol: return minimum for empty list

blueyed:
This can be used then for initial communication for getting the actual
version.

Via/for https://github.com/davidhalter/jedi/pull/1108#issuecomment-401556230.


blueyed:
Alternatively/additionally we could also have a constant instead, since the empty-case will be used always initially.

blueyed:
This does not make much sense by its own.
Closing in favor of #1108 (where this is relevant for).

PR #1159, opened 2018-06-30
Simplify subprocess import path handling: use PYTHONPATH

blueyed:
There might be a reason for the complicated code that I am not aware of.. :)
Hopefully Travis will tell.

### DIFF
--- a/jedi/evaluate/compiled/subprocess/__init__.py
+++ b/jedi/evaluate/compiled/subprocess/__init__.py
@@ -139,9 +139,17 @@ class _CompiledSubprocess(object):
         args = (
             self._executable,
             _MAIN_PATH,
-            os.path.dirname(os.path.dirname(parso_path)),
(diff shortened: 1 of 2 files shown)


davidhalter:
First I thought holy molly, this is so much better. Then I realized why I did it in such a complicated fashion. The issue is that people might have other files like os.py or whatever in the directories that jedi/parso reside in. Therefore we have to use all of the meta_path stuff.

IMO it's the safer way, because you will never import something randomly that you shouldn't. Also we basically don't want to modify the PYTHONPATH after it was imported so people don't have directories in there that are not part of that environment (though this could be fixed by just using `sys.path.pop(0)` twice).

blueyed:
Ok, luckily we have the complicated code already.. :)

blueyed:
btw: only one job failed, might be interesting: https://travis-ci.org/davidhalter/jedi/jobs/398806850

PR #1160, opened 2018-06-30
[postponed] Revisit pytest config

blueyed:
- add testpaths setting
- tox: remove testpaths from posargs default
- s/py.test/pytest/


davidhalter:
I like it! However note that until we drop Python 3.3 (probably the next minor release), I cannot merge this, because of this change:

```
commit 62915686afa37da21ff42605e5f8b9fd7bcdf6ab
Author: Dave Halter <davidhalter88@gmail.com>
Date:   Thu Dec 14 22:29:13 2017 +0100

    Don't use pytest 3.3+ because it removed support for Python 3.3
```

blueyed:
It is only used for our own tests, and apparently we're not testing against Python 3.3 - or Travis should have failed, no?

pytest.ini
```diff
@@ -10,3 +10,5 @@ norecursedirs = .* docs completion refactor absolute_import namespace_package
 # fine as long as we are using `clean_jedi_cache` as a session scoped
 # fixture.
 usefixtures = clean_jedi_cache
+
+testpaths = jedi test
```
blueyed: What do you think about moving this to setup.cfg (as `[tool:pytest]`)?
(There are ideas to deprecate setup.cfg for pytest though.)
Let's maybe just keep it here for now.

davidhalter: I would probably just keep it and see how that all develops. I feel like this is such a core pytest feature and nobody really seems to be using `setup.cfg` so I would just keep it as is.

davidhalter:
> It is only used for our own tests, and apparently we're not testing against Python 3.3 - or Travis should have failed, no?

I removed travis 3.3 tests, because of various issues with travis. Since it is still enabled by default in `tox.ini`, I will still see if 3.3 fails, because I always run tox before releases.

Issue #1096, opened 2018-04-19
Start using Typeshed


davidhalter:
Get typeshed working in Jedi. This is probably related to #839.

davidhalter:
Typeshed is actually used now in the typeshed branch. It's a few months time of work. Please test it!

PR #1161, opened 2018-06-30
[postponed] tox/Travis: test against Python 3.7

blueyed:


davidhalter:
I've tried it before. There's no chance to enable it for now. The problem is basically that they removed a lot of docstrings in favor of argument clinic (without str return converters) that Jedi is relying on.

My goal to work around this issue is to use typeshed (#1096). I have it high on my priority list (since half a year...), but it's quite a bit of work.

See:
https://mail.python.org/pipermail/python-dev/2018-March/152476.html
https://mail.python.org/pipermail/python-dev/2018-March/152484.html
The thread is a bit longer, but these are the most essential messages.

yan12125:
I'm also interested in this, so I [patched jedi to allow python3.7](https://github.com/yan12125/jedi/commit/6e119af9bfcd755c91bec3cde3fe00157f9870e3) and check the status of the test suite under Python 3.7. Besides missing return types for builtin methods, there are at least two CPython changes that break the jedi test:

* A new `breakpoint` function ([PEP-553](https://www.python.org/dev/peps/pep-0553/)), which results in additional completions for integration tests for `break`.
* Completions for `namedtuple`s no longer work as `collections._class_template` is dropped (https://github.com/python/cpython/pull/3454).

In my trial run, there are [89 failures](https://dl.chyen.cc/jedi-pytest-log.txt) in total. As there are so many internal changes in this new Python version, how about opening an issue tracking Python 3.7 breakages?

blueyed:
> how about opening an issue tracking Python 3.7 breakages?

Sounds good to me.

We can use this PR/branch here then to work on it.
I've cherry-picked your https://github.com/yan12125/jedi/commit/6e119af9bfcd755c91bec3cde3fe00157f9870e3, and rebased it - hopefully Travis has Python 3.7 by now.

yan12125:
> I've cherry-picked your yan12125/jedi@6e119af, and rebased it

Thanks!

> hopefully Travis has Python 3.7 by now.

Another option might be testing Python 3.7 with Appveyor, which support Python 3.7: https://www.appveyor.com/docs/build-environment/#python

blueyed:
@yan12125 
Added in 26bd0ce.

daskol:
What should be done in order to fix the issue? Python 3.7 was released almost three months ago but jedi still does not work with third-party packages.

Why this b7e79b4096b574d8ff726d321c547b6f094f3396 commit is not an option?

sawan:
Is this problem in ```elpy``` related to this issue?

[Code completion not working for installed packages](https://github.com/jorgenschaefer/elpy/issues/1474)

davidhalter:
@sawan Django definitely works. You just need to upgrade to latest Jedi and choose the right `jedi.Environment`.

sawan:
> @sawan Django definitely works. You just need to upgrade to latest Jedi and choose the right `jedi.Environment`.

@davidhalter  Thank you for your reply, I will upgrade to 0.13 and report back.

sawan:
> @sawan Django definitely works. You just need to upgrade to latest Jedi and choose the right `jedi.Environment`.

Upgrading to ```jedi 0.13``` seems to have fixed the problem.

davidhalter:
Ok. I'm not sure how far you actually were with this PR.

I changed some things in the typeshed branch. appveyor tests are now passing with 3.7 and travis tests are missing 3.7, but are also passing.

I'd be very happy to merge this PR if there's a way anyone gets it working. I pretty much don't care about how it works, ist should just include `JEDI_TEST_ENVIRONMENT=37` as an env variable and `- 3.7` in the `python` section.

blueyed:
Rebased against master for now - let's see how it turns out.

Created https://github.com/davidhalter/jedi/pull/1289 to switch to Ubuntu Xenial independently.

davidhalter:
I'm closing this, since 3.7 was enabled using other code. Still thanks for trying this here. I feel like I merged some other code of yours that got rid of `./travis_install.sh` and updated to xenial. So I guess this is not relevant anymore.

Thanks again!

PR #1162, opened 2018-07-01
WIP: improve docs

blueyed:
TODO:

- [ ] fix all warnings reported by `sphinx-build -n` (~20 left)
- [ ] add tox env to build docs, use picky mode on Travis there then.  Could have just two envs, or pass `-n` on Travis then.  `-W` is already used to exit non-zero on warnings, but currently Sphinx displays the first warning then only (https://github.com/sphinx-doc/sphinx/issues/4614)

### DIFF
--- a/jedi/evaluate/recursion.py
+++ b/jedi/evaluate/recursion.py
@@ -3,8 +3,8 @@ Recursions are the recipe of |jedi| to conquer Python code. However, someone
 must stop recursions going mad. Some settings are here to make |jedi| stop at
 the right time. You can read more about them :ref:`here <settings-recursion>`.
 
-Next to :mod:`jedi.evaluate.cache` this module also makes |jedi| not
-thread-safe. Why?  ``execution_recursion_decorator`` uses class variables to
+Next to the internal ``jedi.evaluate.cache`` this module also makes |jedi| not
+thread-safe, because ``execution_recursion_decorator`` uses class variables to
 count the function calls.
 
 .. _settings-recursion:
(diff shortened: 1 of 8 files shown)


jedi/evaluate/recursion.py
```diff
@@ -3,8 +3,8 @@
 must stop recursions going mad. Some settings are here to make |jedi| stop at
 the right time. You can read more about them :ref:`here <settings-recursion>`.
 
-Next to :mod:`jedi.evaluate.cache` this module also makes |jedi| not
-thread-safe. Why?  ``execution_recursion_decorator`` uses class variables to
+Next to the internal ``jedi.evaluate.cache`` this module also makes |jedi| not
+thread-safe, because ``execution_recursion_decorator`` uses class variables to
```
blueyed: The module `jedi.evaluate.cache` is not documented (included in the docs), therefore this does not get linked - I've added the "internal" there assuming that non-documented means "internal".

davidhalter: Makes sense.

davidhalter:
Is this ready to be merged? I think there's a lot of good things in here.

blueyed:
@davidhalter 
It is good already, but not finished - I want to fix all the warnings and then ensure that there are no new warnings in CI.
Feel free to merge it already, but it conflicts - so I have to look into it anyway.  Shouldn't be too much work to finish it.

davidhalter:
I merged this PR in the refactor branch and added a lot more doc improvements. Sphinx warnings should be gone.

> add tox env to build docs

I will try to get rid of tox this summer when I drop Python 2. It causes me more problems than it solves. I will probably just work with a Makefile at that point. So please don't put too much effort into that in the future :).

blueyed:
@davidhalter 
Thanks!
Haven't checked it there, but are you using nitpicky mode there?  This is really useful to avoid regressions with regard to missing links/references.

davidhalter:
I'm using `SPHINXOPTS='-n -W' make html` at the moment (because of your comments).

blueyed:
Great!
I forgot about this PR altogether apparently, but lately played more with docs (and Sphinx itself) again.

PR #1164, opened 2018-07-03
Revert using highest possible pickle protocol

micbou:
See discussion in PR https://github.com/davidhalter/jedi/pull/1108. Add a note that the version of the pickle protocol should be increased to 4 when Python 2.7 and 3.3 are dropped.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1164)


davidhalter:
@blueyed  What do you think? I think my comments in #1108 still apply.

blueyed:
I would say to keep it given that my plan work (https://github.com/davidhalter/jedi/pull/1108#issuecomment-401566779).
IIRC the problem was to have the `highest_pickle_protocol` function available in the client - that's when I've started doing the PYTHONPATH PR (#1159).

davidhalter:
I'll just leave this one open, let's see what's going to happen.

davidhalter:
Since @blueyed found a solution for the problem in #1108, I think we can safely close this one.

Issue #1165, opened 2018-07-05
subprocess client: do not mute sys.stderr


blueyed:
It gets muted in https://github.com/davidhalter/jedi/blob/508ed7e5b8efd946e6a6c6239ee926ce67c6d36a/jedi/evaluate/compiled/subprocess/__init__.py#L287-L291, but that results in e.g. (internal) exceptions not being visible anywhere.

What is the leaking that is mentioned there about?

I would say to either remove this, or add an outer `try-except` to the client in `__main__.py` to catch any exceptions and display them on saved_sys_stderr.

davidhalter:
Sometimes if you load certain modules, they write stuff to stderr. This should not leak IMO. However we could try to read stderr all the time and log the output. I think that would make more sense.

blueyed:
Ok, so you agree to unmute it then?

Do you mean to read and log it always, or just in case of errors, like it gets done now already, when it crashed?

blueyed:
See https://github.com/davidhalter/jedi/pull/1169.

davidhalter:
#1169 was merged.

PR #1169, opened 2018-07-10
subprocess: do not mute stderr

blueyed:
Fixes https://github.com/davidhalter/jedi/issues/1165.

### DIFF
--- a/jedi/evaluate/compiled/subprocess/__init__.py
+++ b/jedi/evaluate/compiled/subprocess/__init__.py
@@ -284,11 +284,10 @@ class Listener(object):
 
     def listen(self):
         stdout = sys.stdout
-        # Mute stdout/stderr. Nobody should actually be able to write to those,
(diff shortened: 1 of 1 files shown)


davidhalter:
Note that this won't change anything, because there's still `stderr=subprocess.PIPE`. IMO we should keep that, read what's in there (not sure how easily that will be possible) and log that.

blueyed:
> there's still stderr=subprocess.PIPE

This allows to read it after all, i.e. I think it should be kept.
Are you sure that this PR will not change anything?

davidhalter:
Well, this change will get capture stderr on the parent process side and you will not see anything.

davidhalter:
As you can see it in the commit, I think I addressed your concerns here. Please tell me if there's still something wrong.

blueyed:
Thanks, LGTM (although more complex, but I'm sure for reasons ;))

PR #1176, opened 2018-07-15
Second Try: Environment: improve/fix getting sys.prefix 

davidhalter:
Instead of #1108 I propose this change. It involves pretty much all of what @blueyed did there, but with a different approach about creating subprocesses. I think I cleaned up a mistake we made.

The issue was that if something changed about the environment (e.g. version
switch) or sys.path change, re-creating the environment was possible, but did
not involve the change. The environments have now a `__del__` function that
deletes the subprocess after every time an Environment is garbage collected.

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -3,14 +3,12 @@ Environments are a way to activate different Python versions or Virtualenvs for
 static analysis. The Python binary in that environment is going to be executed.
 """
 import os
-import re
 import sys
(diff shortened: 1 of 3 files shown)


blueyed:
LGTM, rebasing on master should fix AppVeyor then (py33), no?

davidhalter:
The PR passed, it's only the branch that failed, so I guess it was fixed by rebasing it.

PR #1178, opened 2018-07-15
Add params to CallSignature.__repr__

blueyed:
Looks like this for `jedi.Script` then:

> <CallSignature: Script index=0 params=[source=None, line=None, column=None, path=None, encoding='utf-8', sys_path=None, environment=None]>

`_params_str` could be made public, and then could be used in jedi-vim,
which currently has this:

    params = [p.description.replace('\n', '').replace('param ', '', 1)
              for p in signature.params]

https://github.com/davidhalter/jedi-vim/blob/08792d3fd742ec9d8cb75fe1e2a799244ee45fe3/pythonx/jedi_vim.py#L492-L493

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -638,9 +638,18 @@ class CallSignature(Definition):
         """
         return self._bracket_start_pos
 
+    @property
+    def _params_str(self):
+        return ', '.join([p.description[6:]
+                          for p in self.params])
+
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks. Regarding the possibility of making `_params_str` public, we have to refactor param support and call signatures anyway so I'll wait for that to happen first.

PR #1179, opened 2018-07-15
Some minor flake8 fixes

blueyed:


jedi/utils.py
```diff
@@ -96,7 +96,7 @@ def complete(self, text, state):
                     before = text[:len(text) - len(name)]
                     completions = interpreter.completions()
                     logging.debug("REPL completions: %s", completions)
-                except:
+                except Exception:
```
davidhalter: I don't want these fixed. It reraises anyway so please leave them there. We should probably just remove F707. I use it quite often and think it's the right way to treat some things if it has a `raise` in it.

blueyed: Fixed.

PR #1180, opened 2018-07-15
[RFC] Add qa env

blueyed:
Ignores tests with flake8 completely for now.

Many failures, so should not really get enabled yet, but might be a good base for local checking.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,4 @@ max-line-length = 100
 ignore =
   # do not use bare 'except'
   E722,
+exclude = test/*
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,9 @@ setup(name='jedi',
               # coloroma for colored debug output
               'colorama',
           ],
+          'qa': [
+              'flake8==3.5.0',
+          ],
       },
       package_data={'jedi': ['evaluate/compiled/fake/*.pym']},
       platforms=['any'],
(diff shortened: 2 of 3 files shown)


davidhalter:
I generally like it, if we enable it, we should probably also enable it on travis.

I already have flake8 running and ignore the warning about spacing around operators (because that's just not PEP8). Generally I would probably advice creating a configuration that most of Jedi works with. We could then start arguing which rules should be followed in the future and which ones not. I generally think that the PEP8 part of flake8 is pretty poorly done and could be improved tremendously. I even tried to do that in parso, but I'm still not sure how to do certain indentation rules. So I just never really made that public.

blueyed:
Sounds good.

What do you think about https://github.com/ambv/black/?
E.g. the pytest projects has adopted it, and has run just all of its code through it.

blueyed:
Do you prefer "line break before binary operator", or after?  (W503/W504)

I prefer the latter, i.e. we would ignore W503 then.

davidhalter:
Can you quickly look at the git commits on the typeshed branch? I unfortunately commited the changes to `../setup.cfg` there. W503 is already ignored there.

davidhalter:
Merged and seems to work. I'm now ignoring `F401` in `tox.ini`, because I still want that warning in general, but just not for the linter. I feel like this is really a flake8 problem, because the linter is just very simple. This is my workaround.

PR #1199, opened 2018-08-13
_get_virtual_env_from_var: use safe=False

blueyed:
Without this creating an env from VIRTUAL_ENV will always silently fail
if it is not the same/current environment.

TODO:

- [ ] I think it would be good to have some logging in case it would still fail for other reasons, but there is no `logger` yet.

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -147,13 +147,18 @@ class InterpreterEnvironment(_BaseEnvironment):
 
 
 def _get_virtual_env_from_var():
+    """Get virtualenv environment from VIRTUAL_ENV environment variable.
+
+    It uses `safe=False` with ``create_environment``, because the environment
+    variable is considered to be safe / controlled by the user solely.
+    """
     var = os.environ.get('VIRTUAL_ENV')
     if var is not None:
         if var == sys.prefix:
             return SameEnvironment()
 
         try:
-            return create_environment(var)
+            return create_environment(var, safe=False)
         except InvalidPythonEnvironment:
             pass
 
--- a/test/test_api/test_environment.py
+++ b/test/test_api/test_environment.py
@@ -110,3 +110,22 @@ def test_create_environment_venv_path(venv_path):
 def test_create_environment_executable():
     environment = create_environment(sys.executable)
     assert environment.executable == sys.executable
+
+
+def test_get_default_environment_from_env_does_not_use_safe(tmpdir, monkeypatch):
+    fake_python = os.path.join(str(tmpdir), 'fake_python')
+    with open(fake_python, 'w') as f:
+        f.write('')
+
+    def _get_subprocess(self):
+        if self._start_executable != fake_python:
+            raise RuntimeError('Should not get called!')
+        self.executable = fake_python
+        self.path = 'fake'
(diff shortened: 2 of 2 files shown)


blueyed:
Build failure is unrelated (I hope) - "speed issue": https://travis-ci.org/davidhalter/jedi/jobs/415510002#L654

blueyed:
@davidhalter 
I think this one is essential for making `$VIRTUAL_ENV` actually work, and would also warrant a bugfix release.

davidhalter:
I disagree with your premise. It's working very well for me for example. It's working as long as the virtualenv is considered safe. This is usually the case if it's a Python3 venv (created with `python -m venv`) or a virtualenv created environment that has the same binary as the system Python's.

The cases where it doesn't work are afaik the following:

1. venv: If the Python binary doesn't belong to `root` (i.e. is not considered safe).
2. virtualenv: If the Python binary doesn't match the system environments, e.g. some weird installations or I think also Python version upgrades.

So I agree that we might  need to change it, but I still have to think about this, because it's a bit more unsafe and if ever an attacker gets the option to set that path the attacker basically gains execution privileges when Jedi is started.

blueyed:
I've thought you've mentioned somewhere (in the code?) that an environment variable would be safe, i.e. an attacker is not able to control it.

For me the python executable is coming from pyenv, and therefore is owned by the user.

My system Python is 3.7, but I have a project where I still need/use 3.6, and therefore use pyenv for it.

davidhalter:
Could you please add a comment? Why safe is True at that point? I'll merge ASAP at that point.

IMO it's not clear why it's safe at that point if we don't give the information.

blueyed:
I would assume that it is safe because of your own comment here:
https://github.com/blueyed/jedi/blob/aae10c1bfc56911973fb83e6693bb2678830b1d5/jedi/api/environment.py#L215-L217

This however calls `_get_virtual_env_from_var` (which this PR changes).

Also relevant:
https://github.com/blueyed/jedi/blob/aae10c1bfc56911973fb83e6693bb2678830b1d5/jedi/api/environment.py#L380-L389

I think `VIRTUAL_ENV` is safe, since it is controlled by the user.  However, attackers might target it explicitly to overwrite `bin/python` in there, but this is not really a Jedi specific problem, because it is also in the PATH then typically already.

davidhalter:
I agree. I think we should just add this in the source code, so users that are reading it can understand.

blueyed:
Adjusted.
btw: the `safe` arg is a bit confusing / not really clear, and `assert_safe` might be better.

blueyed:
Do you have plans with regard to logging?  (see todo in original comment)

davidhalter:
Great! If you want to add logging, please create a separate pull request, I'm merging this, because it already is great.

Generally Jedi does not use `logging`. It uses it's own logging with `debug.py`. I'm not saying this is a good idea, but it has some advantages for highlighting (which you could maybe also do with logging and some custom work). So for now use that so we are consistent.

For the long-term, we can also just refactor everything to use logging, but that would involve porting some of the highlighting functionality to logging, because I'm using that pretty extensively to see what actually happens.

Issue #1200, opened 2018-08-13
get_cached_default_environment should be invalidated if $VIRTUAL_ENV changes


blueyed:
https://github.com/davidhalter/jedi/blob/691a5b6c21d97d97da6ef14224f1115cc13f9f3f/jedi/api/environment.py#L190-L192

This is used by jedi-vim, but when a user changes `$VIRTUAL_ENV` it should not take 10 minutes to work.

Not sure how to approach this best.

On the one hand a time based cache does not really make sense here, except for when you install a new version, but then it might still take up to 10 minutes to have an effect.

Therefore I suggest changing it to only cache it by the value of `$VIRTUAL_ENV`.

davidhalter:
I agree. We could probably just make it possible to clean the cache, rename this function and create a function that checks if `$VIRTUAL_ENV` changed and in that case clear the cache.

blueyed:
So it could be cached on the module, and a new function would be added to clear the cache?

PR #1201, opened 2018-08-14
get_cached_default_environment: cache by VIRTUAL_ENV only

blueyed:
Fixes https://github.com/davidhalter/jedi/issues/1200

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -9,7 +9,7 @@ import filecmp
 from collections import namedtuple
 
 from jedi._compatibility import highest_pickle_protocol, which
-from jedi.cache import memoize_method, time_cache
+from jedi.cache import memoize_method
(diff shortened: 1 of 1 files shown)


davidhalter:
What I meant to say in #1200 was this:

```
def get_cached_default_environment():
    environment = _get_cached_default_environment()
    if environment.prefix != <virtualenv_variable>:
        _get_cached_default_environment.reset()
        return _get_cached_default_environment()
    return environment

# Rename 
@time_cache(seconds=10 * 60)
def _get_cached_default_environment():
    return get_default_environment()

```

Obviously you would also need to add the reset function for time_cache. It doesn't exist, yet.

davidhalter:
I changed it now in a bit of a different way, but thanks still for your initial thoughts!

Issue #1202, opened 2018-08-14
jedi.api.names with all_scopes=False extracts a class-scoped variable


immerrr:
Here's a repro example:
```
In [8]: jedi.api.names(textwrap.dedent("""
   ...:     from module import f
   ...:     g = f(f)
   ...:     class C:
   ...:         h = g
   ...: """), 'example.py', all_scopes=False)
Out[8]: 
[<Definition module f>,
 <Definition g = f(f)>,
 <Definition class C>,
 <Definition h = g>]

In [9]: _8[-1].parent()
Out[9]: <Definition class C>
```
IIUC `<Definition h = g>` should not be considered a global definition and thus not returned if `all_scopes=False` is used.

davidhalter:
I agree. If you PR this as a breaking test somewhere I will probably have it easier to fix it.

Thanks for the report!

immerrr:
@davidhalter there you go: #1203

davidhalter:
Fixed by yourself, thanks again!

PR #1203, opened 2018-08-16
Fix jedi.api.names(..., all_scopes=False) returning class and function-level definitions 

immerrr:
This PR adds a fix for #1202

### DIFF
--- a/jedi/evaluate/helpers.py
+++ b/jedi/evaluate/helpers.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 from parso.python import tree
 
 from jedi._compatibility import unicode
-from jedi.parser_utils import get_parent_scope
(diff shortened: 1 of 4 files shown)


immerrr:
@davidhalter I seem to have fixed the problem described in #1202. Let me know if it looks OK.

Ideally, I would change `get_parent_scope` to use `is_name_of_func_or_class_def` as described in the comment, but tests start failing when I do it, and apparently it influences flow and reachability checks, so I left that out of the scope of this PR.

immerrr:
@davidhalter could you have a look at this? I would like your opinion on the previous comment.

davidhalter:
Thanks for contributing this. I'm really sorry for not keeping up with pull requests. It's been a lot of work and I was tired. Sorry.

PR #1212, opened 2018-08-31
Use expanded paths when looking for virtualenv root

tacho:
This fixes virtualenv resolution under macOS and Pipenv.

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -154,7 +154,11 @@ def _get_virtual_env_from_var():
     """
     var = os.environ.get('VIRTUAL_ENV')
     if var:
-        if var == sys.prefix:
+        # Under macOS in some cases - notably when using Pipenv - the
+        # sys.prefix of the virtualenv is /path/to/env/bin/.. instead of
+        # /path/to/env so we need to fully resolve the paths in order to
+        # compare them.
+        if os.path.realpath(var) == os.path.realpath(sys.prefix):
             return _try_get_same_env()
 
         try:


jedi/api/environment.py
```diff
@@ -149,7 +149,11 @@ def get_sys_path(self):
 def _get_virtual_env_from_var():
     var = os.environ.get('VIRTUAL_ENV')
     if var is not None:
-        if var == sys.prefix:
+        # Under macOS in some cases - notably when using Pipenv - the
+        # sys.prefix of the virtualenv is /path/to/env/bin/.. instead of
+        # /path/to/env so we need to fully resolve the paths in order to
+        # compare them.
+        if os.path.realpath(var) == os.path.realpath(sys.prefix):
```
blueyed: Any idea why that happens?
Is it related to certain Python versions maybe?

blueyed: And could you investigate why the fallback of `create_environment` below fails for you?
Is it because of an unsafe env (https://github.com/davidhalter/jedi/pull/1199), or something else?

Yasushi: `VIRTUAL_ENV` is set in the configuration file as follows:
```
VIRTUAL_ENV="/Users/yasushi/.local/share/virtualenvs/venvtest-BgwP3k5I"
```
But `sys.prefix' is:
```
% python -c 'import sys; print(sys.prefix)'
/Users/yasushi/.local/share/virtualenvs/venvtest-BgwP3k5I/bin/..
```

I think that some kind of normalization is necessary like this pull request.

blueyed: How did you create the virtualenv?
What is "configuration file"?
What is `which python` / `command -v python`?
What is `python -V``

blueyed:
I agree with @davidhalter here.

Can you provide steps where `sys.prefix` ends up being that weird?
(I suspect that pipenv is doing something here in this regard)

Yasushi:
I tried it with venv. There seems to be no problem. It sounds like a problem with pipenv. 

```
yasushi@irpa% python3 -V
Python 3.7.0
yasushi@irpa% type python3
python3 is /usr/local/bin/python3
yasushi@irpa% python3 -m venv venvtest
yasushi@irpa% source ./venvtest/bin/activate
(venvtest) yasushi@irpa% echo $VIRTUAL_ENV
/Users/yasushi/src/python/venvtest/venvtest
(venvtest) yasushi@irpa% python -c 'import sys; print(sys.prefix)'
/Users/yasushi/src/python/venvtest/venvtest
```

blueyed:
Another open question is why the fallback does not work (https://github.com/davidhalter/jedi/pull/1212/files#diff-c204f420c3d90e6c005f6354c65dc638R162).
Can you print the exception there, please?

Yasushi:
@blueyed 
Firstly, this problem seemed to be fixed in jedi 0.13.1.

In jedi 0.12.1, the exception "The python binary is potentially unsafe." was raised.

blueyed:
Thanks for the update.
The later error appears to be fixed via 56bd7951 then.

But I assume we should still handle the initial issue (likely caused by / triggered by pipenv only).

@Yasushi 
Is this still an issue with pipenv after all?

Yasushi:
@blueyed 
I am using jedi via pyls and I can use it without problem with the latest version. Thank you!

Strange sys.prefix seems to be made by virtualenv (! = venv). And pipenv uses virtualenv.
Unfortunately, I did not understand why sys.prefix got strange.

davidhalter:
Ok. Is there still an issue or have we fixed everything here? I will close next time I look at this PR, if nobody has this issue anymore. It feels like some things were fixed in 0.13.1.

davidhalter:
This is fine and I'm willing to merge. I was just really wondering what was happening. Please resolve the conflicts and I will merge.

tacho:
Thanks, conflicts resolved.

davidhalter:
Thanks!

PR #1215, opened 2018-09-09
Faling test: Definition.parent() fails for Comprehension variables

gatesn:
Managed to reproduce. Haven't looked into a fix yet.

### DIFF
--- a/test/test_api/test_classes.py
+++ b/test/test_api/test_classes.py
@@ -289,6 +289,20 @@ def test_parent_on_completion(Script):
     assert parent.type == 'class'
 
 
+def test_parent_on_comprehension():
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks! I'll merge once I'm going to fix it.

davidhalter:
Was finally able to fix this. I was just doing a huge refactoring (159 files changed, 3962 insertions(+), 3866 deletions(-)) and realized that comprehension contexts are a bit tricky, because they don't really have a "name". Then I realized that you wrote this breaking test a while ago. So it should be all good now :)

gatesn:
That's great! Thank you

Issue #626, opened 2015-09-25
Allow specialized plugins to help Jedi understand metaclasses (and other things)


patrys:
Take Django for example. It uses a lot of metaclasses to support declarative syntax. Would it befeasible for Jedi to let a specialized Django plugin re-write a Model AST node before parsing it to simulate what its metaclass would do?

davidhalter:
Hmm I would actually say that it should be a Jedi plugin (a library called [jedi-django]) or even namespace packages. However, it's going to be difficult to design an API.

immerrr:
I actually kept a similar idea boiling in my head for quite a while now. The idea was to cleanly separate things that Jedi can confidently reason about and the rest. For example, jedi can trace the path of a function argument within a function, but it cannot with certainty say where did the argument come from in the first place. That would be an "entrypoint".

The simplest way to think about "entrypoints" was to take a module — since Jedi as of now operates on per-module basis — and try to find all the places that could change about that module, that would be, from the top of my head:
- the name of the module in question, which depends on import path
- imported modules and objects from those modules, which depend on import path and hooks
- arguments of publicly accessible functions (local functions should be OK)
- values of module-accessible variables (external code might inject variables or change existing ones)

Once we identify an "entrypoint", there should be a way for a plugin to help Jedi core to resolve that entrypoint using heuristics or other a priori knowledge. For example, docstring parser that tries to guess types of the variables would become a plugin. There also could be a plugin that would scan a certain directory and subdirectories looking for usages of a function and returning found argument types. Or a plugin that traces the execution of a certain script and returns arguments found that way.

In that scenario, a metaclass might simply become another entrypoint to be handled by plugins.

davidhalter:
Hmm ok.

@immerrr Have you seen the `evaluate/stdlib.py` module? Would that be a good place for entry points? Upon function/class execution? Or would it be better if name resolution already would trigger it?

immerrr:
Built-in (as in written in C, not necessarily part of the standard library) objects and modules (being objects themselves) are indeed prime examples of entry points.  Built-in functions, well, they are interesting beasts. 

[...]

PR #1254, opened 2018-11-12
Allow extensions for project specific imports

and-fs:
I had the problem to make Jedi able to import our dynamic packages.

Those packages are created by our Interpreter (we embed python) in C and are not physically available (the modules are, the package is not).

It does something like this:

```
import sys, types
mod = types.ModuleType("mylib", "doc for mylib")
mod.__path__ = ['<path to the modules>']
sys.modules["mylib"] = mod
```

Now you can import the files located in the given path(s) by typing:

```
from mylib import mymodule
```

For this I added an extension package which is used by the Importer for failed imports.
This simply looks for extension modules (see the extension package docs) and make them available for the Importer.

PS: this is my first pull request, so sorry for eventually mistakes. Thanx!


and-fs:
Sorry, forgot to remove the Py3 specific debug output function. How can I update this request?

and-fs:
Forget the question, got it (simply commit)

and-fs:
There is something wrong with the libraries - the Py27 tests are using paths from Py34, here the stack trace of the Py27 run from the AppVeyor build:

`E           TypeError: Traceback (most recent call last):
E             File "C:\projects\jedi\jedi\evaluate\compiled\subprocess\__init__.py", line 344, in listen
E               result = False, None, self._run(*payload)
E             File "C:\projects\jedi\jedi\evaluate\compiled\subprocess\__init__.py", line 318, in _run
E               return function(evaluator, *args, **kwargs)
E             File "C:\projects\jedi\jedi\evaluate\compiled\subprocess\functions.py", line 73, in list_module_names
E               for module_loader, name, is_pkg in iter_modules(search_path)
E             File "C:\projects\jedi\jedi\evaluate\compiled\subprocess\functions.py", line 72, in <listcomp>
E               force_unicode(name)
E             File "C:\projects\jedi\jedi\_compatibility.py", line 192, in _iter_modules
E               modname = inspect.getmodulename(fn)
E             File "C:\Python34\lib\inspect.py", line 563, in getmodulename
E               if fname.endswith(suffix):
E           TypeError: endswith first arg must be bytes or a tuple of bytes, not str
jedi\evaluate\compiled\subprocess\__init__.py:263: TypeError`

When I execute the tests locally everything is still fine...

davidhalter:
I'm not yet looking at your code, I just want to understand your problem better, first.

Why don't you call the source folder `/foo/bar/mylib` and give Jedi the `sys.path` + ['/foo/bar']`? This would solve pretty much all issues you are having, wouldn't it?

and-fs:
Due to the following:

1. We're using Jedi not directly but in VS Code.
2. Our python is embedded in a huge C++-App
3. Due to historical reasons we have an own import mechanism which is similar to the Py 3 dynamic packages but using special names as identifier (instead of ., .. aso.), its a bit like
`import rootmodule.modulename`
`import parentmodule.modulename`
`import thismodule.modulename` 
where rootmodule, parentmodule and thismodule depent on the 
location of the importing module

4. Dynamic namespace packages seem not to work proberly resp. cannot be configured when used in VS Code, something like:
`import types`
`mylib = types.ModuleType('mylib', 'Dynamic module.')`
`mylib.__path__ = [path1, path2, ...]`
`sys.modules['mylib '] = mylib`

davidhalter:
Can you explain (3) a bit more? I'm not sure I get it. Just try to make a simple example of the folder structure and two different imports you would be doing.

and-fs:
Hi David,

sorry for the late answer, due to the lack of developers I was involved in some time consuming projects and didn't have time for this one.

Here an example for 3.):

<pre>
+-modulepath/
  |
  +-mod_a /
  |  |
  |  +-mod_b /
  |  |  |
  |  |  +-_pyslot /
  |  |  |  |
  |  |  |  +-pymod_b_1.py   # parentmodule = mod_a, thismodule = mod_b
  |  |  |  +-pymod_b_2.py
  |  |  |
  |  |  +-_testslot /
  |  |  |  |
  |  |  |  +-testdoc.txt
  |  |  |
  |  |  +-__modinit__.py    # parentmodule = mod_a, thismodule = mod_b
  |  |  
  |  +-_pyslot /
  |  |  |
  |  |  +-pymod_a_1.py      # parentmodule = n/a, thismodule = mod_a
  |  |  +-pymod_a_2.py
  |  |
  |  +-__modinit__.py
  |
  +-mod_c /
  |  |
  |  +-_pyslot /
  |  |  |
  |  |  +-pymod_c_1.py     # parentmodule = n/a, thismodule = mod_c
  |  |  +-pymod_c_2.py
  |  |
  |  +-__modinit__.py      # parentmodule = n/a, thismodule = mod_c
</pre>

This could be a structure for the described case, each package has a ``__modinit__.py`` file to do some initialization (like the ``__init__.py`` for packages) and one or more slots starting with ``_``, but only the ``_pyslot`` contains importable Python files.

Now some examples:

```python
# mod_a/__modinit__.py
from thismodule import pymod_a_1 # refers to /mod_a/_pyslot/pymod_a_1.py
from thismodule.mod_b import pymod_b_1 # refers to /mod_a/mod_b/_pyslot/pymod_b_1.py 
```

```python
# mod_a/mod_b/__modinit__.py
from thismodule import pymod_b_1 # refers to /mod_a/mod_b/_pyslot/pymod_b_1.py
from parentmodule import pymod_a_1 # refers to /mod_a/_pyslot/pymod_a_1.py
```

```python
# mod_a/mod_b/_pyslot/pymod_b_1
from thismodule import pymod_b_2 # refers to /mod_a/mod_b/_pyslot/pymod_b_2.py
from parentmodule import pymod_a_1 # refers to /mod_a/_pyslot/pymod_a_1.py
from rootmodule.mod_c import pymod_c_1 # refers to /mod_c/_pyslot/pymod_c_1.py
```

Each _pyslot can also contain "normal" python packages where the standard import mechanism works.

Hopefully I could describe it understandably (and don`t forget: this was designed at the early stage of python 2, somewhen in 2006/2007 - I know that some things could be solved better with currently available import mechanisms, but we have too much projects using this structures for easily changing it).

Thank you!

davidhalter:
Ok. Thanks for your explanation! I think this is a clear case for the plugin architecture that we might eventually implement (#626). There is already a plugin architecture in that you could try to use, it might just change a lot, because this is not released: `jedi/plugins`. I think the imports modifications you are doing could be possible in there.

However I'm deciding to not use your code here. Plugins are the solutions for this. I'm sorry. Thanks anyway for your contribution. It actually looks like code that could have worked :)

PR #1260, opened 2018-12-05
Add fuzzy completion.

and3rson:
I've added a simple fuzzy completion. It can be enabled by setting `settings.fuzzy_completion` to True.

I enable it in `jedi-vim` as follows (tell me if there's a better way, I couldn't find one):

```vimscript
" .vimrc
au FileType python :py3 from jedi import settings; settings.fuzzy_completion = 1
```

This code can be improved even more by using Levenshtein string distance - I've used it before and it was really very accurate and unobtrusive. As for now, I've decided to keep it minimalistic & clean.

This PR implements https://github.com/davidhalter/jedi-vim/issues/72

### DIFF
--- a/jedi/api/completion.py
+++ b/jedi/api/completion.py
@@ -1,3 +1,5 @@
+import re
+
 from parso.python.token import PythonTokenTypes
 from parso.python import tree
 from parso.tree import search_ancestor, Leaf
@@ -33,7 +35,13 @@ def filter_names(evaluator, completion_names, stack, like_name):
         if settings.case_insensitive_completion:
             string = string.lower()
 
-        if string.startswith(like_name):
+        if settings.fuzzy_completion:
+            pattern = '.*'.join(like_name) + '.*'
+            match = re.match(pattern, string)
+        else:
+            match = string.startswith(like_name)
+
+        if match:
             new = classes.Completion(
                 evaluator,
                 name,
--- a/jedi/settings.py
+++ b/jedi/settings.py
@@ -16,6 +16,7 @@ Completion output
 ~~~~~~~~~~~~~~~~~
 
 .. autodata:: case_insensitive_completion
+.. autodata:: fuzzy_completion
 .. autodata:: add_bracket_after_function
 .. autodata:: no_completion_duplicates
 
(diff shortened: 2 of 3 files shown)


and3rson:
The error seems unrelated to this PR and is caused by deprecated usage of `pylint.marks.parametrize` in `test/test_evaluate/test_imports:81`. More info here: https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

I've added a requirement of pylint<4.0.0 to `setup.py`.

jedi/api/completion.py
```diff
@@ -33,7 +35,13 @@ def filter_names(evaluator, completion_names, stack, like_name):
         if settings.case_insensitive_completion:
             string = string.lower()
 
-        if string.startswith(like_name):
+        if settings.fuzzy_completion:
+            pattern = '.*'.join(like_name) + '.*'
+            match = re.match(pattern, string)
```
davidhalter: This doesn't match `foobar` if you type `bar`. Is this intended or did you want to use `re.search` instead of `re.match`? Otherwise I like it very much. Please remove the last `.*` (it's not needed) and use `.*?` to use lazy matching, which I'm pretty sure is more efficient.

Also I would probably move the generation of the pattern to the top with `re.compile`.

Thanks for the PR!

jedi/settings.py
```diff
@@ -62,6 +63,11 @@
 The completion is by default case insensitive.
 """
 
+fuzzy_completion = False
+"""
+The completion is non-fuzzy by default.
+"""
+
 add_bracket_after_function = False
```
davidhalter: I'm trying to get rid of the settings file more and more. It's a bad idea and params are better. Could you just use it as a param for `completions(fuzzy=False)`?

Also please rename completions for this and create this function:
```
def completions(self, **kwargs):
    return self._completions(**kwargs)

def _completions(self, fuzzy=False): ...
```

This makes it possible to use kwargs only arguments in Python 2 and it will be easy to migrate once Jedi is Python 3 only.

davidhalter:
Fixed in #1409 by @jmfrank63.

PR #1262, opened 2018-12-07
Use `pytest.param` when marking single parameters

hoefling:
Marking single parameters in [`pytest.mark.parametrize`](https://docs.pytest.org/en/latest/deprecations.html#marks-in-pytest-mark-parametrize) was deprecated in 3.2 and is removed in the recent 4.0 release, which makes the collection of `test_evaluate/test_imports.py::test_flask_ext` fail. This PR makes the parameters compliant with `pytest>=4`.

Also, since the `pytest.param` feature was introduced in 3.1.0, the `pytest` dependency need to be bumped.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(name='jedi',
       install_requires=install_requires,
       extras_require={
           'testing': [
-              'pytest>=2.3.5',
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks for fixing this!

PR #1270, opened 2018-12-31
Expand user directory when creating environments

rosslannen:
Adds support for creating environments with a path beginning with `~`. Supplying paths with a `~` is default for emacs when using `pythonic-activate` from https://github.com/proofit404/pythonic.


davidhalter:
I'm not sure we want that. This would make it impossible to work in this directory: `os.listdir('~root/')`.

I feel like this should be the responsibility of libraries.

davidhalter:
After thinking about this a bit more, I still think libraries are responsible for this. Thanks anyway for trying to contribute!

PR #1282, opened 2019-02-12
fix get_system_environment misses if same python version has multiple installs

mlangkabel:
The Environment.__init__ may throw an InvalidPythonEnvironment if the call to _get_subprocess() fails. In this case other registered Python environments of the same Python version may still work and shall also be considered.


davidhalter:
Thanks a lot! Sorry for making you wait so long.

mlangkabel:
Wow, thanks for getting back to this one! Actually I thought it wouldn't be merged anymore, because the CI was failing.

davidhalter:
The error was because of something else ;-)

PR #1285, opened 2019-02-15
fixing permission denied errors with project.json

izcram:
We're running into issues with `jedi` on our cluster which uses automount: 
```
PermissionError: [Errno 13] Permission denied: /cluster/home/.jedi/project.json
```

This happens when jedi is walking up the FS tree in order to find a `project.json` file. With automount, an attempt to open or stat a file in the `/cluster/home` directory will create it, however, will not be accessible. This makes jedi unusable in such an environment. This PR fixes the issue for us.

Also besides this (exotic) automount issue it seems to make sense to behave in the same way when encountering `PermissionError` as with `FileNotFoundError` and `NotADirectoryError`, since the file is not accessible in either case.

### DIFF
--- a/jedi/_compatibility.py
+++ b/jedi/_compatibility.py
@@ -348,6 +348,11 @@ try:
 except NameError:
     NotADirectoryError = IOError
 
+try:
+    PermissionError = PermissionError
+except NameError:
+    PermissionError = IOError
+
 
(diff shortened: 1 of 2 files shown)


davidhalter:
Absolutely! Thanks!

PR #1289, opened 2019-02-22
[WIP] ci: Travis: dist=xenial

blueyed:
TODO:

- [ ] keep nightly / use 3.8-dev?!


davidhalter:
If you can PR against typeshed I'm happy to merge it. Please also remove travis_install.sh if you're doing it inline.

davidhalter:
I'd still be very happy to merge something like this. It's just that I'm not sure what to do with the conflict.

The typeshed branch is now on master and will be the future of Jedi. There's also https://github.com/davidhalter/jedi/pull/1161/files.

blueyed:
Rebased.  Should probably keep the `python: nightly` job though - or rather `3.8-dev`?

davidhalter:
I like nightly better than 3.8-dev. It's the dev version anyway.

blueyed:
Ok.

It does not work with some envs currently (but also before (https://travis-ci.org/davidhalter/jedi/builds/497430746)) - forgot about that, and thought only a rebase would be required here.  So it needs more investigation.

davidhalter:
I think it's just when JEDI_TEST_ENVIRONMENT=34. I can try to fix that.

davidhalter:
Fixed it. In the end it was a very small change: 4aa91efc2e951ef412cf39cfb5c56152d39b5e46

Thanks a lot again. This helps a lot.

PR #1290, opened 2019-02-23
tox/Travis: revisit coverage reporting

blueyed:
This is an intermediate step on getting coverage with more jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
   include:
     - python: 3.6
       env:
-        - TOXENV=cov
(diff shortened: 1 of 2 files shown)


davidhalter:
Can you please check the typeshed branch and try to do a PR there? I remember changing some of these things (and since the typeshed branch is the future I would like to have it there :) Sorry!

davidhalter:
Thanks again for trying this here. However I moved away from tox. Feel free to revisit coverage again with `.travis.yml`.

PR #1295, opened 2019-03-01
Fix all resource warnings

micbou:
Environments currently generate resource warnings at deletion. For example, running the script
```python
import jedi
environment = jedi.get_system_environment('3.6')
```
with all warnings enabled gives that output:
```
$ python -Wa script.py
...
sys:1: ResourceWarning: unclosed file <_io.BufferedWriter name=3>
sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=4>
```
While these warnings can be considered harmless, they generate noise when running tests in projects that depend on Jedi.

This PR enables all warnings in tests through the `PYTHONWARNINGS` environment variable and fix them by closing the file descriptors that are left open. [Since `__del__` is not guaranteed to be called when the interpreter exits](https://docs.python.org/3/reference/datamodel.html#object.__del__), we use `atexit` to ensure that the subprocess streams are closed in that case. We also need to wait for the thread reading `stderr` to be terminated (which necessarily happen if the subprocess is killed) before being able to close the stream.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1295)

### DIFF
--- a/jedi/evaluate/compiled/subprocess/__init__.py
+++ b/jedi/evaluate/compiled/subprocess/__init__.py
@@ -12,7 +12,6 @@ import sys
 import subprocess
 import socket
 import errno
-import weakref
 import traceback
 from functools import partial
 from threading import Thread
@@ -22,7 +21,7 @@ except ImportError:
     from Queue import Queue, Empty  # python 2.7
 
 from jedi._compatibility import queue, is_py3, force_unicode, \
-    pickle_dump, pickle_load, GeneralizedPopen
+    pickle_dump, pickle_load, GeneralizedPopen, weakref
 from jedi import debug
 from jedi.cache import memoize_method
 from jedi.evaluate.compiled.subprocess import functions
@@ -37,7 +36,6 @@ _MAIN_PATH = os.path.join(os.path.dirname(__file__), '__main__.py')
 def _enqueue_output(out, queue):
     for line in iter(out.readline, b''):
         queue.put(line)
-    out.close()
 
 
 def _add_stderr_to_debug(stderr_queue):
@@ -54,6 +52,22 @@ def _add_stderr_to_debug(stderr_queue):
 
 def _get_function(name):
     return getattr(functions, name)
+
+
+def _cleanup_process(process, thread):
+    try:
+        process.kill()
+        process.wait()
+    except OSError:
+        # Raised if the process is already killed.
+        pass
+    thread.join()
+    for stream in [process.stdin, process.stdout, process.stderr]:
+        try:
+            stream.close()
+        except OSError:
+            # Raised if the stream is broken.
+            pass
 
 
 class _EvaluatorProcess(object):
@@ -145,6 +159,7 @@ class CompiledSubprocess(object):
     def __init__(self, executable):
         self._executable = executable
         self._evaluator_deletion_queue = queue.deque()
+        self._cleanup_callable = lambda: None
 
     def __repr__(self):
         pid = os.getpid()
@@ -182,6 +197,12 @@ class CompiledSubprocess(object):
         )
         t.daemon = True
         t.start()
+        # Ensure the subprocess is properly cleaned up when the object
+        # is garbage collected.
+        self._cleanup_callable = weakref.finalize(self,
+                                                  _cleanup_process,
+                                                  process,
+                                                  t)
         return process
 
     def run(self, evaluator, function, args=(), kwargs={}):
@@ -202,18 +223,7 @@ class CompiledSubprocess(object):
 
     def _kill(self):
         self.is_crashed = True
-        try:
-            self._get_process().kill()
-            self._get_process().wait()
-        except (AttributeError, TypeError):
-            # If the Python process is terminating, it will remove some modules
-            # earlier than others and in general it's unclear how to deal with
-            # that so we just ignore the exceptions here.
-            pass
-
-    def __del__(self):
-        if not self.is_crashed:
-            self._kill()
+        self._cleanup_callable()
 
     def _send(self, evaluator_id, function, args=(), kwargs={}):
         if self.is_crashed:
--- a/test/run.py
+++ b/test/run.py
@@ -357,9 +357,11 @@ def collect_dir_tests(base_dir, test_files, check_thirdparty=False):
             path = os.path.join(base_dir, f_name)
 
             if is_py3:
-                source = open(path, encoding='utf-8').read()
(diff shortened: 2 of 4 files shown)


test/run.py
```diff
@@ -353,9 +353,11 @@ def collect_dir_tests(base_dir, test_files, check_thirdparty=False):
             path = os.path.join(base_dir, f_name)
 
             if is_py3:
-                source = open(path, encoding='utf-8').read()
+                f = open(path, encoding='utf-8')
```
davidhalter: If we're already cleaning up messes, please use `with open(path, encoding='utf-8') as f`. My bad for not doing that.

tox.ini
```diff
@@ -10,6 +10,8 @@ setenv =
 # https://github.com/tomchristie/django-rest-framework/issues/1957
 # tox corrupts __pycache__, solution from here:
     PYTHONDONTWRITEBYTECODE=1
+# Enable all warnings.
+    PYTHONWARNINGS=always
```
davidhalter: I like!

jedi/evaluate/compiled/subprocess/__init__.py
```diff
@@ -56,6 +56,23 @@ def _get_function(name):
...
 
 
+def _clean_process(process, thread):
+    try:
+        process.kill()
+        process.wait()
+    except (AttributeError, TypeError, OSError):
+        # If the Python process is terminating, it will remove some modules
+        # earlier than others and in general it's unclear how to deal with
+        # that so we just ignore the exceptions here.
```
davidhalter: I like it a lot that you found `atexit`. I tried to search for exactly that, but couldn't find it. IMO this is way cleaner than what I'm doing here (what I'm doing here is just an aberration).

I feel like if we have that, there's no reason anymore to catch any errors. I'm only catching these errors, because `__del__` is called when some objects like the module `signal` have already been garbage collected. I feel like your approach is mostly right.

However if we want to get rid of catching the exceptions, we have to use something like

```
def __init__(self):
    self._cleanup_callback = functools.partial(_cleanup_process, process, t)
    atexit.register(self._cleanup_callback)

def _kill(self):
    self.is_crashed = True
    atexit.unregister(self._cleanup_callback)  # This is the important line
    self._cleanup_callback()
```

If we don't unregister, it will try to close everything again for a lot of potential processes - when Jed is running for a long time. I would like to avoid that. However this looks better IMO. Probably we also need to check if the process was already cleaned up.

If you don't want to do that - or have no time, feel free to leave it as it is. It's better than what we have before anyway :)

test/run.py
micbou: Done.

jedi/evaluate/compiled/subprocess/__init__.py
micbou: > I feel like if we have that, there's no reason anymore to catch any errors.

I can try removing `AttributeError` and `TypeError` but I think `OSError` is still needed (I added it because I was seeing this exception at the end of tests on AppVeyor).

> If we don't unregister, it will try to close everything again for a lot of potential processes - when Jed is running for a long time.

The issue is that `atexit.unregister` is not available on Python 2. Let me figure something out.

micbou: A solution is to use [`weakref.finalize`](https://docs.python.org/3.5/library/weakref.html#weakref.finalize) instead of `atexit`. It's not available on Python 2 so we need to backport it. The backport code is from https://gist.github.com/florentx/52dc3d00feb513d10838 (which is based on the official implementation of `weakref.finalize`). PR is updated.

davidhalter: Ok. How about only using atexit.unregister for Python3? I feel like it's a clean way and Python 2 is leaving this world in a year :)

davidhalter: This would obviously mean that we still have to have the solution we're currently using for Python2, but that's fine IMO.

micbou:
@davidhalter Did you miss my comment about using [`weakref.finalize`](https://docs.python.org/3.5/library/weakref.html#weakref.finalize)? It's a better solution than `atexit` because there is no need to unregister the clean up. PR has been updated with that approach.

micbou:
Rebased. The builds are now green.

micbou:
@davidhalter Is something blocking the PR? Is it because it conflicts with the typeshed branch? I don't mind reopening the PR against that branch if you prefer.

davidhalter:
Sorry, I was a bit stressed with a hundred different things. I forgot about this PR. I'm looking into it ASAP.

jedi/evaluate/compiled/subprocess/__init__.py
```diff
@@ -203,18 +222,7 @@ def get_sys_path(self):
...
-        except (AttributeError, TypeError):
-            # If the Python process is terminating, it will remove some modules
-            # earlier than others and in general it's unclear how to deal with
-            # that so we just ignore the exceptions here.
-            pass
-
-    def __del__(self):
-        if not self.is_crashed:
-            self._kill()
+        _cleanup_process(self._process, self._stderr_thread)
```
davidhalter: Why not `self._cleanup_callable = weakref.finalize(self, _cleanup_process, process, t)` above and then here: `self._cleanup_callable()`?

AFAIK this can lead to calling `_cleanup_process` twice for the same thread/subprocess.

micbou:
No worries. We are all doing this on our free time.

I noticed that the tests no longer pass on AppVeyor (on Travis as well but not for the same reasons and with less failures). I'd like to fix them before continuing to work on this PR since we can't be sure that all warnings are fixed if some tests are failing.

> PS: Don't worry about the merge conflicts, I'll take care of those. It's the least I can do.

I already fixed them on my local repository.

jedi/evaluate/compiled/subprocess/__init__.py
micbou: I think you are right. This could explain why I had to catch `OSError` exceptions in `_cleanup_process`. I'll change it as you suggested.

davidhalter:
I merged the test issues. Can you rebase again? I will then merge. Not a git rebase pro here, would it have been better if I merged the test improvements?

micbou:
Done.

> Not a git rebase pro here, would it have been better if I merged the test improvements?

No, I was going to rebase this PR anyway.

davidhalter:
Thanks a lot for this!

PR #1296, opened 2019-03-01
Improve test_import_completion_docstring robustness

micbou:
The `import_completion_docstring` test fails if there are other modules than `abc` whose name starts with `ab` (e.g. [`absl`](https://github.com/abseil/abseil-py)). This is fixed by filtering out all completions with a name different than `abc`.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1296)


davidhalter:
True, thanks!

PR #1297, opened 2019-03-09
fix add_bracket_after_function had no effect

Brow71189:
The type of a completion was compared to "Function" instead of "function" when the add_bracket_after_function flag was checked. This caused add_bracket_after_function to have no effect.

### DIFF
--- a/test/test_api/test_settings.py
+++ b/test/test_api/test_settings.py
@@ -23,8 +23,9 @@ def test_add_dynamic_mods(Script):
     assert result[0].description == 'class int'
 
 
-def test_add_bracket_after_function(Script):
-    api.settings.add_bracket_after_function = True
+def test_add_bracket_after_function(monkeypatch, Script):
+    settings = api.settings
+    monkeypatch.setattr(settings, 'add_bracket_after_function', True)
     script = Script('''\
 def foo():
     pass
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks for your contribution!

Please add a test that failed with the old code and I'm happy to merge :)

Brow71189:
I added the test (and confirmed that it actually failed with the old code).
I wasn't exactly sure where to put it, but "test_settings.py" seemed a good place. I'm also happy to change that if you want it somewhere else.

test/test_api/test_settings.py
```diff
@@ -21,3 +21,13 @@ def test_add_dynamic_mods(Script):
     result = script.goto_definitions()
     assert len(result) == 1
     assert result[0].description == 'class int'
+
+
+def test_add_bracket_after_function(Script):
+    api.settings.add_bracket_after_function = True
```
davidhalter: The only problem with this is that you set it globally for all tests that run after it. Can you please use the `monkeypatch` fixture? :) Thanks!

Brow71189: that should do it :)

davidhalter:
Thanks a lot for your contribution :)

PR #1300, opened 2019-03-16
Fix transform_path_to_dotted tests on Windows

micbou:
The  `transform_path_to_dotted` tests fail on Windows because paths with forward slashes are used in the tests while the function operates on paths using `os.path.sep` which is a backslash on Windows. This function should always receive normalized absolute paths so the proper fix is to normalize them in the tests as well.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1300)


micbou:
Forgot the test in the docstring.

davidhalter:
Thanks!

Issue #1273, opened 2019-01-26
Jedi crashes on completion with a KeyError on SequenceLiteralContext.mapping


adefossez:
When trying to complete the following file, called `source.py`:

```python
import other
import numpy as np
spec = np.zeros((12, 2))
result = other.power(spec)
result.
```

where `other.py` is defined as

```python
def power(spectrogram):
    return spectrogram[..., 0]**2 + spectrogram[..., 1]**2
```

using the following code:
```python
import jedi

source = open("source.py", "r").read()
line = 5
col = 7
filename = "source.py"
environment = None

script = jedi.Script(source, line, col, filename, environment)
script.completions()
```

When running this code against master, I get a `KeyError` on `SequenceLiteralContext.mapping`. It looks like the key is the token for `spectrogram` and not one of `[`, `(`, `{` like this code would expect. It seems that some code in `syntax_tree.py` wrongly considers that this expression is a list, dict or tuple literal. I included the full traceback hereafter.
Let me know if you need anymore information.
Cheers

Full traceback:
```
Traceback (most recent call last):
  File "bug.py", line 10, in <module>
    script.completions()
  File "/Users/defossez/tmp/jedi/jedi/jedi/api/__init__.py", line 179, in completions
    completions = completion.completions()
  File "/Users/defossez/tmp/jedi/jedi/jedi/api/completion.py", line 98, in completions
    completion_names = self._get_context_completions()
  File "/Users/defossez/tmp/jedi/jedi/jedi/api/completion.py", line 198, in _get_context_completions
    completion_names += self._trailer_completions(dot.get_previous_leaf())
  File "/Users/defossez/tmp/jedi/jedi/jedi/api/completion.py", line 234, in _trailer_completions
    contexts = evaluate_call_of_leaf(evaluation_context, previous_leaf)
  File "/Users/defossez/tmp/jedi/jedi/jedi/evaluate/helpers.py", line 77, in evaluate_call_of_leaf
    return context.eval_node(leaf)
  File "/Users/defossez/tmp/jedi/jedi/jedi/evaluate/base_context.py", line 145, in eval_node
    return self.evaluator.eval_element(self, node)
  File "/Users/defossez/tmp/jedi/jedi/jedi/evaluate/__init__.py", line 200, in eval_element
    return self._eval_element_if_evaluated(context, element)
  File "/Users/defossez/tmp/jedi/jedi/jedi/evaluate/__init__.py", line 212, in _eval_element_if_evaluated
[...]

PR #1307, opened 2019-04-04
Fix keyerror

adefossez:
tentative bug fix for issue #1273 : KeyError on SequenceLiteralContext.mapping

Added extra test in `test/completion/basic.py`, test procedure:

```
git checkout master && py.test test/test_integration.py -T basic.py
# result is 1 failed, 57 passed, 2 skipped, 1 warnings in 1.11 seconds
git checkout fix_keyerror && py.test test/test_integration.py -T basic.py
# result is 58 passed, 2 skipped, 1 warnings in 0.99 seconds
```

I try to run the full test suits but I get quite a few failures even on master. Number of failures is 83 on master counting the new test case, 82 on fix_keyerror branch when running `py.test test/test_integration.py`.

### DIFF
--- a/jedi/evaluate/syntax_tree.py
+++ b/jedi/evaluate/syntax_tree.py
@@ -225,7 +225,7 @@ def eval_atom(context, atom):
             right = eval_atom(context, string)
             context_set = _eval_comparison(context.evaluator, context, context_set, u'+', right)
         return context_set
-    else:
+    elif atom.type == 'atom':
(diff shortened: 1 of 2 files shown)


davidhalter:
I played with this a bit and realized that it's already different on the typeshed branch. Can you try there if it works?

adefossez:
Closing pull request as issue is fixed in typeshed branch.

PR #1308, opened 2019-04-05
Resolve path in get_cached_default_environment() to prevent unnecessary cache busting

nikhilmitrax:
Resolve path in get_cached_default_environment() in api/environment.py to prevent unnecessary cache busting when using pipenv.

This seems to help speed up JEDI (atleast when using it with `pyls`)

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -234,7 +234,12 @@ def _try_get_same_env():
 def get_cached_default_environment():
     var = os.environ.get('VIRTUAL_ENV')
     environment = _get_cached_default_environment()
-    if var and var != environment.path:
+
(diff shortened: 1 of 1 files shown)


davidhalter:
I guess this makes sense, because we do it everywhere else.

nikhilmitrax:
Thanks!

Also, thank you for this amazing project!

I was wondering if this should be ported to a hotfix tag or not, since it's seems to make Jedi orders of magnitude faster than it was. (at least for pipenv users).

I have reverted to using the `master` branch for Jedi, but maybe it might be better to just have a new release with the fix?

rmhsilva:
Thanks, a release with this would be really appreciated!

Issue #1015, opened 2018-01-04
Support completion for NewType "instances"


mwchase:
The `typing` module has a lightweight "NewType" function. From a PEP 484 perspective, the result of
```
MyType = NewType('MyType', int)
```
is a subclass of int, that has a constructor like
```
MyType(5) # typechecks as MyType, is an int at runtime
```
I believe the text editors I've checked have recent versions of Jedi, and they seem to be operating on the runtime behavior of "NewTypes":
```
MyType(5). # has int methods in completion dropdown

def test_func(int_arg: int, nt_arg: MyType):
    int_arg. # int methods in dropdown
    nt_arg. # no int methods in dropdown
```
I'd like variables annotated with a NewType value to have completion behavior as if they were just the underlying type (which is stored in the "type"'s `__supertype__` attribute, if that's relevant/helpful).

I would have put together a pull request, but I don't understand jedi's code well enough.

davidhalter:
If anybody wants to help with this, most of the details are hidden in evaluate/pep0484.py and evaluate/jedi_typing.py.

The latter is statically analyzed by jedi to generate types. A very simple implementation would be something like

```
def NewType(string, type_):
    return type_
```

However that would just work with int/str and a few others and is not generally the right way to do this. If you actually want to do this the right way you have to play with contexts.

mwchase:
I'm messing around with this, and I'm trying to get a sense of what's consistent with Jedi's existing behavior, and how I want to do things.

Looking at some of the tests, it looks like typing information "wins" over even trivial analysis, so we'd want something like
```
ANewType('ANewType', int)
y = ANewType(ANYTHING AT ALL)
#? ANewType()
y
#? ["to_bytes"]
y.t
```
And that the NewTypeContext would forward much of the documentation, as well as instance resolution, to the underlying ClassContext, while possibly including the name in it, somewhere.

Is that the right way of thinking about and approaching this?

davidhalter:
I'm pretty sure you're right. Generally speaking we want a new `MyType` Context that gives you 

I have just reread some parts of PEP 484 that are related to what you want to achieve. It looks like your approach sounds good.

> And that the NewTypeContext would forward much of the documentation, as well as instance resolution, to the underlying ClassContext, while possibly including the name in it, somewhere.

That sounds pretty good. You could even think about adding something like `__getattr__` or `__getattribute__` on the context and redirect some calls. I'm not sure if I have to create something like a proxy object anyway in the future.

[...]

PR #1312, opened 2019-04-10
NewType Context

mwchase:
Changes for https://github.com/davidhalter/jedi/issues/1015

I've got the tests I added last year, and I basically just took the suggestion from https://github.com/davidhalter/jedi/issues/1015#issuecomment-480261678 and changed things around until I thought it made sense in context.

I'm somewhat confused by the test behavior: they all passed before I made a PR to pull in the last few weeks, and they passed against the PR, but then some of the new code failed in one configuration after the PR merged. Then the same environment passed when I added my name to AUTHORS.txt, and made no other changes.

This probably isn't ready to merge, since the new tests are flaky, but it's good to have concrete code to discuss


test/completion/pep0484_typing.py
```diff
@@ -256,6 +256,20 @@ def testdict(x):
...
+y
+
+def testnewtype(y):
+    """
+    :type y: WrappingType
+    """
+    #? WrappingType()
+    y
+    #? ["upper"]
+    y.u
```
davidhalter: Please also add a test where `WrappingType2 = typing.NewType()`. This is a potential issue in your approach.

jedi/evaluate/gradual/typing.py
```diff
@@ -448,6 +450,27 @@ def py__call__(self, func_context_set):
...
 
 
+class NewTypeFunction(_BaseTypingContext):
+    def py__call__(self, arguments):
+        return ContextSet(
+            NewType(
+                self.evaluator,  # MWC: Not sure what I'm doing here
+                contextualized_node.context,
+                contextualized_node.node,
+                contextualized_node.infer(),  # MWC: Not sure what I'm doing here
```
davidhalter: The evaluator part is ok, all contexts have it. The second one, might work for your test case (not even sure), but not all test cases. The contextualized_node here is an arglist node: https://github.com/davidhalter/parso/blob/master/parso/python/grammar37.txt#L124. This is not what you want to be inferred. What we want is probably:

```
...
        ordered_args = arguments.unpack()
        next(ordered_args, None)
        second_arg = next(ordered_args, None)
        if second_arg is None:
            return NO_CONTEXTS
        return ContextSet(
            NewType(
                self.evaluator,
                contextualized_node.context,
                contextualized_node.node,
                second_arg.infer(),
            ) for contextualized_node in arguments.get_calling_nodes())
```

test/completion/pep0484_typing.py
mwchase: Should the desired behavior for `WrappingType2` be "arbitrary subclass of `object`"? That's what made sense to me, since it has to be a type, and with no arguments, it's not narrowed down.

davidhalter:
@mwchase Sorry, I guess I deleted the typeshed branch. Please create a new one against master. The typeshed branch was merged into master.

Issue #1309, opened 2019-04-07
Not showing top-level asynchronous function defintion with jedi.api.names


CXuesong:
I tried to use `jedi.api.names` to generate document outline for the following code
```python
# AsyncFunctions.py
import asyncio

def func() -> int:
    asyncio.run(func_async(0))

async def func_async(arg) -> int:
    await asyncio.sleep(1000)
    return 100
```
with the following code (flattened document outline)
```python
import io
import jedi

with io.open("AsyncFunctions.py", "r") as f:
    content = f.read()
    for n in jedi.api.names(content, f.name, all_scopes=True):
        print(n, n.parent(), n.is_definition())
```

It gives me the following output
```
<Definition module asyncio> <Definition module AsyncFunctions> True
<Definition def func> <Definition module AsyncFunctions> True
<Definition def func_async> <Definition module AsyncFunctions> True
<Definition param arg> <Definition def func_async> True
```
The result here is satisfactory.

However, if I remove `all_scopes=True` (use `False` by default), the output becomes
```
<Definition module asyncio> <Definition module AsyncFunctions> True
<Definition def func> <Definition module AsyncFunctions> True
```

The problem here, is `func_async`, regardless of it being a top-level function, is missing. Is this a bug or by design?

davidhalter:
It's probably a bug :)

CXuesong:
Okay. It seems that I need to make some modifications to [`get_module_names`](https://github.com/davidhalter/jedi/blob/8d0c4d3cec4973a2722652bf33f093ac3f1f385a/jedi/evaluate/helpers.py#L164), because async functions are actually functions wrapped in [`async_stmt`](https://github.com/davidhalter/parso/blob/1e25445176185f8ce27460cfa7162c897dd7527b/parso/python/grammar38.txt#L71).

I also noted there are still top-level function parameters and in-function local variables included in the result even if `all_scopes=False`, i.e. #1202, and curious about the current situation of that issue and PR #1203 . May I try to fix this as well?

davidhalter:
I just merged #1203, so feel free to work on it :)

davidhalter:
#1313 was merged and should solve this issue, thanks for helping out!

PR #1313, opened 2019-04-11
get_module_names should include top-level async functions when all_scopes=False.

CXuesong:
ref = #1309 

Skip extra layer of [`async_stmt`](https://github.com/davidhalter/parso/blob/1e25445176185f8ce27460cfa7162c897dd7527b/parso/python/grammar38.txt#L71), so async functions can be treated equally as sync functions.

### DIFF
--- a/test/test_api/test_defined_names.py
+++ b/test/test_api/test_defined_names.py
@@ -82,18 +82,34 @@ class TestDefinedNames(TestCase):
     def test_class_fields_with_all_scopes_false(self):
         definitions = self.check_defined_names("""
         from module import f
+        import asyncio
+
         g = f(f)
         class C:
             h = g
+            def __init__(self):
+                pass
+
+            async def __aenter__(self):
+                pass
 
         def foo(x=a):
            bar = x
            return bar
-        """, ['f', 'g', 'C', 'foo'])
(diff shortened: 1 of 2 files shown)


test/test_api/test_defined_names.py
```diff
@@ -82,18 +82,34 @@ def f(): pass
...
+            for i in range(duration//100):
+                await wait()
+            return duration//100*100
+        """, ['f', 'asyncio', 'g', 'C', 'foo', 'async_foo'])
+        C_subdefs = definitions[-3].defined_names()
+        foo_subdefs = definitions[-2].defined_names()
+        async_foo_subdefs = definitions[-1].defined_names()
+        self.assert_definition_names(C_subdefs, ['h', '__init__', '__aenter__'])
         self.assert_definition_names(foo_subdefs, ['x', 'bar'])
+        self.assert_definition_names(async_foo_subdefs, ['duration', 'wait', 'i'])
```
davidhalter: Thanks a lot for this contribution! Can you please either parametrize the test or just create a "new" one instead of modifying the old one, because the old one was already quite big :) Thanks!

davidhalter:
Thanks man!

Issue #1319, opened 2019-05-06
typeshed branch installation


kshpytsya:
When `typeshed` branch is installed not in "develop mode", e.g. by doing `pip install git+https://github.com/davidhalter/jedi@typeshed` or `git install /path/to/jedi-typeshed`, the contents of `jedi/third_party/typeshed` does not get installed.
Adding it to `package_data` in `setup.py` seems to fix it, but apparently some manual coding is needed to be able to add the files recursively.

davidhalter:
A `git submodule update --init` is currently needed. I have changed setup.py and MANIFEST.in though, so you should be able to use at least proper sdist/wheel packages in the future.

kshpytsya:
```
$ git clone --branch typeshed --recurse-submodules https://github.com/davidhalter/jedi
Cloning into 'jedi'...
remote: Enumerating objects: 434, done.
remote: Counting objects: 100% (434/434), done.
remote: Compressing objects: 100% (176/176), done.
remote: Total 41251 (delta 273), reused 354 (delta 241), pack-reused 40817
Receiving objects: 100% (41251/41251), 10.05 MiB | 1.40 MiB/s, done.
Resolving deltas: 100% (31128/31128), done.
Submodule 'jedi/third_party/typeshed' (https://github.com/davidhalter/typeshed.git) registered for path 'jedi/third_party/typeshed'
Cloning into '/tmp/665/jedi/jedi/third_party/typeshed'...
remote: Enumerating objects: 30, done.
remote: Counting objects: 100% (30/30), done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 20323 (delta 13), reused 22 (delta 12), pack-reused 20293
Receiving objects: 100% (20323/20323), 4.33 MiB | 1.21 MiB/s, done.
Resolving deltas: 100% (13801/13801), done.
error: Server does not allow request for unadvertised object 766ee7dd8cc7c8c9c28482abed58357124b75ee6
Fetched in submodule path 'jedi/third_party/typeshed', but it did not contain 766ee7dd8cc7c8c9c28482abed58357124b75ee6. Direct fetching of that commit failed.
```

davidhalter:
hmm that's strange. Can you try again. I pushed the branch again (some commits might not have been included)?!

kshpytsya:
Still the same. Tested on two machines in different locations.

davidhalter:
[...]

PR #1320, opened 2019-05-07
Fix installation of typeshed files

kshpytsya:
Issue #1319

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,26 @@
 #!/usr/bin/env python
 
+import fnmatch
+import os
 from setuptools import setup, find_packages
 
 import ast
+
+
+def rglob(top, pat, strip=0):
+    matches = []
+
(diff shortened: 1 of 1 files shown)


davidhalter:
Do we have to add `recursive-include jedi/third_party *` to `MANIFEST.in` as well?

davidhalter:
or just `recursive-include * *.pyi` and something like this:

```
    package_data = {
        # If any package contains *.pyi files, include them:
        '': ['*.pyi'],
    },
```

? Taken from https://stackoverflow.com/a/23936405/552671

I'm really not a packaging expert, it really just confuses me - even after all those years.

davidhalter:
I feel like the current version of setup.py works: https://github.com/davidhalter/jedi/blob/master/setup.py#L47

I have also added a few things to MANIFEST.in. Let me know if you disagree. Thanks for looking into it!

Issue #1257, opened 2018-11-26
Using python 3.7 fails on a Mac


DonJayamanne:
When using Python3.7 on a Mac, we (Python Extension for VS Code) throws the following error:

```
stderr jediProxy, Error (stderr)     _assert_safe(path, safe)
  File "/Users/donjayamanne/.vscode-insiders/extensions/pythonVSCode/pythonFiles/jedi/api/environment.py", line 352, in _assert_safe

stderr jediProxy, Error (stderr)     "The python binary is potentially unsafe.")
jedi.api.environment.InvalidPythonEnvironment: The python binary is potentially unsafe.
```

**Current solution:**
Change instantiation of Jedi from
* `jedi.api.environment.create_environment(sys.executable)` to 
* `jedi.api.environment.Environment(sys.executable)` 

Could be related to https://github.com/davidhalter/jedi/issues/1177

davidhalter:
Can you try he typeshed branch? I might enable travis testing for mac if this is really still an issue.

Dovermore:
I have the same issue. Using anaconda distribution of jedi.

Dovermore:
I find the cause, trying to fix it now.

Dovermore:
On my machine with conda distribution of python (both 3.6 and 3.7) will throw similar error. The reason being the TODO for `get_system_environment` won't pick up conda installed python.

Could it be the same problem here? I've done the fix for the issue on my machine. I will create a pull request for this.

It's my first time contributing. So I'm not familiar with the process of contributing. Please inform me if you can.

I haven't added the test yet. I'm not too sure how to do this. It will be great if you can point me to something I can get started with.

davidhalter:
Can you guys try with latest master? Quite a lot has changed (stubs were merged).

@Dovermore You can just create a Pull Request. No worries, we can then discuss how to proceed. Don't worry about tests, we can worry about that later as well.

davidhalter:
@Dovermore I see what you're trying to change in #1324. However, I fail to see what your exact issue is. What's the error messages you are seeing?

Why are the other Python executables on path not working?

Dovermore:
@davidhalter Hi, on my system it's only checking for executables in 
/bin/, /local/bin/ 
and don't check for executables in
/anaconda/bin/
or some other activated environments (by conda).

The error message is the same.
`stderr jediProxy, Error (stderr)     "The python binary is potentially unsafe.")
[...]

PR #1324, opened 2019-05-11
try making make get_system_environments to return list of envs

Dovermore:
try making make get_system_environments to return a list of envs. A fix of (at least) issues with conda distribution of pythons (for OSX). Possible fix for issue #1257 as well.

Please review the windows part of the function. I'm not sure if this will work correctly.

> It's my first time contributing. So I'm not familiar with the process of contributing. Please inform me if you can.
I haven't added the test yet. I'm not too sure how to do this. It will be great if you can point me to something I can get started with.

Thanks.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -51,5 +51,6 @@ Maksim Novikov (@m-novikov) <mnovikov.work@gmail.com>
 Tobias Rzepka (@TobiasRzepka)
 micbou (@micbou)
 Dima Gerasimov (@karlicoss) <karlicoss@gmail.com>
+Zhuoqun Calvin Huang (@Dovermoe) <calvinyeye0627@gmail.com>
 
 Note: (@user) means a github user name.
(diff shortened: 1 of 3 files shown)


davidhalter:
@Dovermore Thanks for your contribution!

The code is good, thanks for that. It does what it should and apart from the `a` flag that should just be a boolean, there is no need for changes.

I'm still going to close this for two reasons:

1. I don't see a lot of people making use of this - I feel it's not what people generally need and there's ways to work around it by just using create_environment
2. It's a maintenance burden to maintain `which` by myself instead of the Python community.

And I see you were a bit faster than me :)

I still hope you don't give up contributing. It's not that your contribution was bad or anything ;-)

Dovermore:
Ah, I left it as `"a"` because I was thinking if in future there is need it could be extended to include other flags (like fully simulating the terminal `which`)

I didn't quite understand all the logic detecting envs except for the linux part. That's why I didn't find an efficient workaround.

I did put my conda env as the first thing in the path. But it somehow still didn't pick that up. I will have a closer look at what I can do.

Thanks for the response!

PR #1326, opened 2019-05-20
Fix tests on CIs

micbou:
This PR fixes tests that are failing on Travis and AppVeyor because of the following points:
 - wrong path separator in the `pkg_zip_path` variable (which is used in the `correct_zip_package_behavior` tests) on Windows;
 - compiled modules using a different extension on Windows (`.pyd` instead of `.so`);
 - change in the representation of the `AbstractNameDefinition` class, breaking a few docstring tests;
 - version 2.0 of Sphinx (which is a dependency of numpydoc) dropped Python 2 support.

There are still tests failing locally on my machine with Python 2 but I didn't bother trying to fix them since it's Python 2 and they are passing on AppVeyor.

---
This change is [Reviewable](https://reviewable.io/reviews/davidhalter/jedi/1326)

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -102,6 +102,7 @@ class BaseDefinition(object):
         to Jedi, :meth:`jedi.Script.goto_definitions` should return a list of
         definition for ``sys``, ``f``, ``C`` and ``x``.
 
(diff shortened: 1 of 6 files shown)


davidhalter:
Thanks a lot!

Issue #1321, opened 2019-05-07
Crashes when parsing super calls for classes without bases


JCavallo:
I'm having the following crash:

```
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/api/__init__.py", line 178, in completi
ons
    completions = completion.completions()
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/api/completion.py", line 98, in complet
ions
    completion_names = self._get_context_completions()
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/api/completion.py", line 198, in _get_c
ontext_completions
    completion_names += self._trailer_completions(dot.get_previous_leaf())
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/api/completion.py", line 234, in _trail
er_completions
    contexts = evaluate_call_of_leaf(evaluation_context, previous_leaf)
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/evaluate/helpers.py", line 105, in eval
uate_call_of_leaf
    values = eval_trailer(context, values, trailer)
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/evaluate/syntax_tree.py", line 184, in
eval_trailer
    return base_contexts.execute(args)
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/common/context.py", line 65, in mapper
    for context in self._set
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/common/context.py", line 34, in from_se
ts
    sets = list(sets)
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/common/context.py", line 65, in <genexp
r>
    for context in self._set
  File "/home/giovanni/.cache/dein/.cache/.vimrc/.dein/rplugin/python3/deoplete/vendored/jedi/jedi/debug.py", line 81, in wrapper
[...]

PR #1342, opened 2019-06-12
Ignore super calls when super class cannot be inferred

JCavallo:


JCavallo:
Tentative fix for #1321

jedi/plugins/stdlib.py
```diff
@@ -243,6 +243,9 @@ def builtins_super(types, objects, context):
     if isinstance(context, FunctionExecutionContext):
         if isinstance(context.var_args, InstanceArguments):
             su = context.var_args.instance.py__class__().py__bases__()
+            # If super class cannot be infered
+            if len(su) == 0:
+                return NO_CONTEXTS
```
davidhalter: Good job. I think this generally solves this issue.

However now that I thought a bit about this, I feel like it would be better to change `ClassContext.py__bases__` like this:

```
--- a/jedi/evaluate/context/klass.py
+++ b/jedi/evaluate/context/klass.py
@@ -233,14 +233,16 @@ class ClassContext(use_metaclass(CachedMetaClass, ClassMixin, FunctionAndClassBa
         if arglist:
             from jedi.evaluate import arguments
             args = arguments.TreeArguments(self.evaluator, self.parent_context, arglist)
-            return [value for key, value in args.unpack() if key is None]
-        else:
-            if self.py__name__() == 'object' \
-                    and self.parent_context == self.evaluator.builtins_module:
-                return []
-            return [LazyKnownContexts(
-                self.evaluator.builtins_module.py__getattribute__('object')
-            )]
+            lst = [value for key, value in args.unpack() if key is None]
+            if lst:
+                return lst
+
+        if self.py__name__() == 'object' \
+                and self.parent_context == self.evaluator.builtins_module:
+            return []
+        return [LazyKnownContexts(
+            self.evaluator.builtins_module.py__getattribute__('object')
+        )]
 
     def py__getitem__(self, index_context_set, contextualized_node):
         from jedi.evaluate.gradual.typing import LazyGenericClass
```

I feel like this is better, because if you don't pass a parent class to meta classes, `object` is chosen. So `object` in Python is always somewhere in the `mro()`, there's no way around it (except for `object` itself).

JCavallo: I agree, it feels better this way. There is no way `builtins_super` would be called with `object`  as `context.var_args.instance.py__class__()` ?

Because in that case `py__bases__` will still return an empty list.

davidhalter:
Thanks!

PR #1344, opened 2019-06-13
New type context

mwchase:
Redoing the PR. This has changes from some comments on the original PR. The change corresponding to the second comment is likely incorrect.

I wanted to address all of the comments first, but I'm currently trying to work out how to make Travis handle new builds, and I thought it was possible that a new PR would help.


mwchase:
I haven't yet addressed "Make this a str(), it makes more sense, I feel like." because I'm not sure whether I agree, but I covered the other issues.

davidhalter:
Could you please fix the conflict and change the thing I mentioned below? I will merge at that point. Thanks a lot for your work!

> I haven't yet addressed "Make this a str(), it makes more sense, I feel like." because I'm not sure whether I agree, but I covered the other issues.

The thing is: This is a test and it's important to reflect what's actually happening - it is in fact a string we're dealing with there, it's just that `WrappingType` is pointing towards it there. So if you're working with Jedi you will see a string there regardless of this change, it's just that by reading it now you will think that it's not a string (which it is).

davidhalter:
My bad, it's actually `str`, not `str()`.

mwchase:
I think I'm misunderstanding something about the other tests, then. Looking elsewhere in the file, I'm seeing stuff like `#? int()` to check that a member of a sequence of `int`s is an instance of `int`. If the test is making a fake type that subclasses `str`, then an instance of the type is an instance of `str`. So, it looks to me like the following two things are true:

We want to be checking for an instance of `str`, and we check for an instance of `str` by using `str()`.

davidhalter:
You are right, I guess. It's probably another issue in the code, it should maybe be:

```
    def py__call__(self, arguments):
        return self._type_context_set.execute_annotation()
```

davidhalter:
Hmm the appveyor tests are failing. It's probably related to pytest 5.0.0 being released. Consider everything fine with your branch.

davidhalter:
Thanks!

PR #1349, opened 2019-06-21
Update parso requirement

### LINKED ISSUE #1348: Jedi's install_requires claims it works with old parso, but this doesn't appear to be true
According to jedi's `install_requires`, all it needs is `parso >= 0.3.0`.

But judging from https://github.com/python-trio/trio/pull/1121, it looks like actually even parse 0.4.0 isn't good enough – we get:

```
>   from parso.python.tree import ExprStmt, SyncCompFor
E   ImportError: cannot import name 'SyncCompFor' from 'parso.python.tree' (/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/parso/python/tree.py)
```

So I guess the `install_requires` should be updated to say `parse >= 0.5.0`.

njsmith:
Fixes #1348


davidhalter:
True, thanks!

davidhalter:
My comment in #1348 still applies, do I have to create a new release?

PR #1352, opened 2019-06-24
Fix for EmptyCompiledName

### LINKED ISSUE #1351: Exception on completion from compiled object using JPype
This is a followup to #1347 

Although completion is working for some cases, others are generating a RuntimeError in jedi.  

Example:

```
s=jpype.java.lang.String("foo")
src='s.substring(1).con'
script = jedi.Interpreter(src,[locals()])
compl = [i.name for i in script.completions()]
print(compl)
```
Gives `['concat', 'contains', 'contentEquals']`

Remove the "con" and we get...
```
s=jpype.java.lang.String("foo")
src='s.substring(1).'
script = jedi.Interpreter(src,[locals()])
compl = [i.name for i in script.completions()]
print(compl)
```

Gives
```
  """)
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/api/completion.py", line 44, in filter_names
    k = (new.name, new.complete)  # key
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/api/classes.py", line 439, in complete
    return self._complete(True)
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/api/classes.py", line 408, in _complete
    if self._name.api_type == 'param' and self._stack is not None:
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/evaluate/compiled/context.py", line 284, in api_type
    return next(iter(self.infer())).api_type
StopIteration

The above exception was the direct cause of the following exception:

  File "testJedi.py", line 18, in <module>
    compl = [i.name for i in script.completions()]
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/api/__init__.py", line 216, in completions
    return completion.completions()
  File "/home/osboxes/python3-env/lib/python3.7/site-packages/jedi-0.14.0-py3.7.egg/jedi/api/completion.py", line 104, in completions
    return sorted(completions, key=lambda x: (x.name.startswith('__'),
RuntimeError: generator raised StopIteration
```

Thrameos:
Fixes #1351

### DIFF
--- a/jedi/evaluate/compiled/context.py
+++ b/jedi/evaluate/compiled/context.py
@@ -281,7 +281,11 @@ class CompiledName(AbstractNameDefinition):
 
     @property
     def api_type(self):
-        return next(iter(self.infer())).api_type
+        api = self.infer()
+        # If we can't find the type, assume it is an instance variable
+        if not api:
+            return "instance"
+        return next(iter(api)).api_type
 
     @underscore_memoization
     def infer(self):


Thrameos:
I also considered an alternative fix as

```
    def api_type(self):
        try:
             return next(iter(self.infer())).api_type
        except:
             # If something goes wrong assume it is an "instance"
             return "instance"
```

But this seems a bit over general.

davidhalter:
Thanks!

BTW: I would never merge your "alternative fix". Working with Python for a few years taught me that this is a really really bad idea. It's easily the worst pattern that people in Python use. `except Exception` would have been a bit better (you never want to catch stuff like MemoryError in such a place), but still a very bad idea.

The reason for this is that it masks exceptions that are usually caused by the programmer. Once you have a few places like this, it gets really really hard to find the cause of any issue, because it might just be hidden behind something like this. `except Exception` should IMO only be used if it's reraised (or if you're writing a script that is not going to be used in the future).

Thrameos:
Yep.  Fully agree which is why I went with the limited fix.  The same thing happens in Java and it is a mess to track down.  Thank you very much for all your help.

ccordoba12:
@davidhalter, could you release 0.14.1 with this fix? Right now the [python-language-server](https://github.com/palantir/python-language-server), which I help to maintain and is used by VSCode, Spyder and several other editors and IDEs, doesn't work with Jedi 0.14 due to this.

Thanks!

davidhalter:
@ccordoba12 Next release is coming soon. I'm just trying to fix a few bugs that are left in the current release. It will probably take two to four weeks (I'm on vacation for that whole time, haha).

ccordoba12:
Ok, no problem :+1: Thanks for letting us know about it.

For now we're requiring a Jedi version less than 0.14 to manage things until 0.14.1 is out.

PR #1356, opened 2019-07-02
Harmonizing function signature in interpreter

linupi:
Using jedi in a REPL (based on ptpython) I saw that the function signature of functions in some cases (e.g. input entered through the REPL) did not contain default values of key word arguments. I saw that things around the signature have been moving since the last release so I tried to provide a test highlighting the problem.


davidhalter:
Thanks for this. I changed your tests slightly, but they should still carry the same meaning.

Issue #1354, opened 2019-06-26
Traceback for tab completion on windows 10 in REPL and sys module 


jmfrank63:
jedi==0.14.0
parso==0.5.0
pyreadline==2.1

python version 3.7.3

On windows 10 REPL completion works if you resolve $(python -m jedi repl) and put the resolved replstartup.py in your PYTHONPATH. e.g. C:\Users\jmfra\scoop\apps\python\3.7.3\Lib\site-packages\jedi\api\replstartup.py

The following works fine:

import os

os.<tab>

However with the sys module jedi crashes:

```
 15:46:11  ~  $ python
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
REPL completion using Jedi 0.14.0
>>> import sys
>>> sys.ERROR:root:REPL Completion error:
Traceback (most recent call last):
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\utils.py", line 97, in complete
    completions = interpreter.completions()
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\api\__init__.py", line 216, in completions
    return completion.completions()
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\api\completion.py", line 99, in completions
    completion_names = self._get_context_completions()
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\api\completion.py", line 204, in _get_context_completions
    completion_names += self._trailer_completions(dot.get_previous_leaf())
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\api\completion.py", line 249, in _trailer_completions
    python_contexts = convert_contexts(contexts)
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\evaluate\gradual\conversion.py", line 159, in convert_contexts
    for stub_context in contexts
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\common\context.py", line 32, in from_sets
    for set_ in sets:
  File "C:\Users\jmfra\scoop\apps\python\current\lib\site-packages\jedi\evaluate\gradual\conversion.py", line 159, in <genexpr>
    for stub_context in contexts
[...]

PR #1370, opened 2019-07-14
Fix for failing assertion on native modules Issue #1354

jmfrank63:
Not finished yet. Got a failing test
```
self = <test.test_utils.TestSetupReadline testMethod=test_import>

    def test_import(self):
        s = 'from os.path import a'
        assert set(self.completions(s)) == {s + 'ltsep', s + 'bspath'}
        assert self.completions('import keyword') == ['import keyword']
    
        import os
        s = 'from os import '
        goal = {s + el for el in dir(os)}
        # There are minor differences, e.g. the dir doesn't include deleted
        # items as well as items that are not only available on linux.
        difference = set(self.completions(s)).symmetric_difference(goal)
        difference = {x for x in difference if not x.startswith('from os import _')}
        # There are quite a few differences, because both Windows and Linux
        # (posix and nt) libraries are included.
>       assert len(difference) < 38
E       AssertionError: assert 72 < 38
E        +  where 72 = len({'from os import EX_NOTFOUND', 'from os import GRND_NONBLOCK', 'from os import GRND_RANDOM', 'from os import O_BINARY', 'from os import O_DIRECT', 'from os import O_LARGEFILE', ...})

test/test_utils.py:83: AssertionError
============================================== 1 failed, 2393 passed, 38 skipped, 5 xfailed in 129.70 seconds ===============================================
ERROR: InvocationError for command /Users/johannes/Documents/Projects/Python/jedi/.tox/py37/bin/pytest (exited with code 1)
```

and there is not yet a failing test for the fix. I have added a stub but I still struggle with writing tests.


davidhalter:
By the way, your test only seems to be failing on your local machine. So I would think that this is related to your Python setup somehow. It would be interesting, but that is a whole other issue.

About the failing test, if you need help let me know.

jmfrank63:
Hi David,

testing is definitely my weak point, but I am working on it. Here after
thinking about it, I think I am confused because the failure is actually an
assertion. If I take out the assertion the code works.

So my intuition somehow tells me to move the assertion into the test suite
and remove it entirely from the code. But jedi is a complicated piece of
code and I am not having the overview of side effects.

Just to elaborate my thoughts I did use an assertion in code once but
caught its failure to take a different path (I was sorting tiles). But here
no decision is taken so if you are fine I would remove the assertion
entirely from the code and do the testing in the test suite.

Thanks

Johannes

davidhalter:
I feel like you already had a failing test in your initial issue, just only in the terminal, which was something like this:

```
def test_x():
    import sys
    assert Interpreter('sys.', [locals()]).completions()
```

That's all I meant by a failing test :)
```

**Comment by jmfrank63** [comment]
All right, the checks all pass, the test tests all core modules which takes a couple of seconds, let me know if you think this is overkill.

File: test/test_api/test_interpreter.py [hunk 1]
```diff
@@ -440,5 +440,10 @@ def syslogs_to_df():
     # Apparently the function starts on the line where the decorator starts.
     assert c.line == syslogs_to_df.__wrapped__.__code__.co_firstlineno + 1
 
-def test_infer_from_stub():
-    assert True
+def test_core_module_completes():
+    import sys
+    import importlib
+    for module_name in sys.builtin_module_names:
```
davidhalter: While I endorse your efforts, I feel like this is not necessary. This just takes longer than we need. Either test just `sys` or test a couple of modules like `time`, `sys` and maybe `gc`. You could do that easily with parametrize:

```
@pytest.mark.parametrize('module_name', ['sys', 'time'])
def test_core_module_completes(module_name):
```

It's a nice way to create multiple tests. BTW, even if we were to test all builtin modules, you should probably do it like this: `pytest.mark.parametrize('module_name', sys.builtin_module_names)`.

jedi/evaluate/gradual/conversion.py
```diff
@@ -40,7 +40,10 @@ def _stub_to_python_context_set(stub_context, ignore_compiled=False):
 
 
 def _infer_from_stub(stub_module, qualified_names, ignore_compiled):
-    assert isinstance(stub_module, StubModuleContext), stub_module
+    from jedi import evaluate
+    assert isinstance(stub_module, StubModuleContext) or \
+           isinstance(stub_module, evaluate.compiled.mixed.MixedObject) , stub_module
```
davidhalter: Can you just import `from jedi.evaluate.compiled.mixed import MixedObject`? I like it better that way :)

Also please remove the comment below, I will merge once that's done.

jmfrank63: Me too, however we are back to the circular import, because mixed imports `to_stub` from `conversion.py`.
I saw you plan to merge `to_stub` anyway so we could defer this to a new issue and fix the import this way.

davidhalter: Why would it be a circular import? As long as it's in the function it's no problem, it's just if you move it out of the function that it's going to be circular.

davidhalter:
Thanks for this! And sorry for the confusion around the commits.

jmfrank63:
Hi David,

no worries, thanks for letting me having part in your awesome library. I
gained more experience and learned a lot again.

If you have another issue you think I can take on let me know. Maybe
merging the to_stub as listed in the TODO comment, or something you think
is appropriate for me.

Thanks

Johannes

davidhalter:
If I could suggest one thing (that is probably not too hard): Take the work in https://github.com/davidhalter/jedi/pull/1260 and finish it. There's a few comments in there that the author in there did never respond to. Since a lot of time has passed, you can just create a new pull request to finish it.

PR #1371, opened 2019-07-15
Signature of functools partial

linupi:
using jedi through a ptpython based REPL (use case as in #1356) we often face the problem that signatures of partials are not displayed. I dug a bit into it and tried to provide at least the relevant tests as I didn't come up with a proper solution myself. In the tests I try to reproduce the behavior of inspect.signature on functions generated by `functools.partial` .

### DIFF
--- a/test/test_api/test_call_signatures.py
+++ b/test/test_api/test_call_signatures.py
@@ -544,6 +544,50 @@ def test_arg_defaults(Script, environment, code):
         assert signatures[0].params[1].description == 'param arg1=1'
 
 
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks! I will probably look into signatures quite soon and we could probably fix this and similar issues.

The problem is that currently no good way to work with signatures exists.

davidhalter:
I tested it in a slightly different way, so I didn't end up using your code. Thanks for creating it anyway. It made it a lot easier to reproduce the issue. If only other people also provided such good reproduction steps :)

PR #1380, opened 2019-08-12
Remove forgotten debug/print in filename completion.

Carreau:
This is in the latest 0.15, and when forwarding path completions to
jedi, print a lot of stuff on the screen.


davidhalter:
Thanks! Will release this with another small change in the next 1-2 days.

davidhalter:
Released in 0.15.1.

PR #1382, opened 2019-08-13
Use scandir on py3.5+ for less disk access on filename completion

### LINKED ISSUE #1381: consider scandir (instead of listdir) for filenames
Python 3.6+ has scandir (with [fallback on PyPI](https://pypi.org/project/scandir/) if not available) which should be significantly faster that listdir especially if you call is_dir or is_file.

Carreau:
On Python 3.5+, we can make use of scandir that not only list the
content of the directory as an iterator but caches some infomations (for
example, `is_dir()`; this avoid extra stats call to the underlying
filesytem and can be – according to pep 471 –  2x to 20 time faster
especially on NFS filesystem where stats call is expensive.

From a quick this is the only place where scandir would make sens, as
most other places only require the name.

Fixes #1381

### DIFF
--- a/jedi/_compatibility.py
+++ b/jedi/_compatibility.py
@@ -27,6 +27,27 @@ is_py35 = is_py3 and sys.version_info[1] >= 5
 py_version = int(str(sys.version_info[0]) + str(sys.version_info[1]))
 
 
+if is_py35:
+    """
+    A super-minimal shim around listdir that behave like
+    scandir for the information we need.
+    """
+    class _DirEntry:
+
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks!

davidhalter:
Let me know if you need a release soon. Otherwise it might take 1-2 months.

Carreau:
Nah, that's fine thanks. We'll see once we've switched IPython to jedi for
file completion all the time, but it might still take a couple of month.

PR #1384, opened 2019-08-13
Complete path after ~.

Carreau:
Note this is mostly to discuss as if I understood one of your message on
Twitter, this was not possible without fuzzy completion.

I tried with just this patch and that works great.

Note that unlike IPython that right now does :

    ~/<tab> -> /Full/Path/to/user/home

But with this patch this just complete things correctly without
expanding the tab. And I think not expanding the tab is actually better.

Anyway, open that to better understand the why you were waiting for
fuzzy completion.

### DIFF
--- a/jedi/api/file_name.py
+++ b/jedi/api/file_name.py
@@ -37,6 +37,8 @@ def file_name_completions(evaluator, module_context, start_leaf, string,
     like_name_length = len(os.path.basename(string) + like_name)
 
     addition = _get_string_additions(module_context, start_leaf)
+    if string.startswith('~'):
+        string = os.path.expanduser(string)
     if addition is None:
         return
     string = addition + string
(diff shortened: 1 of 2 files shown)


davidhalter:
You are right. This is just a bit different from IPython functionality. But I probably agree that expanding `~` is not the correct action, but it should be possible to complete it this way. So I'm all for this patch.

If you want you can add a test. Otherwise I will after merging (might take a few days, but it will happen).

Carreau:
> If you want you can add a test

Ok, I attempted to write a test; It picks the first file that does not start with `.`it finds in `~/` and check whether Script suggest that as a completion.

I did not create a tempdir as I don't want to mess with filesystem too close to `~`.

Not sure about windows behavior.

Let's see if this Passes on CI.

davidhalter:
It looks like there's still a minor problem on Windows: https://ci.appveyor.com/project/davidhalter/jedi/builds/26944908/job/ahyyh6fc87rnd07i

Not sure about if the backslash should be there or not.

davidhalter:
Finally got around to fixing Windows issue. Thanks again for this PR :)

PR #1394, opened 2019-09-10
Add API to find the virtual_env of single venv path

15cm:
Extract a `find_virtualenv` function to find single venv with given path. This function provides the API for _pyenv_ support in _python-language-server_: [Support pyenv virtualenv for LSP project roots ](https://github.com/15cm/python-language-server/commit/e3433c891dbce8131de005a3b9e4dba586bc8a33)

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -53,5 +53,6 @@ micbou (@micbou)
 Dima Gerasimov (@karlicoss) <karlicoss@gmail.com>
 Max Woerner Chase (@mwchase) <max.chase@gmail.com>
 Johannes Maria Frank (@jmfrank63) <jmfrank63@gmail.com>
+Sinkerine (@15cm) <i@15cm.net>
 
 Note: (@user) means a github user name.
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -251,6 +251,15 @@ def _get_cached_default_environment():
     return get_default_environment()
 
 
+def find_virtualenv(path, safe=True):
+    if os.path.isdir(path):
+        try:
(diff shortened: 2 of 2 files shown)


davidhalter:
Why not use `list(find_virtualenvs([path]))`?

I feel like if we want to publish your API we should stay consistent with `get_system_environment`, which raises `InvalidPythonEnvironment` if it doesn't work.

Do you really prefer the exception solution to the iterator?

15cm:
> Why not use `list(find_virtualenvs([path]))`?
> 
> I feel like if we want to publish your API we should stay consistent with `get_system_environment`, which raises `InvalidPythonEnvironment` if it doesn't work.
> 
> Do you really prefer the exception solution to the iterator?

Because `find_virtualenvs` finds venv in all directories under each path:
[paths] -> path -> [dirs] -> dir -> dir/bin/python
But I want a API that returns a venv for a given directory:
dir -> dir/bin/python

In fact I think it's more sensible for `find_virtualenvs` to do this:
[paths] -> path -> path/bin/python
But since this is not the case, I have to extract an API out of the function.

As for the exception, I just follow the logic of `find_virtualenvs`, where the exception is ignored.

davidhalter:
Ok. I understand your point and I also agree that it's not obvious. Why are you then not using `create_environment`? It looks like that would work for you.

15cm:
That's exactly what I want, thank you. BTW I will appreciate if you can make the comment of `find_virtualenvs` clearer.

davidhalter:
Feel free to suggest something better. I get that it is a problem, but I'm not the best writer :).

Issue #1395, opened 2019-09-10
Test failure in test_file_path_completions


Apteryks:
While working on packaging 0.15.1 for GNU Guix, I encountered the following repeatable test failure:

````
 =================================== FAILURES ===================================
____ test_file_path_completions[example.py-rb"../jedi/tes-None-expected12] _____

Script = functools.partial(<class 'jedi.api.Script'>, environment=<Environment: 3.7.0 in /gnu/store/b7fqhszxl02g6pfm3vw6b3cjz472qrly-python-3.7.0>)
file = 'example.py', code = 'rb"../jedi/tes', column = None, expected = ['t/']

    @pytest.mark.parametrize(
        'file, code, column, expected', [
            # General tests / relative paths
            (None, '"comp', None, ['ile', 'lex']),  # No files like comp
            (None, '"test', None, [s]),
            (None, '"test', 4, ['t' + s]),
            ('example.py', '"test%scomp' % s, None, ['letion' + s]),
            ('example.py', 'r"comp"', None, "A LOT"),
            ('example.py', 'r"tes"', None, "A LOT"),
            ('example.py', 'r"tes"', 5, ['t' + s]),
            ('example.py', 'r" tes"', 6, []),
            ('test%sexample.py' % se, 'r"tes"', 5, ['t' + s]),
            ('test%sexample.py' % se, 'r"test%scomp"' % s, 5, ['t' + s]),
            ('test%sexample.py' % se, 'r"test%scomp"' % s, 11, ['letion' + s]),
            ('test%sexample.py' % se, '"%s"' % join('test', 'completion', 'basi'), 21, ['c.py']),
            ('example.py', 'rb"' + join('..', 'jedi', 'tes'), None, ['t' + s]),
    
            # Absolute paths
            (None, '"' + join(root_dir, 'test', 'test_ca'), None, ['che.py"']),
            (None, '"%s"' % join(root_dir, 'test', 'test_ca'), len(root_dir) + 14, ['che.py']),
    
            # Longer quotes
            ('example.py', 'r"""test', None, [s]),
            ('example.py', 'r"""\ntest', None, []),
            ('example.py', 'u"""tes\n', (1, 7), ['t' + s]),
            ('example.py', '"""test%stest_cache.p"""' % s, 20, ['y']),
            ('example.py', '"""test%stest_cache.p"""' % s, 19, ['py"""']),
    
            # Adding
            ('example.py', '"test" + "%stest_cac' % se, None, ['he.py"']),
[...]

PR #1398, opened 2019-09-13
test: test_completion: Dynamically resolve current directory name.

Apteryks:
This fixes issue #1395 (see:
https://github.com/davidhalter/jedi/issues/1395).

* test/test_api/test_completion.py(current_dirname): New variable.
(test_file_path_completions): Use it.


davidhalter:
Makes sense, thanks!

PR #1406, opened 2019-09-23
find_virtualenvs: look at provided dir also

blueyed:
`find_virtualenvs($VIRTUAL_ENV)` should work, without using `dirname()` first.

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -283,7 +283,7 @@ def find_virtualenvs(paths=None, **kwargs):
                 continue
 
             directory = os.path.abspath(directory)
-            for path in os.listdir(directory):
+            for path in [directory] + os.listdir(directory):
                 path = os.path.join(directory, path)
                 if path in _used_paths:
                     # A path shouldn't be inferred twice.
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -283,7 +283,12 @@ def find_virtualenvs(paths=None, **kwargs):
                 continue
 
             directory = os.path.abspath(directory)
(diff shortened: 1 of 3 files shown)


test/test_api/test_environment.py
```diff
@@ -108,6 +108,13 @@ def test_scanning_venvs(venv_path):
                for venv in find_virtualenvs([parent_dir]))
 
 
+def test_find_virtualenvs_finds_via_VIRTUAL_ENV_root(venv_path):
+    """Test that is checks the directory itself, not just childs."""
+    assert "VIRTUAL_ENV" not in os.environ
+    found_envs = find_virtualenvs([venv_path])
+    assert [venv.path for venv in found_envs] == [venv_path]
```
blueyed: Note that this is very similar to the test above, which uses `dirname()`.

Good to keep both of them probably for now.

jedi/api/environment.py
```diff
@@ -283,7 +283,7 @@ def py27_comp(paths=None, safe=True):
                 continue
 
             directory = os.path.abspath(directory)
-            for path in os.listdir(directory):
+            for path in [directory] + os.listdir(directory):
                 path = os.path.join(directory, path)
                 if path in _used_paths:
                     # A path shouldn't be inferred twice.
```
blueyed: This could also handle only directories in the first place?!

davidhalter:
Before I comment about the code, have you seen `create_environment`? And would that not solve your use case?

davidhalter:
I feel like `create_environment` does the job. If you disagree let me know. I will reopen in that case.

PR #1421, opened 2019-10-10
Support anaconda system env

galaunay:
## Context

I am the current maintainer of [Elpy](https://github.com/jorgenschaefer/elpy).
We use jedi to provide completion, access to the documentation, etc... (all the nice things).

To avoid having to ask every user to install jedi systemwide, we run jedi in a dedicated virtual environment ('running virtualenv') and specify the environment to work on ('working environment') through the `environment` argument: 
```
env = jedi.create_environment("/path/to/working-environment")
script = jedi.Script(..., environment=env, ...)
```

It works very nicely when the working environment is a virtual one.
When the working environment is the system environment itself, we provide a path that looks like this: `env = jedi.create_environment("/usr")`, so that jedi can find the binaries in `/usr/bin/python`.

(Please let me know if we are not supposed to use jedi like this)

## Problem

On windows and with Anaconda (I don't know if it is happening with other OS/python distributions), the system python binaries are stored under `.../Continuum/anaconda3/python.exe`.
Using `env = jedi.create_environment(".../Continuum")` does not work, as jedi is only looking for python binaries under a `Scripts` folder.

## Proposed solution

This PR just modifies the `_get_executable_path()` function to check for python binaries under a `anaconda3` folder as well on Windows.

I can add a few tests if you think this PR is relevant.

If I managed to keep you interested until this point, I will take the opportunity to thank you for Jedi, it is a wonderful tool to work with !

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -361,7 +361,11 @@ def _get_executable_path(path, safe=True):
     """
 
     if os.name == 'nt':
-        python = os.path.join(path, 'Scripts', 'python.exe')
+        intermediate_folders = ['Scripts', 'anaconda3']
+        for folder in intermediate_folders:
+            python = os.path.join(path, folder, 'python.exe')
+            if os.path.exists(python):
+                break
     else:
         python = os.path.join(path, 'bin', 'python')
     if not os.path.exists(python):


galaunay:
Just realised we could create Jedi's environments by giving the path to the python binaries:
 `env = jedi.create_environment("../Continuum/anaconda3/python.exe")`.

Hope I did it quickly enough to avoid wasting your time.

davidhalter:
Feel free to create a PR that improves documentation about this :) I'm pretty sure it's just not documented well enough.

galaunay:
Well I just checked and it is indeed written:
``` python
def create_environment(path, safe=True):
    """
    Make it possible to manually create an Environment object by specifying a
    Virtualenv path or an executable path.
    """
```
I just need to learn to read docstrings until the end...

PR #1431, opened 2019-10-26
Fix annotation string generated from wrong object

Endilll:
For `inspect.Parameter` objects
`'clip': <Parameter "clip: vapoursynth.VideoNode">`
`'width': <Parameter "width: Union[int, NoneType] = None">`
created from compiled object, jedi unexpectedly outputted
`clip: <class 'inspect._empty'>`
`width: None=None`
in call signature. With this PR applied, output is expected
`clip: <class 'vapoursynth.VideoNode'>`
`width: typing.Union[int, NoneType]=None`


davidhalter:
Thanks for looking into this!

Issue #1437, opened 2019-11-08
Wrong completions when function parameters type comment is only one parameter with a dot


perrinjerome:
Type of function parameter is not evaluated correctly when there is only one parameter and this parameter contain a dot, like in this example:

```python
import logging
def test(x):
   # type: (logging.Logger) -> None
   x.inf
```

I understood that in this case the node from [`_split_comment_param_declaration`](https://github.com/davidhalter/jedi/blob/34f131e9b33d6895399f80ba7c1d9548e0937a6b/jedi/inference/gradual/annotation.py#L79) here
https://github.com/davidhalter/jedi/blob/34f131e9b33d6895399f80ba7c1d9548e0937a6b/jedi/inference/gradual/annotation.py#L88-L93
 has type `atom_expr` :

```
PythonNode(atom_expr, [<Name: logging@1,0>, PythonNode(trailer, [<Operator: .>, <Name: Logger@1,8>])])
```

so this condition is false:

https://github.com/davidhalter/jedi/blob/34f131e9b33d6895399f80ba7c1d9548e0937a6b/jedi/inference/gradual/annotation.py#L93-L95

and this branch is evaluated:

https://github.com/davidhalter/jedi/blob/34f131e9b33d6895399f80ba7c1d9548e0937a6b/jedi/inference/gradual/annotation.py#L103-L105

and we have `["logging"]` instead of the expected `["logging.Logger"]`

---

For full reference, here's the rest of my findings:

```python
from logging import Logger
def test(x):
   # type: (Logger) -> None
   x.inf
```

was completing properly, in this case `node.type` is `name`.

also this case with a trailing ,

```python
import logging
def test(x):
   # type: (logging.Logger,) -> None
   x.inf
```

because in this case it's a `testlist_start_expr`
```
PythonNode(testlist_star_expr, [PythonNode(atom_expr, [<Name: logging@1,0>, PythonNode(trailer, [<Operator: .>, <Name: Logger@1,8>])]), <Operator: ,>])
```

( I'm sending a pull request for this issue )

davidhalter:
BTW: If you want the issue to autoclose from a commit message, you have to write `fixes` instead of `fix for` in the commit message.

Thanks again for the fix!

perrinjerome:
Ah yes, *fixes* ! sorry for that and thank you !

PR #1438, opened 2019-11-08
Fix inference from type comment for function parameter with dot

perrinjerome:
fix for https://github.com/davidhalter/jedi/issues/1437


davidhalter:
Thanks a lot for the fix! I guess this makes sense.

Also note that the Python 3.8 CI issues don't matter. They have existed before. I just finally need to fix that (Python 3.8 has passed before it changed - some beta versions are not the same anymore).

PR #1439, opened 2019-11-08
third-party: use python/typeshed

blueyed:
I could not find a specific reason to use a fork here, since there
appear to be no changes.

Also updates it to the current version, fixing/improving many things.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "jedi/third_party/typeshed"]
-	path = jedi/third_party/typeshed
-	url = https://github.com/davidhalter/typeshed.git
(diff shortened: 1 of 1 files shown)


davidhalter:
There are definitely changes, it's just that we should probably update from time to time.

blueyed:
Changes in your repo?
(Github just said that it was behind master?!)

davidhalter:
There is a `jedi` branch :)

blueyed:
I see.. :)
There are conflicts, so I think it is better for you to update it.

blueyed:
btw: you could change the default branch in your fork.

davidhalter:
I updated the default branch. :)

PR #1440, opened 2019-11-14
Find active conda environment and set it as default (if there is one)

shanest:
As we discussed in `jedi-vim` issue 907 (https://github.com/davidhalter/jedi-vim/issues/907), this is a PR that makes `get_default_environment` use an activated conda environment if it doesn't find a virtualenv.  And then it still defaults to system python if it finds neither.  (I also made `find_virtualenvs` looks for conda environments.) 

I've installed a version of `jedi-vim` that points to this version of `jedi`, and it works well for me :).

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -53,5 +53,6 @@ micbou (@micbou)
 Dima Gerasimov (@karlicoss) <karlicoss@gmail.com>
 Max Woerner Chase (@mwchase) <max.chase@gmail.com>
 Johannes Maria Frank (@jmfrank63) <jmfrank63@gmail.com>
+Shane Steinert-Threlkeld (@shanest) <ssshanest@gmail.com>
 
 Note: (@user) means a github user name.
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -19,6 +19,7 @@ _VersionInfo = namedtuple('VersionInfo', 'major minor micro')
 
 _SUPPORTED_PYTHONS = ['3.8', '3.7', '3.6', '3.5', '3.4', '2.7']
 _SAFE_PATHS = ['/usr/bin', '/usr/local/bin']
+_CONDA_VAR = 'CONDA_PREFIX'
 _CURRENT_VERSION = '%s.%s' % (sys.version_info.major, sys.version_info.minor)
 
 
@@ -189,6 +190,10 @@ def get_default_environment():
     if virtual_env is not None:
         return virtual_env
 
+    conda_env = _get_virtual_env_from_var(_CONDA_VAR)
+    if conda_env is not None:
+        return conda_env
+
     return _try_get_same_env()
 
 
@@ -279,7 +284,7 @@ def find_virtualenvs(paths=None, **kwargs):
             yield virtual_env
             _used_paths.add(virtual_env.path)
 
-        conda_env = _get_virtual_env_from_var('CONDA_PREFIX')
+        conda_env = _get_virtual_env_from_var(_CONDA_VAR)
         if conda_env is not None:
             yield conda_env
             _used_paths.add(conda_env.path)
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -179,7 +179,8 @@ def _calculate_sha256_for_file(path):
 
 def get_default_environment():
     """
-    Tries to return an active Virtualenv. If there is no VIRTUAL_ENV variable
+    Tries to return an active Virtualenv or conda environment.
+    If there is no VIRTUAL_ENV variable or no CONDA_PREFIX variable set
     set it will return the latest Python version installed on the system. This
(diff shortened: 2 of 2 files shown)


davidhalter:
I think `get_cached_default_environment` also needs patches or am I wrong?

davidhalter:
BTW: Don't worry about travis tests, I need to fix those.

shanest:
> I think `get_cached_default_environment` also needs patches or am I wrong?

I'm not sure that I see the need.  The check for the prefix seems to be about pipenv, so hard-coding the 'VIRTUAL_ENV' variable is fine there.  I've never encountered the issue in the comments using `conda` on macOS (i.e. `CONDA_PREFIX` always points to the base, not the `bin` subdirectory).  If you think that issue may arise with `conda` as well, I can change `get_cached_default_environment` as well.

shanest:
At your suggestion, I went ahead and put in a simple patch for `conda` environments in `get_cached_default_environment`, even though I haven't yet come across the particular problem.

davidhalter:
Thanks a lot! I'd be happy to see you contribute further fixes :)

bs:
I'm still seeing this behavior. My steps:

- start vim from an active conda environment other than base
- edit a python file

Vim is showing it's loaded the env that's sourced in my ```.zshrc``` via ```conda activate X```, not the env I've switched to. Is vim sourcing the shell and then flipping the env back?

Happy to provide any more details needed. Thanks for all of your work!

davidhalter:
Since I don't run conda, other people will have to debug this. Maybe @shanest can help?

@bs which Jedi version are you running?

bs:
Here's where things start to get really fishy:

- Environment I'm running nvim from: 'base'

From the CLI, let's confirm that we've got the correct Python and Jedi installed and in our $PATH.

```
❯ which python && python --version && pip list | grep jedi
/usr/local/anaconda3/envs/esp/bin/python
Python 3.7.7
jedi                               0.17.0
```

- And now from with nvim, let's see what shell it's in

```
:!echo $SHELL
/bin/zsh
```

- And which conda env the shell is in

```
:!conda env list
# conda environments:
#
base                     /usr/local/anaconda3
esp                   *  /usr/local/anaconda3/envs/esp
snake_jazz               /usr/local/anaconda3/envs/snake_jazz
```

- And python version from the shell:

```
:!python -c "import sys; print(sys.version)"
3.7.7 (default, Mar 26 2020, 10:32:53)
[Clang 4.0.1 (tags/RELEASE_401/final)]
```

- What about the python version in nvim?

```E319: No "python" provider found. Run ":checkhealth provider"```

Ruh roh! Checkhealth shows ...

```
## Python 2 provider (optional)
  - WARNING: No Python executable found that can `import neovim`. Using the first available executable for diagnostics.
  - ERROR: Python provider error:
    - ADVICE:
      - provider/pythonx: Could not load Python 2:
          /usr/bin/python2 does not have the "neovim" module. :help |provider-python|
          /usr/bin/python2.7 does not have the "neovim" module. :help |provider-python|
          python2.6 not found in search path or not executable.
          /usr/local/anaconda3/envs/esp/bin/python is Python 3.7 and cannot provide Python 2.
  - INFO: Executable: Not found

## Python 3 provider (optional)
  - INFO: `g:python3_host_prog` is not set.  Searching for python3 in the environment.
  - INFO: Multiple python3 executables found.  Set `g:python3_host_prog` to avoid surprises.
  - INFO: Executable: /usr/local/anaconda3/envs/esp/bin/python3
  - INFO: Other python executable: /usr/bin/python3
  - INFO: Python version: 3.7.7
  - INFO: pynvim version: 0.4.1
  - OK: Latest pynvim is installed.
```

OK so ```:python``` maps to python2 and ```:python3``` gives:

```
3.7.7 (default, Mar 26 2020, 10:32:53)
[Clang 4.0.1 (tags/RELEASE_401/final)]
```

What's strange is the path ```/usr/local/anaconda3/envs/esp/bin/python3``` is correct in :checkhealth.

- Airline after opening a python file into a buffer shows a message about ```searching  interpreter``` and then lands on

```dotfiles/setup.py  Python 3.7.7 64-bit ('snake_jazz': conda)```

- Then doing something that invokes COC throws

```
[coc.nvim] Jedi error: Traceback (most recent call last):
  File "completion.py", line 698, in <module>
[coc.nvim] Jedi error: Traceback (most recent call last):
[coc.nvim] Jedi error:     raise RuntimeError("Jedi version %s too old, requires >= 0.17.0" % (jedi.__version__))
RuntimeError: Jedi version 0.15.2 too old, requires >= 0.17.0
```

I'm pouring over my my system to find any conda or pip env with Jedi 0.15.2 installed. The only install I can find other than my current env is in ```base``` and it's ```0.14.1```.

Thanks for the quick response!

davidhalter:
Maybe the Jedi version is a submodule and you need to update it? I have no idea about coc.

PR #1441, opened 2019-11-15
Improve annotations for params of compiled functions

Endilll:
### Problem
Currently annotation strings for parameters of compiled functions are created directly from annotation objects by means of `str()`. This leads to noisy annotations that are shown to user: `clip: <class 'vapoursynth.VideoNode'>, width: typing.Union[int, NoneType]=None`. Just two of them may look fine, but few dozens of them are definitely not.
### Desired output
Annotations for parameters of compiled functions as if they had sources. Using example above, I believe it should be as following: `clip: VideoNode, width: Optional[int]`
### Proposed solution
This is just a proof-of-concept of desired output to be achievable without (much) hacking.

It tries to be as compatible as possible with the rest of codebase, so it's (a bit) ugly. One more reason why there's no best solution for this is that any attempt to subscript `Optional` leads to `Union[..., NoneType]`.

Going to the hacking side, I checked that even though `Union.__origin__` and `Union.__args__` aren't documented, they are available on every CPython since 3.5. And of course, I'd like to use `get_origin()` and `get_args()` that are available since 3.8.

I'm not familiar with how things are done in jedi codebase and probably missing something, so I'm looking forward to your feedback and working further on this PR accordingly.

### DIFF
--- a/jedi/inference/compiled/access.py
+++ b/jedi/inference/compiled/access.py
@@ -96,10 +96,58 @@ def safe_getattr(obj, name, default=_sentinel):
     return attr
 
 
-SignatureParam = namedtuple(
-    'SignatureParam',
-    'name has_default default default_string has_annotation annotation annotation_string kind_name'
-)
+# SignatureParam = namedtuple(
+#     'SignatureParam',
+#     'name has_default default default_string has_annotation annotation annotation_string kind_name'
+# )
+
+class SignatureParam:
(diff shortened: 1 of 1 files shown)


davidhalter:
It's intentional that the annotations are a bit noisy. IMO if you just want to get rid of `typing.`, just `str.replace` it.

Compiled annotations are now quite a bit better than when you wrote this pull request. So feel free to try again and let me know if some things are not working for you.

Issue #1443, opened 2019-11-18
Broken code completion with (Type[TypeVar]) -> TypeVar


pappasam:
Thank you in advance for looking into this! Jedi is an amazing library and I rely on it every day.

## Issue

`typing.TypeVar`, when used in conjunction with `typing.Type`, breaks completion in `jedi==0.15.1`.

* Python versions tested: `3.7.5` and `3.8.0`, to the same effect.

## Benefits to fixing this issue (eg, why prioritize?)

Although this issue may seem esoteric, it's a pattern I love and one I think may be underused because of its lack of jedi support.

IMHO, fixing this issue will make generically typed constructor functions work well with text editors, which will increase the usage of generically typed functions as constructors (a pattern I much love). This will move the whole Python community toward "typed Duck typing" and into a brave new future for Python.

That said, I'm not a Jedi master. I can try looking into it, but I raise the issue in case someone else with more command of the force can be of service.

## Broken Example: typing.Type with typing.TypeVar

```python
import jedi

source = """
from typing import Type, TypeVar, Any

T = TypeVar("T")

class MyClass:
    def __init__(self, my_attr: Any) -> None:
        self.my_attr = my_attr

def generic_func(a: Any, b: Type[T]) -> T:
    return b(a)

my_class = generic_func("another_value", MyClass)
my_class.
""".strip()

script = jedi.Script(source, 13, 9, "source.py")
print([completion.complete for completion in script.completions()])
```

This prints the following:

```console
[]
```

## Working Example: typing.Type with a class object

```python
import jedi

source = """
from typing import Type, TypeVar, Any

T = TypeVar("T")

class MyClass:
    def __init__(self, my_attr: Any) -> None:
        self.my_attr = my_attr

def generic_func(a: Any, b: Type[MyClass]) -> MyClass:
    return b(a)

my_class = generic_func("another_value", MyClass)
my_class.
""".strip()

script = jedi.Script(source, 13, 9, "source.py")
print([completion.complete for completion in script.completions()])
```

This prints:

```console
['my_attr', '__annotations__', '__class__', '__delattr__', '__dict__',
'__dir__', '__doc__', '__eq__', '__format__', '__getattribute__', '__hash__',
'__init__', '__init_subclass__', '__module__', '__ne__', '__new__',
'__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__',
'__slots__', '__str__']
```

## Notes

[...]

PR #1448, opened 2019-11-27
Fix missing inference for typing.Type[typing.TypeVar]

pappasam:
Fixes: https://github.com/davidhalter/jedi/issues/1443

Fix requires:

1. Adding a `get_generics` method to TypingClassValueWithIndex.
2. The `elif` block suggested by @davidhalter 
3. A special case handling of class_values. Based on some experimentation, the most straightforward way I could think to handle cases where objects are class values, and not class instances, was to add a boolean flag that would be set True only by 'Type' values.

Notes: 

* I reverted an initial change because I was using the version of parso that is currently deployed on pypi. Jedi's current master requires the development version of parso.
* All existing tests currently pass. I can work on adding some tests if @davidhalter agrees with this initial approach.

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -553,7 +553,7 @@ class Definition(BaseDefinition):
                 typ = 'def'
             return typ + ' ' + self._name.get_public_name()
 
-        definition = tree_name.get_definition(include_setitem=True) or tree_name
+        definition = tree_name.get_definition() or tree_name
         # Remove the prefix, because that's not what we want for get_code
         # here.
         txt = definition.get_code(include_prefix=False)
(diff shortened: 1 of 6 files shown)


davidhalter:
I just fixed the CI, so this will pass once we merge. Just ignore the Python 3.8 failure for now (on Travis), this is my responsibility. 

If you add a few tests (I recommend using the files in `test/completion/pep*`). In these tests `#? int()` means that the next line resolves to an int instance.

pappasam:
@davidhalter tests have been added. I've followed the same pattern provided by the other TypeVar tests.

davidhalter:
Great work! Thanks for this.

If you want I'm happy to search another issue you could fix. Just let me know, there's a couple :)

Issue #1446, opened 2019-11-25
Jedi showing system and home directory folders


Precrisk:
**OS: Arch Linux**
**Editor: Neovim**
**Code Completion: ncm2, ncm2-jedi**

**Expected Behavior:** 
after entering keyword `import` jedi will show modules from current project, standard library and system/virtualenv installed libraries. 

**Current Behavior:**
after entering keyword `import` jedi shows modules from all the sources listed above AND system directories like usr, dev, bin, etc, home, etc. and directories in home folder like Videos, Downloads, etc. 
there is no way to import these "modules" as these are not valid modules and even if these were modules these are not in `path`. 
and some 'modules' listed here like 'Calculator', 'Tasks', etc are valid python module but they are in different directory which i cannot import from my current project. 

Screenshots:
![Screenshot from 2019-11-25 08-48-26](https://user-images.githubusercontent.com/47627625/69512260-5decbc80-0f69-11ea-84a4-b1b795f64bda.png)
(this occurs even without `from home.precrisk.Python`)

pappasam:
I have the same issue.

OS: Ubuntu 18.04
Editor: Neovim
Code Completion: deoplete + LanguageClient-neovim + jedi-language-server

davidhalter:
Essentially the problem is that Jedi doesn't find a root for your Python project. Create a `setup.py`, `requirements.txt` or probably also an `__init__.py` somewhere to signal that it's a top level folder.

I'm happy to discuss improvements, but note that it's not easy, because we're trying to satisfy a lot of different stakeholders here :).

ComputerHardman:
I have the same issue.

OS: Windows 10 (1809, 1803, 1709)
Editor: VSCode
Code Completion: jedi with the python extension

I have been able to replicate this with this code snippet 
```
import jedi
source = "import Pro"
script = jedi.Script(source, 1, len('import t'), 'example.py')
print(script)
completions = script.completions()
print(completions)
print(completions[0].complete)
```
the code snippet will show that folders like `Program Files (x86)` are suggested as completions.

davidhalter:
If anyone has a better solution for this, please look at `_get_sys_path` in `jedi/api/project.py` and let me know. The logic is not that complicated. It is definitely intentional that a lot of sys paths are added, but this might be annoying for import completion. 

So in the end there's probably a balance we have to find to avoid issues like yours, but it's probably not avoidable that some unwanted folders are included.

Precrisk:
[...]

PR #1451, opened 2019-11-29
Fix: no longer shows folders recursively to root

pappasam:
Fixes: https://github.com/davidhalter/jedi/issues/1446

In the previous implementation, Jedi would's traverse_parents function
traversed parent directories to the system root every time. This would
inadvertently add every folder to the system root every time. Obviously,
this is not the behavior desired for the import system.

This pull request provides a new argument to the traverse_parents
function, "root", which represents the root parent for the search. This
argument defaults to None, thereby preserving the existing behavior of
the function.

I chose to duplicate some code for performance reasons. Since I'm trying
to avoid too much path manipulation magic, we do:

* a search to a valid specified root, OR
* a simple upward search until hitting the system root when there is no
valid root specified.

### DIFF
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -110,7 +110,10 @@ class Project(object):
                 suffixed += discover_buildout_paths(inference_state, inference_state.script_path)
 
                 if add_parent_paths:
-                    traversed = list(traverse_parents(inference_state.script_path))
+                    traversed = list(traverse_parents(
+                        inference_state.script_path,
+                        root=self._path,
+                    ))
 
                     # AFAIK some libraries have imports like `foo.foo.bar`, which
                     # leads to the conclusion to by default prefer longer paths
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -110,7 +110,17 @@ class Project(object):
                 suffixed += discover_buildout_paths(inference_state, inference_state.script_path)
 
                 if add_parent_paths:
-                    traversed = list(traverse_parents(inference_state.script_path))
+                    # Collect directories in upward search until we:
+                    #   1. Hit any directory without an __init__.py, AND
+                    #   2. Are above self._path.
+                    traversed = []
+                    no_init = False
+                    for parent_path in traverse_parents(inference_state.script_path):
+                        if not os.path.isfile(os.path.join(parent_path, "__init__.py")):
+                            no_init = True
+                        if no_init and not parent_path.startswith(self._path):
+                            break
+                        traversed.append(parent_path)
 
                     # AFAIK some libraries have imports like `foo.foo.bar`, which
                     # leads to the conclusion to by default prefer longer paths
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -110,16 +110,15 @@ class Project(object):
                 suffixed += discover_buildout_paths(inference_state, inference_state.script_path)
 
                 if add_parent_paths:
-                    # Collect directories in upward search until we:
-                    #   1. Hit any directory without an __init__.py, AND
-                    #   2. Are above self._path.
+                    # Collect directories in upward search by:
+                    #   1. Skipping directories with __init__.py
+                    #   2. Stopping immediately when above self._path
                     traversed = []
-                    no_init = False
                     for parent_path in traverse_parents(inference_state.script_path):
-                        if not os.path.isfile(os.path.join(parent_path, "__init__.py")):
-                            no_init = True
-                        if no_init and not parent_path.startswith(self._path):
+                        if not parent_path.startswith(self._path):
                             break
+                        if os.path.isfile(os.path.join(parent_path, "__init__.py")):
+                            continue
                         traversed.append(parent_path)
 
                     # AFAIK some libraries have imports like `foo.foo.bar`, which
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -110,10 +110,7 @@ class Project(object):
                 suffixed += discover_buildout_paths(inference_state, inference_state.script_path)
 
                 if add_parent_paths:
-                    traversed = list(traverse_parents(
-                        inference_state.script_path,
-                        root=self._path,
-                    ))
+                    traversed = list(traverse_parents(inference_state.script_path))
 
                     # AFAIK some libraries have imports like `foo.foo.bar`, which
                     # leads to the conclusion to by default prefer longer paths
--- a/jedi/common/utils.py
+++ b/jedi/common/utils.py
@@ -2,15 +2,31 @@ import os
 from contextlib import contextmanager
 
 
-def traverse_parents(path, include_current=False):
+def traverse_parents(path, root=None, include_current=False):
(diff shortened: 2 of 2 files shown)


pappasam:
Tests pass locally on Python 3.7.

davidhalter:
My question here is probably much more: Why do we even do this part: `a search to a valid specified root`.

Not that it's completely wrong, but if we have a valid root, it mostly likely is the only place where we need a sys path entry. Would you agree or not? At the same time there's definitely structures where the code is in e.g. `lib/foo/` instead of `foo/`. But maybe we should find a separate solution for these kind of things. :)

We can filter even before that, like `if add_parent_paths and self._path is None:`. But that definitely wouldn't cover all cases. I'm not sure what the best solution would be.

davidhalter:
Actually, it could be interesting to do your solution, but only start when there's no `__init__.py` files anymore. This would cover the `lib/foo` case I wrote above and pretty much every other case as well. This could be easily done by checking each path for an `__init__.py` and ignoring that directory if an `__init__.py` is there. What do you think?

> I chose to duplicate some code for performance reasons.

Don't worry about performance here, it's not often called. Focus on readability (it's fine now).

pappasam:
Hmm, maybe I'll rethink this PR.

@davidhalter To confirm, you're suggesting that, for `_get_sys_path`, we only want to add parent paths if:

1. They contain an `__init__.py`.
2. All child directories of each parent also contain an `__init__.py`.

For example, in `project/foo/bar`, where `project.foo.bar.__init__.py` and `project.foo.__init__.py` exist, but `project.__init__.py` does not exist, we should return `project/foo/bar` and `project/foo`, but not `project`. Is that correct? 

If yes, I think a better solution would simply change the usage of `traverse_parents` within `_get_sys_path` and to not change the `_get_sys_path` implementation at all.

pappasam:
A test for the implementation of my above interpretation failed tests on integration:

```traceback
__________________________ test_static_analysis[attribute_error.py] ___________________________

static_analysis_case = <StaticAnalysisCase: attribute_error.py>
environment = <Environment: 3.7.5 in /home/sroeca/src/Personal/jedi/.tox/py>

    def test_static_analysis(static_analysis_case, environment):
        skip_reason = static_analysis_case.get_skip_reason(environment)
        if skip_reason is not None:
            pytest.skip(skip_reason)
        else:
>           static_analysis_case.run(assert_static_analysis, environment)

test/test_integration.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test/run.py:322: in run
    compare_cb(self, analysis, self.collect_comparison())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

case = <StaticAnalysisCase: attribute_error.py>
actual = [(9, 20, 'attribute-error'), (24, 12, 'attribute-error'), (27, 17, 'name-error'), (35, 4, 'attribute-error'), (38, 2, 'attribute-error'), (44, 12, 'name-error'), ...]
desired = [(9, 20, 'attribute-error'), (24, 12, 'attribute-error'), (27, 17, 'name-error'), (35, 4, 'attribute-error'), (38, 2, 'attribute-error'), (44, 12, 'name-error'), ...]

    def assert_static_analysis(case, actual, desired):
        """A nicer formatting for static analysis tests."""
        a = set(actual)
        d = set(desired)
    Test %r failed.
    not raised  = %s
    unspecified = %s
    """ % (case, sorted(d - a), sorted(a - d))
E       AssertionError:
E         Test <StaticAnalysisCase: attribute_error.py> failed.
E         not raised  = []
E         unspecified = [(110, 7, 'import-error')]
E
E       assert [(9, 20, 'att...-error'), ...] == [(9, 20, 'attr...-error'), ...]
E         Left contains one more item: (110, 7, 'import-error')

test/test_integration.py:27: AssertionError
================ 1 failed, 2587 passed, 45 skipped, 5 xfailed in 72.13 seconds ================
ERROR: InvocationError for command /home/sroeca/src/Personal/jedi/.tox/py/bin/pytest (exited with code 1)
___________________________________________ summary ___________________________________________
ERROR:   py: commands failed
```

pappasam:
Since filtering for `__init__.py` fails the above integration test, I'll simply move my path check to `_get_sys_path` to avoid changing too much code and push that update.

pappasam:
@davidhalter ok, I think I've translated your suggestions into code. Tests still pass locally and the extra, unnecessary directories are gone from my import completion.

jedi/api/project.py
```diff
@@ -110,7 +110,17 @@ def _get_sys_path(self, inference_state, environment=None, add_parent_paths=True
...
                 if add_parent_paths:
-                    traversed = list(traverse_parents(inference_state.script_path))
+                    # Collect directories in upward search until we:
+                    #   1. Hit any directory without an __init__.py, AND
+                    #   2. Are above self._path.
+                    traversed = []
+                    no_init = False
+                    for parent_path in traverse_parents(inference_state.script_path):
+                        if not os.path.isfile(os.path.join(parent_path, "__init__.py")):
+                            no_init = True
```
davidhalter: IMO it probably makes more sense to just:

```
if os.path.isfile(os.path.join(parent_path, "__init__.py")):
    continue
```

However that should be after the break (to avoid not hitting that). The `no_init` will not be needed anymore.

This is IMO the better solution, because sometimes people have python packages nested in weird ways (they compensate with path manipulation magic to get it working in the end).

What do you think?

pappasam: I see. So we avoid adding any directory to the path that does have an `__init__.py`.

I'll push an update and you can see what you think.

pappasam: The suggested implementation breaks a ton of integration tests. Maybe that's desired and the tests are wrong, but these tests are a bit beyond my understanding of the library at the moment. I'll push the implementation, but @davidhalter please check it. We can easily revert if you change your mind.

davidhalter:
Ok. Thanks for looking into this! I think I'm going to finish this. It might be quite complicated, my first guess is relative imports. But then again stubs are failing. This is probably a more fundamental matter. I will probably merge your work locally and add an additional commit of mine.

No idea what's the issue yet, but thanks anyway for fixing part of the issue!

davidhalter:
After playing around with this quite a bit, I finally found a solution that might be working just good enough so (almost) nobody complains anymore. I found that the easiest way to do that is probably to make import completions behave a bit differently than import type inference. Type inference is now a bit more forgiving than completions.

Let me know what you think, happy to make further improvements. Please let me know if this fixes the issue for you.

Also still happy to help you with any more issues you're trying to fix here :)

pappasam:
Awesome, thanks for all the help, and for this awesome library! I'll keep sending issue fixes when I find them / updating issues otherwise.

PR #1460, opened 2019-12-16
Fix simple typo: wheter -> whether

### LINKED ISSUE #1459: Fix simple typo: wheter -> whether
There is a small typo in jedi/inference/value/dynamic_arrays.py.
Should read `whether` rather than `wheter`.

timgates42:
Closes #1459


davidhalter:
Thanks for fixing those :)

Issue #952, opened 2017-08-11
PEP484 support for CompiledObjects (strings)


rp-:
Compiled python objects don't get ```__annotations__``` evaluated and so have no completion support for e.g. return types.

I tried to hack it in, but I couldn't really figure out how to parse the ```__annotations__``` string to a jedi.tree.PythonNode and correctly feed it to the pep0484 infer functions.
But what I understood from the code is, it should be possible, but I'm not sure how well it fits in the current design.

davidhalter:
> but I'm not sure how well it fits in the current design.

Then maybe we have to refactor the current design. :)

---

Anyway good thoughts! I agree that we should make that possible. There might need to be a kind of different API, but that's fine. I think if you're doing this, just try to do it as clean as possible and if you're not sure how to add new abstractions, we can always discuss here.

rp-:
Ok, sadly I'm lacking some knowledge of jedi code.
I can't really figure out how to create the prober PythonNodes that should go into the pep0484._evaluate_for_annotion function.
Right now I try to parse the ```__annotation__``` string from the CompiledObject and pass the result nodes to the pep0484.infer_return_types function, which doesn't really work.

davidhalter:
I just today added support for `__annotations__['return']`. However it only works for types and not strings. Just be aware of that. (Change not included in 0.14.0)

Keeping this issue open to eventually add support for strings as well.

davidhalter:
Works now with Python3.6+ on master branch.

Issue #1458, opened 2019-12-13
Use `__annotations__` to decide type of dynamic attributes as specified in PEP526


fmagin:
https://github.com/davidhalter/jedi/issues/1347 is related to this and already uses this for return types implemented via https://github.com/davidhalter/jedi/commit/aa2dc6be09da315ff8ca28553fde0bda01ee0132
It seems that accessing and parsing `__annotations__` is thus something that is considered useful for Interpreter contexts. This would also be another way to deal with issues like #1453 or https://github.com/davidhalter/jedi/issues/1378 that are currently dealt with via https://github.com/davidhalter/jedi/commit/923fcf95d991286302a6233eb4eb760b02fad1e3 which is not an ideal solution compared to this IMO

My suggestion is to also parse `__annotations__` on objects and classes and thus support [PEP526: Class and instance variable annotations](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations). Those are mostly ignored by the runtime and can be used to provide type information to IDEs and completers to specify that some attribute that might be hidden behind some `__getattr__` magic or similar has a certain type without having to actually run the code (as is currently done). This should also allow to infer the type of an attribute that is e.g. declared as `List[T]` which could otherwise only be annotated if turning it into a property and annotating the property itself.

Examples:

```python
from __future__ import annotations # See below
class SomeClass():
    def __init__(self, x: int, text: str):
        self.x = x
        self.text = text


class DynamicVarClass(object):
    some_dynamic_attribute: SomeClass

    def __getattr__(self, attr):
        if attr == 'some_dynamic_attribute':
            return SomeClass(1,'foobar')

    def __dir__(self):
        return ['some_dynamic_attribute'] + object.__dir__(self)

class TypedAttributeClass():
    l: List[int]
    def __init__(self):
        self.l = [1,2,3]
```

```
In [14]: DynamicVarClass().__annotations__
Out[14]: {'class_var': 'SomeClass'}

In [16]: TypedAttributeClass().__annotations__                                                                                                    
Out[16]: {'l': 'List[int]'}
```

It must be noted that according to [PEP 563](https://www.python.org/dev/peps/pep-0563/) those annotations will always be strings that have to be [resolved](https://www.python.org/dev/peps/pep-0563/#id7). Without the `from __future__ import annotations` the dict contains the class/type directly.

PoC/Test:

```python
from __future__ import annotations
import jedi
jedi.set_debug_function() 


class SomeClass():
[...]

PR #1461, opened 2019-12-18
WIP: Improve Usage of Type Annotations (Dynamic and Static) in Interpreter Context

fmagin:
This is a follow up to https://github.com/davidhalter/jedi/issues/1458.

General goals/ideas:
- Interpreter should use `__annotations__` of the object it is trying to auto complete on (mostly functions for now).
  - This should work already, 1c62ecb extends this so [PEP 563](https://www.python.org/dev/peps/pep-0563) annotations should work too. There are still some [pitfalls](https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-type-checking), but this should be strictly better than the current implementation 
- If a `-stub` package is available there should be the option for the completer to use this instead of what the object provides.
  - First this should work for any typical stub package that is installed in the used virtualenv and for `.pyi` files in the actual module.
  - There are two extremes here that I have actually encountered and have some interest in covering. I think the latter is better and is also useful in general.
     - An RPC implementation that wants to lie about the annotations of the object (based on the annotations of the remote object)
     - the possibility to force that all completion for a module happens based on local stub files, which would be applied based on the bridged fake type and not the true type (RPC object) of the object so no RPC calls need to happen for completion.
 
Current Todos:
- [x] Handle PEP 563 annotations
- [ ] Test for PEP 563 annotations
- [ ] Decide on general implementation for  annotations in `.pyi` files next to `.py` files
- [ ] Decide on general implementation for  annotations in separate `-stub` package

Questions:
I do not have a good high level concept of the jedi architecture yet. The first issue is where the stub based inference happens for Script contexts, how much of this can be used for Interpreter contexts and if that requires refactoring to move code out of the interpreter context so it can be shared.

### DIFF
--- a/jedi/inference/compiled/access.py
+++ b/jedi/inference/compiled/access.py
@@ -531,7 +531,8 @@ class DirectObjectAccess(object):
 
     def get_return_annotation(self):
         try:
-            o = self._obj.__annotations__.get('return')
+            import typing
+            o = typing.get_type_hints(self._obj).get('return')
         except AttributeError:
             return None
 
--- a/jedi/inference/compiled/access.py
+++ b/jedi/inference/compiled/access.py
@@ -533,6 +533,15 @@ class DirectObjectAccess(object):
         try:
             import typing
             o = typing.get_type_hints(self._obj).get('return')
+            return self._create_access_path(o)
+        except ImportError:
+            # Old behaviour
+            pass
+        except AttributeError:
+            return None
+
+        try:
+            o = self._obj.__annotations__.get('return')
         except AttributeError:
             return None
 
--- a/jedi/inference/compiled/value.py
+++ b/jedi/inference/compiled/value.py
@@ -49,7 +49,6 @@ class CompiledObject(Value):
     def py__call__(self, arguments):
         return_annotation = self.access_handle.get_return_annotation()
         if return_annotation is not None:
-            # TODO the return annotation may also be a string.
             return create_from_access_path(
                 self.inference_state,
                 return_annotation


jedi/inference/compiled/access.py
```diff
@@ -504,7 +504,8 @@ def _get_signature(self):
 
     def get_return_annotation(self):
         try:
-            o = self._obj.__annotations__.get('return')
+            import typing
+            o = typing.get_type_hints(self._obj).get('return')
```
davidhalter: Good idea. I didn't really know about this.

fmagin: Is there some way to get the original object ( let's call it `owner`) that the function that is being looked at( `_obj`) came from? To handle generics this is relevant, because otherwise this will just be a type variable like `{'return': ~T}` and `owner.__parameters__` and `owner.__orig_class__.__args__` is required to resolve this. There is no nice API call for that yet and `__orig_class__` is technically an "undocumented internal attribute", so if this is too much work, I don't think it is worth the hassle for now.

davidhalter: Not that I know of. For classes in general it's pretty hard to figure out their context unfortunately.

For functions it's a bit easier. For functions you can use stuff like `func.__globals__` or in case of a method `func.__method__.__globals__`. There's also the `__code__` object on every function, where you have stuff like `co_filename`, `co_firstlineno`. For classes there's unfortunately mostly nothing, it's just `Class.__module__`.

fmagin: I can't think of a situation where the context of a class (nor class instance) would be needed to resolve generics.

There is `_obj.__self__` for bound methods which should easily cover the [case](https://mypy.readthedocs.io/en/latest/generics.html#defining-generic-classes)  I was thinking of. For generic functions the input arguments would need to be checked, but that definitely sounds like more work. And not something I really needed so far.

davidhalter:
This might fix #952

davidhalter:
Sorry, I somehow over read your question.

> I do not have a good high level concept of the jedi architecture yet. The first issue is where the stub based inference happens for Script contexts, how much of this can be used for Interpreter contexts and if that requires refactoring to move code out of the interpreter context so it can be shared.

I'm not sure what exactly you mean. What are you trying to accomplish? If you have an example where you would want type inference to be different, that would also help. You could also do that stuff in a separate pull request.

BTW: Stubs are used in "interpreter" contexts as well AFAIK and if that's not the case that's a bug. If it doesn't work, there's ways of loading stubs and putting them together with the non-stub objects in `jedi.inference.gradual.typeshed` and there's logic that merges stuff for Python (see `jedi.inference.compiled.mixed`).

---- 

I would really like to merge the changes you have made now, it's just a matter of getting the CI passing and writing a few tests.

fmagin:
I am not sure when I will have to the time to look into the CI failures, because this will probably require me to setup the test environments to properly investigate. The issue is probably fairly simple, but I have more urgent things to do until at least next week.

davidhalter:
I used `get_type_hints` now and did it myself. Sorry, but I didn't want to wait anymore, haha :)

Anyway, I'm still welcome to discuss other changes you might want to do. Now stuff like `Tuple[int]` works as well if it's used as compiled annotation.

davidhalter:
I'm closing because of inactivity, but I'm still happy to discuss any changes you want to make. You had definitely some good ideas!

Issue #1401, opened 2019-09-15
Django classes plugin


ANtlord:
Hello! Regarding this [comment](https://github.com/davidhalter/jedi/issues/626#issuecomment-518029536) I create new issue.

So... what is the object a class is passed of as argument `cls` of the [function](https://github.com/davidhalter/jedi/blob/master/jedi/plugins/stdlib.py#L785)?

I though it's a class of that object which the completions for. I test the code
```python
import jedi
source = '''
from django.db import models

class Product(models.Model):
    product_name = models.CharField()
p1 = Product()
p1_name = p1.product_name
p1_name.'''
script = jedi.Script(source, 8, len('p1_name.'), 'example.py')
completions = script.completions()
```
so argument `cls` is `ModelBase` and that is true for `p1` but it is looking for completion for `p1_name`. I understand that the object is clone of `p1.product_name` but I don't understand the relation that Jedi sees. Would you mind to explain this?

davidhalter:
Not sure if I understand you correctly, please let me know if I misunderstood you.

The cs is `<ClassValue: <Class: Product@4-6>>` and metaclasses is `S{<ClassValue: <Class: ModelBase@67-378>>}`. This is exactly what you want. The way type inference works in Jedi is this way:
1. `p1_name.` -> Find `p1_name=...`
2. Found `p1_name = p1.product_name`
3. Find `p1`
4. Found `p1 = Product()`
5. Find `Product`
6. Found `class Product`
7. Execute `Product`, because of `p1 = Product()` -> Returns an instance of `Product`.
8. Search `product_name` on that instance, because of the second last line.
9. Find completions of the returned object of (8)

Now what you need to do is provide the proper filters so that when `product_name` is looked at in (8) gets the return value of a `string` and not `CharField`.

Does that help?

Filters BTW are just there to find the right attribute on an instance/module/class/object.

ANtlord:
That helps. Sorry for my "logical" sequence in that question. It's miracle that you got me.

 But I've got new question. How can I create a filter of str, int, list etc.?

davidhalter:
Please have a look at the `EnumMeta` case in `get_metaclass_filters`. There a DictFilter is created. To create a DictFilter with e.g. an attribute `foo` that is an int, you could for example do this:

```
DictFilter({'foo': cls.evaluator.builtins_module.py__getattribute__('int')})
```

[...]

PR #1467, opened 2020-01-01
Django classes plugin

ANtlord:
The PR is related to https://github.com/davidhalter/jedi/issues/1401


jedi/plugins/stdlib.py
```diff
@@ -781,6 +781,63 @@ def _os_path_join(args_set, callback):
...
+        model_instance_field_type, = cls.inference_state.builtins_module.py__getattribute__('bool')
+        return DjangoModelField(model_instance_field_type, field).name
+
+    if field_tree_instance.name.string_name == 'DecimalField':
+        model_instance_field_type, = cls.inference_state.import_module(('decimal',)).py__getattribute__('Decimal')
+        return DjangoModelField(model_instance_field_type, field).name
+
+    if field_tree_instance.name.string_name == 'ForeignKey':
+        # TODO: infer related object class and make a filter for that class
+        return
```
davidhalter: This could look something like

```
if isinstance(field_tree_instance, TreeInstance):
    # Private access, don't worry about it and let it be my problem
    argument_iterator = field_tree_instance._arguments.unpack()
    next(argument_iterator, None)
    key, lazy_values = next(argument_iterator, (None, None))
    if key is None and lazy_values is not None:
        for value in lazy_values.infer():
            string = value.get_safe_value(default=None)
            if isinstance(value, str):
                for v in cls.parent_context().py__getattribute__(value):
                    yield DjangoModelField(v, field)
            else:
                yield DjangoModelField(value, field)
```

jedi/plugins/stdlib.py
```diff
@@ -781,6 +781,77 @@ def _os_path_join(args_set, callback):
...
+        model_instance_field_type, = cls.inference_state.builtins_module.py__getattribute__('bytes')
+        return DjangoModelField(model_instance_field_type, field).name
+
+    if field_tree_instance.name.string_name == 'BooleanField':
+        model_instance_field_type, = cls.inference_state.builtins_module.py__getattribute__('bool')
+        return DjangoModelField(model_instance_field_type, field).name
+
+    if field_tree_instance.name.string_name == 'DecimalField':
+        model_instance_field_type, = cls.inference_state.import_module(('decimal',)).py__getattribute__('Decimal')
+        return DjangoModelField(model_instance_field_type, field).name
```
davidhalter: I think in general it would be nice to create something like

```
mapping = {'IntegerField': (None, 'str'), 'DecimalField': ('decimal', 'Decimal')}

def infer_django_field(...):
    try:
        module_name, attribute_name = mapping[field_tree_instance.name.string_name]
    except KeyError:
        pass
    else:
        if module_name is None:
            module = cls.inference_state.builtins_module
        else:
            module = cls.inference_state.import_module((module_name,))
        attribute, = module.py__getattribute__(attribute_name)
        return DjangoModelField(...).name
    # Handle foreign key stuff here.
```

jedi/plugins/stdlib.py
```diff
@@ -790,10 +861,37 @@ def wrapper(cls, metaclasses):
                 return [DictFilter({
                     name.string_name: EnumInstance(cls, name).name for name in filter_.values()
                 })]
+
+            if metaclass.py__name__() == 'ModelBase' \
+                    and metaclass.get_root_context().py__name__() == 'django.db.models.base':
+                django_dict_filter = new_django_dict_filter(cls)
+                if django_dict_filter is not None:
+                    return django_dict_filter
```
davidhalter: Please put all of the code you wrote into `jedi/plugins/django.py`. This get_metaclass_filters function can also be copied (obviously without the EnumMeta stuff).

jedi/plugins/stdlib.py
```diff
@@ -790,10 +861,37 @@ def wrapper(cls, metaclasses):
...
+    def name(self):
+        return ValueName(self, self._name.tree_name)
+
+    def _get_wrapped_value(self):
+        obj, = self._cls.execute_with_values()
+        return obj
+
+    def get_filters(self, origin_scope=None):
+        for f in self._get_wrapped_value().get_filters():
+            yield f
```
davidhalter: AFAIK you don't need to define `get_filters` here, that's pretty much what it already does if you don't implement it.

jedi/plugins/stdlib.py
ANtlord: Don't you think if dictionary is overhead in this case? What about a NamedTuple or a class with defined `__slots__`.

davidhalter: Speed doesn't matter at all here. Don't worry about it unless you profile and see that this is indeed an issue.

Or to say it in Knuth's words: "Premature optimization is the root of all evil" :)

Also you have to be careful in optimizing dicts away in Python. They are usually really fast and a dict access might be faster than all the comparisons that you are doing here, which sounds maybe counter intuitive, but dict access is really really fast, because it's what Python uses for everything. Comparisons (and math in general) are really slow, because of all the type checking. So intution can often be wrong for such things (also you are doing a ton of dict accesses anyway. `field_tree_instance.name.`, `cls.inference_state.builtins_module` and `.name is also one. So you have like 10 dict accesses anyway, which is definitely more than the one I proposed :).

davidhalter:
I think now we just need a few tests. Just make sure first that the CI is passing and master is merged.

IMO you can just add Django to the dev requirements for now. I will then care about the test infrastructure. You can add the `django` to `extras_require=...` in `setup.py`.

ANtlord:
My word, I broke all the tests. I'll fix 'em asap

davidhalter:
You changed the typeshed commit, please undo that.

davidhalter:
Now that the CI is passing, start thinking about adding some tests to `test/completion/django.py`. Don't worry about getting the CI passing for all versions, just install Django and get on Python versions passing. I will care about the rest later.

jedi/plugins/django.py
```diff
@@ -0,0 +1,95 @@
...
+                     string = value.get_safe_value(default=None)
+                     if value.name.string_name == 'str':
+                         foreign_key_class_name = value._compiled_obj.get_safe_value()
+                         # TODO: it has only one element in current state. Handle rest of elements.
+                         for v in cls.parent_context.py__getattribute__(foreign_key_class_name):
+                             return DjangoModelField(v, field).name
+                     else:
+                         return DjangoModelField(value, field).name
+
+        raise Exception('Should be handled')
```
blueyed: Better to just log/print it - otherwise it breaks also already working fields.
(changed it to `raise NotImplementedError('TODO: {}'.format(field))` before to get more info)

The failing field was - JFI:
```python
    user = models.ForeignKey(
        settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, blank=True, null=True
    )
```

davidhalter:
@blueyed Good point! Thanks for bringing it up. We could just introduce those stubs like the typeshed stubs. But for now I'd prefer to have a good test suite, so we can get a first version out. We can then probably use the stubs in a different PR.

blueyed:
@davidhalter
> We could just introduce those stubs like the typeshed stubs.

Do you mean the stubs that django-stubs provides?

My question was also if e.g. Jedi could make use of mypy plugins, since I think django-stubs is not just stubs, but also the plugin for mypy in the first place.

I agree that this should not hold this PR back of course - I'm looking forward to better Django support with Jedi very much.. :)

davidhalter:
> Do you mean the stubs that django-stubs provides?

Exactly. We could add it as a subrepo in `jedi/third_party/django-stubs`.

> My question was also if e.g. Jedi could make use of mypy plugins, since I think django-stubs is not just stubs, but also the plugin for mypy in the first place.

Without looking into it, I'm 99% sure that the code will not match Jedi, because there are just fundamental design differences. Mypy is generally not that flexible (it only has to do type inference, no completions or goto) and has a very strange type system. So reusing code is going to be hard. But I'll still check for the 1% chance :).

> I agree that this should not hold this PR back of course - I'm looking forward to better Django support with Jedi very much.. :)

Me too, it's pretty much my job to daily work with those models, haha. :)

blueyed:
> > Do you mean the stubs that django-stubs provides?
> 
> Exactly. We could add it as a subrepo in `jedi/third_party/django-stubs`.

From what I can tell this would not be enough without the mypy plugin.
I've only tried it quickly though, and might be wrong.

> > I'm looking forward to better Django support with Jedi very much.. :)
> Me too, it's pretty much my job to daily work with those models, haha. :)

You might want to try django-stubs then also (only install the stubs, do not enable/configure the mypy plugin) to see if that help Jedi already.  I recommend it otherwise in general, if you're using mypy / typing with your models.. :)

davidhalter:
@ANtlord Are you still working on this or should I take over? :)

ANtlord:
@davidhalter I apologize about the delay. I'm busy seeking a job. I expect I'm going to have a day on the weekend to dive into Jedi test system.

davidhalter:
@ANtlord Don't worry, I'm also happy doing it if you don't have time. But you started this, so I'm happy to let you finish it.

jedi/plugins/django.py
ANtlord: I think I can handle that case but `cls.inference_state.import_module(('django.contrib.auth.models',))` returns empty set in spite of module exists

test/test_inference/test_django.py
```diff
@@ -0,0 +1,64 @@
...
+    field_model_type,
+    expected_fields,
+    Script,
+):
+    source = source_tpl_basic_types.format(field_name, field_model_type)
+    result = Script(source).complete()
+    result = {x.name for x in result}
+    expected_fields_public = [x for x in expected_fields if x[0] != '_']
+    for field in expected_fields_public:
+        assert field in result
```
davidhalter, line 58: Please move this stuff to something like `test/completion/django.py`. The tests in there (have a look at other files), would look like this:

```
class BusinessModel(models.Model):
    char_field = models.CharField()
    date_field = models.DateField()

m = BusinessModel()
#? str()
m.char_field
import datetime
#? datetime.date()
m.date_field
```

davidhalter:
Looks great. Is this more or less finished now?

ANtlord:
Well, I described two bugs in the plugin documentation, which I can't fix just now. In spite of that it's finished.

ANtlord:
How does such black magic work?
```
#? int()
model_instance.integer_field
```

Is it worth to cover cases when the plugin doesn't help with completion?

ANtlord:
Build still failed. Those jobs finish successfully which has proper Python version for environment variable JEDI_TEST_ENVIRONMENT. Do I miss something?

davidhalter:
Don't worry. I will deal with that. I will just find a way to disable Django tests for these cases. As far as I'm concerned, everything is working as it should.

> How does such black magic work?

It's a simple algorithm that searches for `#?` and checks what the types are on the next line.

> Is it worth to cover cases when the plugin doesn't help with completion?

I guess it's fine as it is, but if you want to add such cases, feel free! :)

ANtlord:
@davidhalter Well, I've added a TODO test for ManyToManyField. I'm gonna improve the plugin when I get a chance. The failed jobs in the build are the last problem in the iteration. As far as I got you are able to fix it by yourself, don't you?

davidhalter:
Yeah, I'll take it from here. I'm going to merge and make minor adjustments.

Thanks again for your work! It might take a bit until this is going to be merged, because I'm going to do it locally. However consider it merged :).

Issue #1478, opened 2020-01-23
Feature request: support TypedDict completions


pappasam:
Jedi now supports dictionary completions: https://github.com/davidhalter/jedi/issues/951

Unfortunately, for those of us trying to get the most out of `TypedDict`, the latest master does not support completions for TypedDicts. If we can get the dictionary keys of a TypedDict to autocomplete, I believe users would start using TypedDict a lot more.

```python
# python 3.8
from typing import TypedDict

class MyDict(TypedDict):
    hello: str
    world: int

def hello(x: MyDict):
    z = x["  # it'd be awesome if this autocompleted "hello" and "world"
```

As a stretch goal, we could also consider completing dictionary keys as the dictionary is being created.

```python
# python 3.8
from typing import TypedDict

class MyDict(TypedDict):
    hello: str
    world: int
h: MyDict = {
    "   # it'd be awesome if this autocompleted "hello" and "world"
}
```

Happy to start chipping away at this if you can point me in a profitable direction. And great to see that we're getting dictionary key completion!!!

davidhalter:
I think while the second one would be nice, that's quite a bit of additional code just for completions, so not saying to not do that, it's just probably less realistic.

However the first one looks like we should definitely support it!

If you look into it, one good example of this is probably the Django migration: https://github.com/davidhalter/jedi/pull/1467/files. Or the Enum stuff, see https://github.com/davidhalter/jedi/blob/47e2cf95d2a7d2d58fbba6fd2a78fda892071301/jedi/plugins/stdlib.py#L785-L816.

In Python 3.8, the TypedDict is a `_TypedDictMeta`, so that could work pretty well. The problem is probably that there's a stub in typeshed that takes preference over the actual `TypeDict` implementation. To work around this, you can probably just replace `TypedDict: object` with

```
class _TypedDictMeta(type): 
class TypedDict(dict, metaclass=_TypedDictMeta): ...
```
in  `jedi/third_party/typeshed/stdlib/3/typing.pyi`. This is probably also we can contribute to typeshed, because the current `object` definition is definitely less helpful than the `TypeDict` that inherits from `dict`. I'm not sure if that's good enough, because we might also need to add:

```
for c in self._wrapped_name.infer():
    yield c
```

to `jedi.inference.gradual.typing.TypingModuleName` (search for TypeDict), because the TypeDict might be completely ignored for now (for no reasons at all, but there's a TODO).

Also your class will probably look something like this:

```
[...]

PR #1495, opened 2020-02-07
TypedDict completions: tests added

pappasam:
Continues the conversation in this issue: https://github.com/davidhalter/jedi/issues/1478

Note: many tests are failing at present, but not the ones I've added. Notably, one of the tests for actually completing TypedDict keyword parameters that includes quotes is failing weirdly. This tells me there may be an improperly-handled edge case for key completion somewhere. @davidhalter do you have any ideas about where this could be coming from?

```traceback
case = <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:561 "d['fo']">, actual = {"'foo"}, desired = {"'foo'"}

    def assert_case_equal(case, actual, desired):
        """
        Assert ``actual == desired`` with formatted message.

        This is not needed for typical pytest use case, but as we need
        ``--assert=plain`` (see ../pytest.ini) to workaround some issue
        due to pytest magic, let's format the message by hand.
        """
    Test %r failed.
    actual  = %s
    desired = %s
    """ % (case, actual, desired)
E       AssertionError:
E         Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:561 "d['fo']"> failed.
E         actual  = {"'foo"}
E         desired = {"'foo'"}
E
E       assert {"'foo"} == {"'foo'"}
E         Extra items in the left set:
E         "'foo"
E         Extra items in the right set:
E         "'foo'"
```

### DIFF
--- a/test/completion/pep0484_typing.py
+++ b/test/completion/pep0484_typing.py
@@ -569,9 +569,7 @@ class Bar(Foo):
 
     #? int()
     another_variable
-    #? str()
-    foo
-    #? int()
+    #?
     an_int
 
 def typed_dict_test_foo(arg: Bar):
--- a/test/completion/pep0484_typing.py
+++ b/test/completion/pep0484_typing.py
@@ -510,7 +510,6 @@ class Foo(typing.TypedDict):
     foo: str
     bar: typing.List[float]
     an_int: int
-    foo
     #! ['foo: str']
     foo
     #? str()
@@ -525,8 +524,10 @@ def typed_dict_test_foo(arg: Foo):
 
     #? str()
     a_string
-    #? [float()]
-    a_list_of_strings
+    #? list()
+    a_list_of_floats
+    #? float()
+    a_list_of_floats[0]
     #? int()
     an_int
 
@@ -558,7 +559,41 @@ Foo.set
 #? ['setdefault']
 d.setdefaul
 
-#? 5 ["'foo'"]
+#? 5 ["'foo"]
 d['fo']
 #? 5 ['"bar"']
 d["bar"]
+
+class Bar(Foo):
+    another_variable: int
+
+    #? int()
+    another_variable
+    #? str()
+    foo
+    #? int()
+    an_int
+
+def typed_dict_test_foo(arg: Bar):
+    a_string = arg['foo']
+    a_list_of_floats = arg['bar']
+    an_int = arg['an_int']
+    another_variable = arg['another_variable']
+
+    #? str()
+    a_string
+    #? list()
+    a_list_of_floats
+    #? float()
+    a_list_of_floats[0]
+    #? int()
+    an_int
+    #? int()
+    another_variable
+
+    #? ['isupper']
+    a_string.isuppe
+    #? ['pop']
+    a_list_of_floats.po
+    #? ['as_integer_ratio']
+    an_int.as_integer_rati
--- a/test/completion/pep0484_typing.py
+++ b/test/completion/pep0484_typing.py
@@ -509,11 +509,33 @@ dynamic_annotation('')
 class Foo(typing.TypedDict):
     foo: str
     bar: typing.List[float]
+    an_int: int
     foo
     #! ['foo: str']
     foo
     #? str()
     foo
+    #? int()
+    an_int
+
+def typed_dict_test_foo(arg: Foo):
+    a_string = arg['foo']
+    a_list_of_floats = arg['bar']
+    an_int = arg['an_int']
+
+    #? str()
+    a_string
+    #? [float()]
+    a_list_of_strings
+    #? int()
+    an_int
+
+    #? ['isupper']
+    a_string.isuppe
+    #? ['pop']
+    a_list_of_floats.po
+    #? ['as_integer_ratio']
+    an_int.as_integer_rati
 
 #! ['class Foo']
 d: Foo
--- a/jedi/inference/gradual/typing.py
+++ b/jedi/inference/gradual/typing.py
@@ -5,8 +5,9 @@ values.
 
 This file deals with all the typing.py cases.
 """
+import itertools
 from jedi import debug
-from jedi.inference.compiled import builtin_from_name
+from jedi.inference.compiled import builtin_from_name, create_simple_object
 from jedi.inference.base_value import ValueSet, NO_VALUES, Value, \
     LazyValueWrapper
 from jedi.inference.lazy_value import LazyKnownValues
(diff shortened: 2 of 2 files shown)


test/completion/pep0484_typing.py
```diff
@@ -509,11 +509,33 @@ def dynamic_annotation(x: int):
...
+    an_int
+
+def typed_dict_test_foo(arg: Foo):
+    a_string = arg['foo']
+    a_list_of_floats = arg['bar']
+    an_int = arg['an_int']
+
+    #? str()
+    a_string
+    #? [float()]
```
davidhalter: This line is wrong syntax, you should write `#? float()` instead of `#? [float()]`. For multiple results you could just write `#? float() str()`. It's separated by spaces.

pappasam: By `[float()]`, I mean a "list of floats". I don't know how many floats will be present in the result. If `[float()]` is the wrong syntax, how would I express an expected "list of floats"?

pappasam:
Regarding changing the test to `"'foo"` from `"'foo'"`, this sounds simple enough, but I'm curious why the same doesn't need to be done for the test on line 564 with `'"bar"'`? Sounds like there's either a subtle bug we're letting slide, or that there's a meaningful difference between double and single quotes that I don't fully understand.

test/completion/pep0484_typing.py
davidhalter: It's not really possible to test this AFAIK with just one test. We could obviously make the test algorithm a bit smarter. However I usually test stuff like this with:

```
#? list()
a_list_of_strings
#? float()
a_list_of_strings[undefined_index]
```

BTW: I think you mean `a_list_of_floats`, so the name is wrong :)

davidhalter:
Regarding `foo`. Currently Jedi doesn't add the quote at the end of a string if the next character is a quote. For `'fo'` the next character is `'`. However for `'bar'` it's `r`.

pappasam:
Sadly, looks like integration tests are failing with inheritance. Here are the relevant test failures; hopefully you're able to see where and why things are going weirdly.

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:572 '    foo'> failed.
  actual  = {'test.completion.pep0484_typing:def foo'}
  desired = {'builtins:instance str()'}
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:574 '    an_int'> failed.
  actual  = set()
  desired = {'builtins:instance int()'}

assert set() == {'builtins:instance int()'}
  Extra items in the right set:
  'builtins:instance int()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:583 '    a_string'> failed.
  actual  = set()
  desired = {'builtins:instance str()'}

assert set() == {'builtins:instance str()'}
  Extra items in the right set:
  'builtins:instance str()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:585 '    a_list_of_floats'> failed.
  actual  = set()
  desired = {'builtins:instance list()'}

assert set() == {'builtins:instance list()'}
  Extra items in the right set:
  'builtins:instance list()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:587 '    a_list_of_floats[0]'> failed.
  actual  = set()
  desired = {'builtins:instance float()'}

assert set() == {'builtins:instance float()'}
  Extra items in the right set:
  'builtins:instance float()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:589 '    an_int'> failed.
  actual  = set()
  desired = {'builtins:instance int()'}

assert set() == {'builtins:instance int()'}
  Extra items in the right set:
  'builtins:instance int()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:591 '    another_variable'> failed.
  actual  = set()
  desired = {'builtins:instance int()'}

assert set() == {'builtins:instance int()'}
  Extra items in the right set:
  'builtins:instance int()'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:594 '    a_string.isuppe'> failed.
  actual  = set()
  desired = {'isupper'}

assert set() == {'isupper'}
  Extra items in the right set:
  'isupper'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:596 '    a_list_of_floats.po'> failed.
  actual  = set()
  desired = {'pop'}

assert set() == {'pop'}
  Extra items in the right set:
  'pop'
  Use -v to get the full diff
```

```text
AssertionError:
  Test <IntegrationTestCase: /home/sroeca/src/Personal/jedi/test/completion/pep0484_typing.py:598 '    an_int.as_integer_rati'> failed.
  actual  = set()
  desired = {'as_integer_ratio'}

assert set() == {'as_integer_ratio'}
  Extra items in the right set:
  'as_integer_ratio'
  Use -v to get the full diff
```

test/completion/pep0484_typing.py
```diff
@@ -536,7 +559,41 @@ class Foo(typing.TypedDict):
...
+
+class Bar(Foo):
+    another_variable: int
+
+    #? int()
+    another_variable
+    #? str()
+    foo
+    #? int()
+    an_int
```
davidhalter: `foo` and `an_int` should be `#?`. They don't infer anything, this is not a bug, Python would pretty much say the same thing.

pappasam: Ah, I see. I've changed the check for `an_int` to `#?` and removed the check for `foo` (it's apparently a function we've defined for other tests, so it has a weird type that's tied to other tests and I think it's prudent to avoid testing for that).

davidhalter:
I'll look into it.

davidhalter:
Merged to master in bb91b96286462299fc222298e36fc7ad6ddd357c

Fixed the issues in 609737322d55090243ffecf0c26c7225859bf83d. The issue was actually that the mro doesn't work anymore if the class is already not a class anymore. I realized that we could just do this when an instance of the TypedDict would be created. For now this works great.

Thanks again for your work on this one!

PR #1498, opened 2020-02-11
Add venv to .gitignore

Jma353:
I would like to contribute to Jedi, but I noticed that `venv`'s were not gitignored.  When installing all the `requirements.txt`'s, I want to isolate them to a `venv`.  Hence, this brief PR.


davidhalter:
Thanks! Hope to see other contributions :)

PR #1505, opened 2020-02-15
Add a couple of docstrings

PeterJCLaw:
This adds a couple of docstrings I wrote while first trying to understand the code, though a little tidied up. These are therefore based on observation of the outputs of these functions, so please do check that they're actually correct!


davidhalter:
Thanks a lot! I'm always a bit lazy and don't really write enough docstrings, so I really appreciate the effort!

It also doesn't help that I'm basically the only one that works on Jedi's core. So I generally just do what's "right" for me.

PR #1508, opened 2020-02-19
Remove the word 'class' from annotation_string

goldberglior:
Currently, 'foo(x: int)' results with annotation_string="<class 'int'>".
Change this to 'int'.


davidhalter:
Makes sense, thanks!

PR #1511, opened 2020-02-23
Improve generic type annotation inference

### LINKED ISSUE #1497: No inference for function with generic list argument and return
While `jedi` is able to correctly infer similar generic typed functions, it seems to have an issue with `List[T]` as a parameter or return type (or both). Changing the parameter and return type to another generic type, for example `Tuple[T]`, seems to fix this, though changing only one (for example to signature `(List[T]) -> Tuple[T]`) still demonstrates the issue.

I've put a bunch of examples in https://gist.github.com/PeterJCLaw/388aeb36c9c8c6d9db14c8407501bbf9, which what I was experimenting with, but the failing case is:

``` python
from typing import List, TypeVar

T = TypeVar('T')

def typed_list_generic_passthrough(x: List[T]) -> List[T]:
    return x

# the failure happens whether or not this list has a type annotation
list_str = ['abc', 'def']  # type: List[str]

for d in typed_list_generic_passthrough(list_str):
    #? str()
    d
```

PeterJCLaw:
This fixes #1497 by generalising the mechanism used for type annotation matching.

This is still somewhat work-in-progress, though I'm putting up a PR to get some feedback (and for a CI run).

Things outstanding:
 - [x] ~it would be good to reduce the amount of duplication in `_infer_type_vars`~
   Done.

 - [x] ~see if we can push some of the logic out to the types, rather than having it all in `_infer_type_vars`~
   Now that it's tidied up, there's not all that much extra logic here. This might still be possible, but there doesn't seem to be an obvious set of things to pull out.

 - [x] ~I've probably ended up duplicating a number of existing tests, maybe those should be rationalised~
    I've not looked into this in detail, though I doubt this is a big issue.

### DIFF
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -390,6 +390,8 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
         `annotated_argument_class`: represents the annotated class of the
             argument being passed to the object annotated by `annotation_value`.
         """
+        if not isinstance(annotated_argument_class, DefineGenericBase):
+            return
 
         annotation_generics = annotation_value.get_generics()
         actual_generics = annotated_argument_class.get_generics()
@@ -472,9 +474,6 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                     if not isinstance(py_class, GenericClass):
                         py_class = element
 
-                    if not isinstance(py_class, DefineGenericBase):
-                        continue
-
                     merge_pairwise_generics(annotation_value, py_class)
 
     elif isinstance(annotation_value, GenericClass):
@@ -507,11 +506,7 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                 for parent_class in py_class.py__mro__():
                     class_name = parent_class.py__name__()
                     if annotation_name == class_name:
-                        if not isinstance(parent_class, DefineGenericBase):
-                            continue
-
                         merge_pairwise_generics(annotation_value, parent_class)
-
                         break
 
     return type_var_dict
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -333,13 +333,14 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
     unpacks the `Iterable`.
     """
     type_var_dict = {}
+    annotation_name = annotation_value.py__name__()
+
     if isinstance(annotation_value, TypeVar):
         if not is_class_value:
-            return {annotation_value.py__name__(): value_set.py__class__()}
-        return {annotation_value.py__name__(): value_set}
+            return {annotation_name: value_set.py__class__()}
+        return {annotation_name: value_set}
     elif isinstance(annotation_value, TypingClassValueWithIndex):
-        name = annotation_value.py__name__()
-        if name == 'Type':
+        if annotation_name == 'Type':
             given = annotation_value.get_generics()
             if given:
                 for nested_annotation_value in given[0]:
@@ -351,7 +352,7 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                             is_class_value=True,
                         )
                     )
-        elif name == 'Callable':
+        elif annotation_name == 'Callable':
             given = annotation_value.get_generics()
             if len(given) == 2:
                 for nested_annotation_value in given[1]:
@@ -363,8 +364,7 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                         )
                     )
     elif isinstance(annotation_value, GenericClass):
-        name = annotation_value.py__name__()
-        if name == 'Iterable':
+        if annotation_name == 'Iterable':
             given = annotation_value.get_generics()
             if given:
                 for nested_annotation_value in given[0]:
@@ -375,7 +375,7 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                             value_set.merge_types_of_iterate(),
                         )
                     )
-        elif name == 'Mapping':
+        elif annotation_name == 'Mapping':
             given = annotation_value.get_generics()
             if len(given) == 2:
                 for value in value_set:
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -491,9 +491,6 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
             # the elements from the set first, then handle them, even if we put
             # them back in a set afterwards.
             for element in value_set:
-                if not hasattr(element, 'is_instance'):
-                    continue
-
                 if element.is_instance():
                     py_class = element.get_annotated_class_object()
                 else:
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -445,37 +445,37 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                         )
                     )
         elif annotation_name == 'Tuple':
-            # TODO: this logic is pretty similar to the general logic below, can
-            # we combine them?
+            annotation_generics = annotation_value.get_generics()
+            tuple_annotation, = annotation_value.execute_annotation()
+            # TODO: is can we avoid using this private method?
+            if tuple_annotation._is_homogenous():
+                # The parameter annotation is of the form `Tuple[T, ...]`,
+                # so we treat the incoming tuple like a iterable sequence
+                # rather than a positional container of elements.
+                for nested_annotation_value in annotation_generics[0]:
+                    _merge_type_var_dicts(
+                        type_var_dict,
+                        _infer_type_vars(
+                            nested_annotation_value,
+                            value_set.merge_types_of_iterate(),
+                        ),
+                    )
+            else:
+                # The parameter annotation has only explicit type parameters
+                # (e.g: `Tuple[T]`, `Tuple[T, U]`, `Tuple[T, U, V]`, etc.) so we
+                # treat the incoming values as needing to match the annotation
+                # exactly, just as we would for non-tuple annotations.
 
-            for element in value_set:
-                py_class = element.get_annotated_class_object()
-                if not isinstance(py_class, GenericClass):
-                    py_class = element
+                # TODO: this logic is pretty similar to the general logic below, can
+                # we combine them?
 
-                if not isinstance(py_class, DefineGenericBase):
-                    continue
+                for element in value_set:
+                    py_class = element.get_annotated_class_object()
+                    if not isinstance(py_class, GenericClass):
+                        py_class = element
 
-                annotation_generics = annotation_value.get_generics()
-                tuple_annotation, = annotation_value.execute_annotation()
-                # TODO: is can we avoid using this private method?
-                if tuple_annotation._is_homogenous():
-                    # The parameter annotation is of the form `Tuple[T, ...]`,
-                    # so we treat the incoming tuple like a iterable sequence
-                    # rather than a positional container of elements.
-                    for nested_annotation_value in annotation_generics[0]:
-                        _merge_type_var_dicts(
-                            type_var_dict,
-                            _infer_type_vars(
-                                nested_annotation_value,
-                                value_set.merge_types_of_iterate(),
-                            ),
-                        )
-                else:
-                    # The parameter annotation has only explicit type parameters
-                    # (e.g: `Tuple[T]`, `Tuple[T, U]`, `Tuple[T, U, V]`, etc.) so we
-                    # treat the incoming values as needing to match the annotation
-                    # exactly, just as we would for non-tuple annotations.
+                    if not isinstance(py_class, DefineGenericBase):
+                        continue
 
                     actual_generics = py_class.get_generics()
                     merge_pairwise_generics(annotation_generics, actual_generics)
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -429,8 +429,9 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                                 nested_annotation_value,
                                 value_set,
                                 is_class_value=True,
-                            )
+                            ),
                         )
+
         elif annotation_name == 'Callable':
             given = annotation_value.get_generics()
             if len(given) == 2:
@@ -440,8 +441,9 @@ def _infer_type_vars(annotation_value, value_set, is_class_value=False):
                         _infer_type_vars(
                             nested_annotation_value,
                             value_set.execute_annotation(),
-                        )
+                        ),
                     )
+
         elif annotation_name == 'Tuple':
             annotation_generics = annotation_value.get_generics()
             tuple_annotation, = annotation_value.execute_annotation()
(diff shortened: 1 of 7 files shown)


davidhalter:
My first impression is: This looks really good and it also looks like it was a lot of work. So thanks a lot!

I feel like this is in a very good shape and I'll try to review ASAP. It's just a bit late today :).

davidhalter:
I thought about this quite a bit more. Most of it is definitely in a good shape and works better than before. So I would even merge this. The only thing I don't like is the duplication (but this has been an issue before as well).

jedi/inference/gradual/annotation.py
```diff
@@ -333,25 +333,47 @@ def iter(iterable: Iterable[_T]) -> Iterator[_T]: ...
...
             given = annotation_value.get_generics()
             if given:
-                for nested_annotation_value in given[0]:
-                    _merge_type_var_dicts(
-                        type_var_dict,
-                        _infer_type_vars(
-                            nested_annotation_value,
-                            value_set,
-                            is_class_value=True,
+                if is_class_value:
```
davidhalter: Just wondering, why is needed? What's the example that makes this necessary?

PeterJCLaw: I don't recall exactly, though according to https://github.com/davidhalter/jedi/commit/c03ae0315e84585c55927fe470477bd7faa9869a, this is to make nested `Type[T]` cases work. I don't seem to have added a test in that commit, so likely there already was one somewhere in the tests I added (or elsewhere).

I'll check this and add some comments around what the branches do as part of tidying this up (hopefully later this week).

PeterJCLaw: I've added a general description of `is_class_value` to 54e29eede19947c27f63cfa7929e23d47327b135, does that help?

blueyed:
# [Codecov](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=h1) Report
> Merging [#1511](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=desc) into [master](https://codecov.io/gh/davidhalter/jedi/commit/1874e9be8132fec2ece54b12548ec551f0020abc&el=desc) will **decrease** coverage by `0.19%`.
> The diff coverage is `94.00%`.

[Impacted file tree graph](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master    #1511      +/-   ##
==========================================
- Coverage   93.56%   93.37%   -0.20%     
==========================================
  Files          77       77              
  Lines       10552    10637      +85     
==========================================
+ Hits         9873     9932      +59     
- Misses        679      705      +26     
```


| [Impacted Files](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [jedi/inference/gradual/typing.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvZ3JhZHVhbC90eXBpbmcucHk=) | `88.14% <0.00%> (-5.61%)` | :arrow_down: |
| [jedi/parser\_utils.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9wYXJzZXJfdXRpbHMucHk=) | `89.34% <0.00%> (-1.97%)` | :arrow_down: |
| [jedi/inference/recursion.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvcmVjdXJzaW9uLnB5) | `96.72% <0.00%> (-1.64%)` | :arrow_down: |
| [jedi/inference/value/klass.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvdmFsdWUva2xhc3MucHk=) | `96.89% <0.00%> (-1.37%)` | :arrow_down: |
| [jedi/inference/syntax\_tree.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2Uvc3ludGF4X3RyZWUucHk=) | `95.15% <0.00%> (-0.36%)` | :arrow_down: |
| [jedi/inference/value/iterable.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvdmFsdWUvaXRlcmFibGUucHk=) | `93.47% <0.00%> (-0.20%)` | :arrow_down: |
| [jedi/inference/base\_value.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvYmFzZV92YWx1ZS5weQ==) | `93.91% <0.00%> (ø)` | |
| [jedi/inference/compiled/getattr\_static.py](https://codecov.io/gh/davidhalter/jedi/pull/1511/diff?src=pr&el=tree#diff-amVkaS9pbmZlcmVuY2UvY29tcGlsZWQvZ2V0YXR0cl9zdGF0aWMucHk=) | `72.91% <0.00%> (+0.09%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=footer). Last update [1874e9b...d06efd0](https://codecov.io/gh/davidhalter/jedi/pull/1511?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

PeterJCLaw:
@davidhalter thanks for your feedback. I think I've addressed the duplication in `_infer_type_vars`. Please could you have another look?

jedi/inference/gradual/annotation.py
```diff
@@ -331,27 +331,107 @@ def iter(iterable: Iterable[_T]) -> Iterator[_T]: ...
...
+                    ),
+                )
+
     if isinstance(annotation_value, TypeVar):
         if not is_class_value:
-            return {annotation_value.py__name__(): value_set.py__class__()}
-        return {annotation_value.py__name__(): value_set}
+            return {annotation_name: value_set.py__class__()}
+        return {annotation_name: value_set}
     elif isinstance(annotation_value, TypingClassValueWithIndex):
```
davidhalter: I think we could just make this an if (instead of elif) and use

```
annotation_generics = annotation_value.get_generics()
```

just before that, because it's used in every if branch AFAIK (albeit with different names).

davidhalter:
While I have now looked at this code for quite a long time, I'm still not sure if there exists a way to move some of this stuff to its classes.

But anyway the code you wrote is great. I'm trying to think a bit more about this. If I don't come up with something, I think we can simply merge it. :)

jedi/inference/gradual/annotation.py
```diff
@@ -362,45 +442,73 @@ def iter(iterable: Iterable[_T]) -> Iterator[_T]: ...
...
-                    except AttributeError:
-                        continue
-                    key_values, value_values = method()
+        else:
+            # Note: we need to handle the MRO _in order_, so we need to extract
+            # the elements from the set first, then handle them, even if we put
+            # them back in a set afterwards.
+            for element in value_set:
+                if not hasattr(element, 'is_instance'):
+                    continue
```
davidhalter: Should not be necessary. If it's not available that's a bug.

davidhalter:
Now it took me a lot of time, but I feel like I figured out a better way to put this logic into the classes like this:

```
class BaseValue:
    def infer_type_vars(self, value_set, is_class_value=False):
        return {}


class ValueSet:
    def infer_type_vars(self, value_set, is_class_value=False):
        dct = {}
        for v in self._set:
            _merge_type_var_dicts(dct, v.infer_type_vars(value_set, is_class_value=is_class_value))
        return dct


class TypeVar():
    def infer_type_vars(self, value_set, is_class_value=False):
        if not is_class_value:
            return {self.py__name__(): value_set.py__class__()}
        return {self.py__name__(): value_set}

class TypingClassValueWithIndex():
    def infer_type_vars(self, is_class_value=False):
        annotation_name = self.py__name__()
        if annotation_name == 'Type':
            generics = self.get_generics()
            if generics:
                return given[0].infer_type_vars(value_set, is_class_value=True)
        elif annotation_name == 'Callable':
            generics = self.get_generics()
            if len(generics) == 2:
                return generics[1].infer_type_vars(value_set.execute_annotation())
        elif annotation_name == 'Tuple':
            generics = self.get_generics()
            tuple_annotation, = self.execute_annotation()
            # TODO: is can we avoid using this private method?
            if tuple_annotation._is_homogenous():
                # The parameter annotation is of the form `Tuple[T, ...]`,
                # so we treat the incoming tuple like a iterable sequence
                # rather than a positional container of elements.
                return generics[0].infer_type_vars(value_set.merge_types_of_iterate())
            else:
                # The parameter annotation has only explicit type parameters
                # (e.g: `Tuple[T]`, `Tuple[T, U]`, `Tuple[T, U, V]`, etc.) so we
                # treat the incoming values as needing to match the annotation
                # exactly, just as we would for non-tuple annotations.

                py_classes = ValueSet([e.get_annotated_class_object() for e in value_set])
                return merge_pairwise_generics(self, py_classes)
        return {}


class GenericClass():
    def infer_type_vars(self, is_class_value=False):
        if annotation_name == 'Iterable' and not is_class_value:
            generics = self.get_generics()
            if generics:
                return generics[0].infer_type_vars(value_set.merge_types_of_iterate())
        else:
            # Note: we need to handle the MRO _in order_, so we need to extract
            # the elements from the set first, then handle them, even if we put
            # them back in a set afterwards.
            for element in value_set:
                if element.is_instance():
                    py_class = element.get_annotated_class_object()
                else:
                    py_class = element

                for parent_class in py_class.py__mro__():
                    class_name = parent_class.py__name__()
                    if annotation_name == class_name:
                        return merge_pairwise_generics(self, parent_class)
        return {}


def merge_pairwise_generics(annotation_value, annotated_argument_class):
    """ XXX your docstring """
    if not isinstance(annotated_argument_class, DefineGenericBase):
        return {}

    annotation_generics = annotation_value.get_generics()
    actual_generics = annotated_argument_class.get_generics()

    type_var_dict = {}
    for annotation_generics_set, actual_generic_set in zip(annotation_generics, actual_generics):
        _merge_type_var_dicts(
            type_var_dict,
            annotation_generics_set.infer_type_vars(
                actual_generic_set,
                # This is a note to ourselves that we have already
                # converted the instance representation to its class.
                is_class_value=True,
            ),
        )
    return type_var_dict
```

I tried to incorporate all your ideas/code. What do you think? Is this a better way forward? I think some parts are quite a bit nicer, but now `merge_pairwise_generics` is used in multiple files. I haven't tested this code at all, it was just a lot of trying around and this finally came out.

jedi/inference/gradual/annotation.py
PeterJCLaw: Unfortunately calling `annotation_value.get_generics()` at the top level breaks in the case where the `annotation_value` isn't generic. (Currently that branch drops out the bottom with `type_var_dict` untouched)

Do you mean having `annotation_generics = annotation_value.get_generics()` just inside the `if` block here? That would probably work, though I'm not sure it gains much in the current spelling of things.

jedi/inference/gradual/annotation.py
PeterJCLaw: da9d3121 :thumbsup:

PeterJCLaw:
I think there might be something in a refactor like that. I'm having a play with that in another branch, though I think it's still going to be a bit more complex than your snippet suggests (there's a few branches missing, not sure if that's intentional). I suspect the ideal result also involves some changes to classes not yet touched on this branch, in order to align some of the representations of similar concepts.

davidhalter:
Thanks a lot for trying! :) Just let me know when you are stuck. This ticket has a high priority for me.

PeterJCLaw:
@davidhalter I've put up a separate PR with the refactors at https://github.com/PeterJCLaw/jedi/pull/1. Would be great to get your feedback on that. I've kept it separate just so it's clear which changes are functional versus refactor.

test/completion/pep0484_generic_parameters.py
```diff
@@ -223,6 +223,21 @@ class Specialised(Mapping[int, str]):
 values(specialised_instance)[0]
 
 
+# Test that classes which have gneeric ancestry but neither they nor their
```
davidhalter: typo :)

davidhalter:
Great couple of commits and tests. Looking forward to merging this :)

test/completion/pep0484_generic_parameters.py
PeterJCLaw: ta, 7e9ad9e7.

PeterJCLaw:
@davidhalter how do you want to do the merges here? I'd usually merge functional changes and refactors separately - do you want to merge this PR and then I'll open another one for the refactoring? Or would you rather I just put the refactoring in this PR as well?

davidhalter:
I pretty much don't care too much about the git history here :) So you can choose basically. Should I just merge both PRs?

PeterJCLaw:
@davidhalter sure, go for it :)

davidhalter:
I merged both branches and checked a few things again (for my understanding). It took me like 3 hours to finally understand why the `Type` stuff works that you made. I really like the end result, thanks a lot for contributing. Happy to see more of your pull requests in the future :)

If you want to keep working on Jedi stuff, there's basically one pull request I want to merge that was kind of abandoned. It's the Django PR #1467. If you want to work on it I would of course be grateful, because I'm currently working on avoiding nasty parso bugs. But I can of course understand if you don't have time or are not interested in Django :).

There's two commits I made that changed the behavior slightly (more or less avoids AttributeErrors):
61e9371849ce992573814de3ebb8d3ad1daf889c
ebb27867488a45c9fb21ff7081b118c8f057cc59

PeterJCLaw:
Nice catches :+1:, great to get this merged :)

Are there some docs somewhere on what "compiled" means in this context - I saw that while reading the source, but didn't feel I understood it and it didn't seem to actually come up in any of the test cases I created.

davidhalter:
Compiled objects are basically pure Python objects. This means something like `int` or `[1, 2]`, etc. The problem is in that case that `is_instance` serves multiple functions. If I had the motivation to change it I would probably make something like `def get_class(self) -> Optional[ClassValue]` in `BaseValue` to avoid checking like this.

Compiled objects were way more common before stubs. Before typeshed it was very normal to have compiled objects, e.g. `1` would be a compiled objects. Now it's something more complicated that doesn't actually get executed within a Python interpreter. So typically compiled objects are things like extensions: `PyQt`, `gtk`, etc. However you can also sometimes get compiled objects in Jedi for the standard library, e.g. when you access the docstring for `math.cos`.

PeterJCLaw:
Ok, thanks. I'm not sure I completely understand, though I probably need to have a bit of a play with the code.

I'm definitely interested in helping get Django support in (we use Django at work), though I have some other commitments over the next month or so, so I'm not sure I'll have time to pick that up soon.

PR #1512, opened 2020-02-23
Update sith's module docstring to match the available operations

PeterJCLaw:


sith.py
```diff
@@ -95,6 +94,7 @@ def from_cache(cls, record):
             args = json.load(f)
         return cls(*args)
 
+    # Changing this? Also update the module docstring above.
```
davidhalter: Why this comment?

PeterJCLaw: As a prompt to keep the docstring and the actual list of valid options in sync (hopefully preventing the need for PRs like this one in the future).

The alternative might be to move to using `docopt`s `(a|b|c)` spelling and remove the checking in this class, but the comment seemed less intrusive.

davidhalter:
Now I get it lol. Thanks!

davidhalter:
Why this broke master is an absolute mistery to me. But it probably has to do with the CI and not with you.

PR #1522, opened 2020-03-12
Add partialmethod, fixes #1519

### LINKED ISSUE #1519: No signature for functools partialmethod
`Script.get_signatures()` does not return function signatures for `functools.partialmethod`. Same issue was raised and fixed for `functools.partial` #1371

example:
```python
import jedi

partialmethod = '''
import functools

class X:
    def a(self, n, m):
        print(n, m)
    
    b = functools.partialmethod(a, 1)

X().a(1, 2)
X().b(2)
'''

script = jedi.Script(partialmethod)
print(script.get_signatures(line=10, column=len('X().a(')))
print(script.get_signatures(line=11, column=len('X().b(')))
```
output:
```
[<Signature: index=0 a(n, m)>]
[]
```

ffe4:
Returns correct method signature but test/completion/stdlib.py fails

I have added the test output below. The second kwarg test seems to pass, but the others don't. Since the first test fails because it gets what I assume is the value that's expected for the next test, I suspect that the issue is that I just increase the `arg_count` in L541.

While I'm not sure what method VSCode is using for getting the signature, using my fix I also get `self` included in the signature (a.k.a. parameter hints), which is not the case for normal methods.

```
> pytest test/test_integration.py -k stdlib
```
```
============================= test session starts ==============================
platform darwin -- Python 3.8.2, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /Users/Daniel/projects/ffe4/jedi, inifile: pytest.ini
collected 1841 items / 1755 deselected / 86 selected

test/test_integration.py ....................................FFF........ [ 54%]
.......................................                                  [100%]

=================================== FAILURES ===================================
_________________________ test_completion[stdlib:167] __________________________

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:167 "X().a('')[0]">
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x1072b7d90>
environment = <Environment: 3.8.2 in /Library/Frameworks/Python.framework/Versions/3.8>
has_typing = True

    def test_completion(case, monkeypatch, environment, has_typing):
        skip_reason = case.get_skip_reason(environment)
        if skip_reason is not None:
            pytest.skip(skip_reason)
    
        if 'pep0484_typing' in case.path and sys.version_info[0] == 2:
            pytest.skip('ditch python 2 finally')
    
        _CONTAINS_TYPING = ('pep0484_typing', 'pep0484_comments', 'pep0526_variables')
        if not has_typing and any(x in case.path for x in _CONTAINS_TYPING):
            pytest.skip('Needs the typing module installed to run this test.')
        repo_root = helpers.root_dir
        monkeypatch.chdir(os.path.join(repo_root, 'jedi'))
>       case.run(assert_case_equal, environment)

../test/test_integration.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../test/run.py:215: in run
    return testers[self.test_type](compare_cb, environment)
../test/run.py:257: in run_inference
    return compare_cb(self, is_str, should)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:167 "X().a('')[0]">
actual = {'builtins:instance str()'}, desired = {'builtins:instance int()'}

    def assert_case_equal(case, actual, desired):
        """
        Assert ``actual == desired`` with formatted message.
    
        This is not needed for typical pytest use case, but as we need
        ``--assert=plain`` (see ../pytest.ini) to workaround some issue
        due to pytest magic, let's format the message by hand.
        """
    Test %r failed.
    actual  = %s
    desired = %s
    """ % (case, actual, desired)
E       AssertionError: 
E         Test <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:167 "X().a('')[0]"> failed.
E         actual  = {'builtins:instance str()'}
E         desired = {'builtins:instance int()'}
E         
E       assert {'builtins:instance str()'} == {'builtins:instance int()'}
E         Extra items in the left set:
E         'builtins:instance str()'
E         Extra items in the right set:
E         'builtins:instance int()'

../test/test_integration.py:17: AssertionError
_________________________ test_completion[stdlib:169] __________________________

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:169 "X().a('')[1]">
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x1072d6ca0>
environment = <Environment: 3.8.2 in /Library/Frameworks/Python.framework/Versions/3.8>
has_typing = True

    def test_completion(case, monkeypatch, environment, has_typing):
        skip_reason = case.get_skip_reason(environment)
        if skip_reason is not None:
            pytest.skip(skip_reason)
    
        if 'pep0484_typing' in case.path and sys.version_info[0] == 2:
            pytest.skip('ditch python 2 finally')
    
        _CONTAINS_TYPING = ('pep0484_typing', 'pep0484_comments', 'pep0526_variables')
        if not has_typing and any(x in case.path for x in _CONTAINS_TYPING):
            pytest.skip('Needs the typing module installed to run this test.')
        repo_root = helpers.root_dir
        monkeypatch.chdir(os.path.join(repo_root, 'jedi'))
>       case.run(assert_case_equal, environment)

../test/test_integration.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../test/run.py:215: in run
    return testers[self.test_type](compare_cb, environment)
../test/run.py:257: in run_inference
    return compare_cb(self, is_str, should)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:169 "X().a('')[1]">
actual = set(), desired = {'builtins:instance str()'}

    def assert_case_equal(case, actual, desired):
        """
        Assert ``actual == desired`` with formatted message.
    
        This is not needed for typical pytest use case, but as we need
        ``--assert=plain`` (see ../pytest.ini) to workaround some issue
        due to pytest magic, let's format the message by hand.
        """
    Test %r failed.
    actual  = %s
    desired = %s
    """ % (case, actual, desired)
E       AssertionError: 
E         Test <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:169 "X().a('')[1]"> failed.
E         actual  = set()
E         desired = {'builtins:instance str()'}
E         
E       assert set() == {'builtins:instance str()'}
E         Extra items in the right set:
E         'builtins:instance str()'

../test/test_integration.py:17: AssertionError
_________________________ test_completion[stdlib:173] __________________________

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:173 'tup[0]'>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x106c8b070>
environment = <Environment: 3.8.2 in /Library/Frameworks/Python.framework/Versions/3.8>
has_typing = True

    def test_completion(case, monkeypatch, environment, has_typing):
        skip_reason = case.get_skip_reason(environment)
        if skip_reason is not None:
            pytest.skip(skip_reason)
    
        if 'pep0484_typing' in case.path and sys.version_info[0] == 2:
            pytest.skip('ditch python 2 finally')
    
        _CONTAINS_TYPING = ('pep0484_typing', 'pep0484_comments', 'pep0526_variables')
        if not has_typing and any(x in case.path for x in _CONTAINS_TYPING):
            pytest.skip('Needs the typing module installed to run this test.')
        repo_root = helpers.root_dir
        monkeypatch.chdir(os.path.join(repo_root, 'jedi'))
>       case.run(assert_case_equal, environment)

../test/test_integration.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../test/run.py:215: in run
    return testers[self.test_type](compare_cb, environment)
../test/run.py:257: in run_inference
    return compare_cb(self, is_str, should)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

case = <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:173 'tup[0]'>
actual = set(), desired = {'builtins:instance int()'}

    def assert_case_equal(case, actual, desired):
        """
        Assert ``actual == desired`` with formatted message.
    
        This is not needed for typical pytest use case, but as we need
        ``--assert=plain`` (see ../pytest.ini) to workaround some issue
        due to pytest magic, let's format the message by hand.
        """
    Test %r failed.
    actual  = %s
    desired = %s
    """ % (case, actual, desired)
E       AssertionError: 
E         Test <IntegrationTestCase: /Users/Daniel/projects/ffe4/jedi/test/completion/stdlib.py:173 'tup[0]'> failed.
E         actual  = set()
E         desired = {'builtins:instance int()'}
E         
E       assert set() == {'builtins:instance int()'}
E         Extra items in the right set:
E         'builtins:instance int()'

../test/test_integration.py:17: AssertionError
================ 3 failed, 83 passed, 1755 deselected in 5.95s =================
```

### DIFF
--- a/jedi/plugins/stdlib.py
+++ b/jedi/plugins/stdlib.py
@@ -513,13 +513,12 @@ class PartialObject(object):
         )
 
 
-class PartialMethodObject(object):
-    def __init__(self, actual_value, arguments):
+class PartialMethodObject(ValueWrapper):
+    def __init__(self, actual_value, arguments, instance=None):
+        super(PartialMethodObject, self).__init__(actual_value)
         self._actual_value = actual_value
         self._arguments = arguments
-
-    def __getattr__(self, name):
-        return getattr(self._actual_value, name)
+        self._instance = instance
 
     def _get_function(self, unpacked_arguments):
         key, lazy_value = next(unpacked_arguments, (None, None))
@@ -534,7 +533,9 @@ class PartialMethodObject(object):
         if func is None:
             return []
 
-        arg_count = 1
+        arg_count = 0
+        if self._instance is not None:
+            arg_count = 1
         keys = set()
         for key, _ in unpacked_arguments:
             if key is None:
@@ -544,7 +545,7 @@ class PartialMethodObject(object):
         return [PartialSignature(s, arg_count, keys) for s in func.get_signatures()]
 
     def py__get__(self, instance, class_value):
-        return ValueSet([self])
+        return ValueSet([PartialMethodObject(self._actual_value, self._arguments, self._instance)])
 
     def py__call__(self, arguments):
         func = self._get_function(self._arguments.unpack())
@@ -552,7 +553,7 @@ class PartialMethodObject(object):
             return NO_VALUES
 
         return func.execute(
-            MergedPartialArguments(self._arguments, arguments)
+            MergedPartialArguments(self._arguments, arguments, self._instance)
         )
 
 
@@ -568,15 +569,18 @@ class PartialSignature(SignatureWrapper):
 
 
 class MergedPartialArguments(AbstractArguments):
-    def __init__(self, partial_arguments, call_arguments):
+    def __init__(self, partial_arguments, call_arguments, instance=None):
         self._partial_arguments = partial_arguments
         self._call_arguments = call_arguments
+        self._instance = instance
 
     def unpack(self, funcdef=None):
         unpacked = self._partial_arguments.unpack(funcdef)
         # Ignore this one, it's the function. It was checked before that it's
         # there.
         next(unpacked)
+        if self._instance is not None:
+            yield None, LazyKnownValue(self._instance)
         for key_lazy_value in unpacked:
             yield key_lazy_value
         for key_lazy_value in self._call_arguments.unpack(funcdef):
@@ -592,7 +596,7 @@ def functools_partial(value, arguments, callback):
 
 def functools_partialmethod(value, arguments, callback):
     return ValueSet(
-        PartialMethodObject(instance, arguments)
+        PartialMethodObject(instance, arguments, instance) # FIXME pass correct instance as last arg
         for instance in value.py__call__(arguments)
     )
 
--- a/jedi/plugins/stdlib.py
+++ b/jedi/plugins/stdlib.py
@@ -513,6 +513,49 @@ class PartialObject(object):
         )
 
 
+class PartialMethodObject(object):
+    def __init__(self, actual_value, arguments):
+        self._actual_value = actual_value
+        self._arguments = arguments
+
+    def __getattr__(self, name):
+        return getattr(self._actual_value, name)
+
+    def _get_function(self, unpacked_arguments):
+        key, lazy_value = next(unpacked_arguments, (None, None))
+        if key is not None or lazy_value is None:
+            debug.warning("Partial should have a proper function %s", self._arguments)
+            return None
+        return lazy_value.infer()
+
+    def get_signatures(self):
+        unpacked_arguments = self._arguments.unpack()
+        func = self._get_function(unpacked_arguments)
+        if func is None:
+            return []
+
+        arg_count = 1
+        keys = set()
+        for key, _ in unpacked_arguments:
+            if key is None:
+                arg_count += 1
+            else:
+                keys.add(key)
+        return [PartialSignature(s, arg_count, keys) for s in func.get_signatures()]
+
+    def py__get__(self, instance, class_value):
+        return ValueSet([self])
+
+    def py__call__(self, arguments):
+        func = self._get_function(self._arguments.unpack())
+        if func is None:
+            return NO_VALUES
+
+        return func.execute(
+            MergedPartialArguments(self._arguments, arguments)
+        )
+
+
 class PartialSignature(SignatureWrapper):
     def __init__(self, wrapped_signature, skipped_arg_count, skipped_arg_set):
         super(PartialSignature, self).__init__(wrapped_signature)
@@ -543,6 +586,13 @@ class MergedPartialArguments(AbstractArguments):
 def functools_partial(value, arguments, callback):
     return ValueSet(
         PartialObject(instance, arguments)
+        for instance in value.py__call__(arguments)
+    )
+
+
+def functools_partialmethod(value, arguments, callback):
+    return ValueSet(
+        PartialMethodObject(instance, arguments)
         for instance in value.py__call__(arguments)
     )
 
@@ -744,6 +794,7 @@ _implemented = {
     },
     'functools': {
         'partial': functools_partial,
+        'partialmethod': functools_partialmethod,
         'wraps': _functools_wraps,
     },
     '_weakref': {
--- a/jedi/plugins/stdlib.py
+++ b/jedi/plugins/stdlib.py
@@ -473,49 +473,9 @@ def collections_namedtuple(value, arguments, callback):
     return ValueSet([ClassValue(inference_state, parent_context, generated_class)])
 
 
-class PartialObject(object):
-    def __init__(self, actual_value, arguments):
-        self._actual_value = actual_value
-        self._arguments = arguments
-
-    def __getattr__(self, name):
-        return getattr(self._actual_value, name)
-
-    def _get_function(self, unpacked_arguments):
-        key, lazy_value = next(unpacked_arguments, (None, None))
-        if key is not None or lazy_value is None:
-            debug.warning("Partial should have a proper function %s", self._arguments)
-            return None
-        return lazy_value.infer()
-
-    def get_signatures(self):
-        unpacked_arguments = self._arguments.unpack()
-        func = self._get_function(unpacked_arguments)
-        if func is None:
-            return []
-
-        arg_count = 0
-        keys = set()
-        for key, _ in unpacked_arguments:
-            if key is None:
-                arg_count += 1
-            else:
-                keys.add(key)
-        return [PartialSignature(s, arg_count, keys) for s in func.get_signatures()]
-
-    def py__call__(self, arguments):
-        func = self._get_function(self._arguments.unpack())
-        if func is None:
-            return NO_VALUES
-
-        return func.execute(
-            MergedPartialArguments(self._arguments, arguments)
-        )
-
-
-class PartialMethodObject(ValueWrapper):
+class PartialObject(ValueWrapper):
     def __init__(self, actual_value, arguments, instance=None):
-        super(PartialMethodObject, self).__init__(actual_value)
+        super(PartialObject, self).__init__(actual_value)
         self._actual_value = actual_value
         self._arguments = arguments
         self._instance = instance
@@ -545,7 +505,7 @@ class PartialMethodObject(ValueWrapper):
         return [PartialSignature(s, arg_count, keys) for s in func.get_signatures()]
 
     def py__get__(self, instance, class_value):
-        return ValueSet([PartialMethodObject(self._actual_value, self._arguments, self._instance)])
+        return ValueSet([PartialObject(self._actual_value, self._arguments, self._instance)])
 
     def py__call__(self, arguments):
         func = self._get_function(self._arguments.unpack())
@@ -596,7 +556,9 @@ def functools_partial(value, arguments, callback):
 
 def functools_partialmethod(value, arguments, callback):
     return ValueSet(
-        PartialMethodObject(instance, arguments, instance) # FIXME pass correct instance as last arg
+        # XXX last argument is a placeholder. See:
+        # https://github.com/davidhalter/jedi/pull/1522#discussion_r392474671
+        PartialObject(instance, arguments, True)
         for instance in value.py__call__(arguments)
     )
 
--- a/test/completion/stdlib.py
+++ b/test/completion/stdlib.py
@@ -158,6 +158,23 @@ tup[0]
 #? float()
 tup[1]
 
+class X:
+    def function(self, a, b):
+        return a, b
+    a = functools.partialmethod(function, 0)
+    kw = functools.partialmethod(function, b=1.0)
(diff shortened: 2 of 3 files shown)


test/test_inference/test_signature.py
```diff
@@ -82,6 +95,11 @@ def func(a, b, c):
         (partial_code + 'b(', 'func(b, c)'),
         (partial_code + 'c(', 'func(b)'),
         (partial_code + 'd(', None),
+
+        (partialmethod_code + 'X().a(', 'func(a, b, c)'),
+        (partialmethod_code + 'X().b(', 'func(b, c)'),
+        (partialmethod_code + 'X().c(', 'func(b)'),
+        (partialmethod_code + 'X().d(', None),
```
davidhalter: Good tests! Please also add a test for `X.c(` and `X.d`, just to make sure that stuff doesn't raise exceptions. I'm not really interested that it behaves correctly, it's more about avoiding exceptions.

davidhalter: Good tests! Please also add a test for `X.c(` and `X.d`, just to make sure that stuff doesn't raise exceptions. I'm not really interested that it behaves correctly, it's more about avoiding exceptions.

test/completion/stdlib.py
```diff
@@ -158,6 +158,23 @@ def function(a, b):
...
+class X:
+    def function(self, a, b):
+        return a, b
+    a = functools.partialmethod(function, 0)
+    kw = functools.partialmethod(function, b=1.0)
+
+#? int()
+X().a('')[0]
+#? str()
+X().a('')[1]
```
davidhalter: Please test `X.a('')` here as well. Also it would make sense to test `X.a(X(), '')` just to make sure that there are no exceptions (people are probably not going to use it like that. Same for kw below.

jedi/plugins/stdlib.py
```diff
@@ -513,6 +513,49 @@ def py__call__(self, arguments):
         )
 
 
+class PartialMethodObject(object):
+    def __init__(self, actual_value, arguments):
+        self._actual_value = actual_value
+        self._arguments = arguments
+
+    def __getattr__(self, name):
+        return getattr(self._actual_value, name)
```
davidhalter: You can just remove that and use `class PartialMethodObject(ValueWrapper)` (see also for example `ClassMethodGet`).

It's really my bad that PartialObject is not a proper Wrapper, yet.

davidhalter:
I think the errors are happening for a very simple reason: `MergedPartialArguments` doesn't know anything about the calling instance. I would propose to do the following:

```
    def __init__(self, actual_value, arguments, instance=None):
        self._actual_value = actual_value
        self._arguments = arguments
        self._instance = instance

    def py__get__(self, instance, class_value):
        return ValueSet([PartialMethodObject(self._actual_value, self._arguments, self._instance)])
```

Obviously `MergedPartialArguments` also needs to know about instance and can then do the following:

```
    def unpack(self, funcdef=None):                                          
        unpacked = self._partial_arguments.unpack(funcdef)                   
        # Ignore this one, it's the function. It was checked before that it's
        # there.                                                             
        next(unpacked)                          
        if self._instance is not None:
            yield None, LazyKnownValue(self._instance)   
        for key_lazy_value in unpacked:                                      
            yield key_lazy_value                                             
        for key_lazy_value in self._call_arguments.unpack(funcdef):          
            yield key_lazy_value                                             
```

I feel like that would fix it. You can then probably also change it for `arg_count = 1` like:

```
if self._instance is not None:
    arg_count = 1
```

I feel like this makes quite a few things possible AND you can probably reuse the same class completely for partial and partialmethod. So happy to further assist you, but if you don't have time I'm also fine doing it.

test/completion/stdlib.py
ffe4: I think `X.a('')` should throw an exception, so if I should test for that I don't know what comment I should place above. I've only added the `X.a(X(), '')` cases for now.

jedi/plugins/stdlib.py
```diff
@@ -592,7 +596,7 @@ def functools_partial(value, arguments, callback):
 
 def functools_partialmethod(value, arguments, callback):
     return ValueSet(
-        PartialMethodObject(instance, arguments)
+        PartialMethodObject(instance, arguments, instance) # FIXME pass correct instance as last arg
```
ffe4: I don't think this is the instance I should be passing, and the last argument really only acts like a boolean at the moment.

ffe4:
This version passes the old tests but not the new cases with the instance passes as an argument `X.a(X(), '')`. I don't think I'm passing the correct instance in L599, and I was hoping you could just tell me how to get it.

I can now replace PartialObject with PartialMethodObject and still pass the two tests I've been running, don't know if there are others. If just passing the correct instance makes the tests pass, I thought I'd replace the old PartialObject for the final commit.

test/test_inference/test_signature.py
ffe4: I've added the following:
```python
(partialmethod_code + 'X.c(', 'func(b)'),
(partialmethod_code + 'X.d(', None),
```
Although I think `X.c(` should technically include `self` in the signature

test/completion/stdlib.py
davidhalter: > I think X.a('') should throw an exception

No it shouldn't :) Nothing should ever throw exceptions in Jedi. That case is just there to actually avoid exceptions, I don't really care about the result as long as it doesn't raise exceptions. Obviously it needs to pass tests, but as long as that works it's fine I guess.

test/test_inference/test_signature.py
davidhalter: Doesn't really matter, nobody is going to use it like that anyway :) That might start working if we get the instance stuff working.

jedi/plugins/stdlib.py
davidhalter: Just pass nothing :) It should just be `PartialMethodObject(instance, arguments)`. AFAIK that should do it.

test/completion/stdlib.py
ffe4: Yes, you're right, I was thinking of something else. Added the tests now, and they are fine.

jedi/plugins/stdlib.py
ffe4: If I am going to replace the original `PartialObject` I'll have to at least pass a bool to distinguish between the two behaviours. Should I refactor the code with a `is_bool` argument and replace the `is None` checks?

ffe4: Your [previous comment](https://github.com/davidhalter/jedi/pull/1522#issuecomment-598467833) suggested creating an instance parameter in `PartialMethodObject` and passing it to `MergedPartialArguments`, since it had to know about the calling instance. So I assumed we'd have to pass something in there.

davidhalter: Yeah, I might not have been clear enough. The moment where the instance is defined is when `py__get__` is called.

ffe4: That might actually explain some of the issues I had while trying to understand the code and get it to work. Still, since I was getting more test failures when omitting the argument I assumed that would be the right track, though.

But I think the test failures might actually be more consistent when I do not pass an instance. The signature tests all start failing, however.

I'll submit another commit with the other changes (refactor, tests), and pass a bool as a placeholder arg for now since that way at least the tests pass. I feel my usefulness on this issue is exhausted, so it probably makes more sense to make this a proper PR and let you figure out the hard parts :)

jedi/plugins/stdlib.py
```diff
@@ -503,13 +504,16 @@ def get_signatures(self):
                 keys.add(key)
         return [PartialSignature(s, arg_count, keys) for s in func.get_signatures()]
 
+    def py__get__(self, instance, class_value):
+        return ValueSet([PartialObject(self._actual_value, self._arguments, self._instance)])
```
davidhalter: The issue here is that you use `self._instance` instead of instance. This is also why you had to add the random boolean.

I'm working on it now locally, so please don't push further changes.

davidhalter:
Thanks again for your work on this. The only thing I really needed to change was the bug with instance passing I mentioned above.

The rest is just a few additional tests and now stuff like `X.a('')` now return slightly different results, which doesn't really matter.

ffe4:
Ah, yes, that was probably part of my earlier confusion regarding the instance passing.

And thanks for your help and patience. This was my first time trying to contribute to a project, and I feel it might not have been the easiest subject matter to start off with, but I definitely got something out of it for the future.

davidhalter:
> And thanks for your help and patience. This was my first time trying to contribute to a project, and I feel it might not have been the easiest subject matter to start off with, but I definitely got something out of it for the future.

Yeah, Jedi is pretty hard to contribute to. There's very few people that even try :). Rest assured that most projects are easier to understand. Jedi is quite big and complex, because understanding Python is just hard.

Apart from that, you did great! Really, I mean it. You tried a lot of things yourself and I would be happy to see you contributing again :).

PR #1526, opened 2020-03-15
Minor fix for doc string of BaseDefinition.type

cc-a:
Included "statement" as a valid return type for BaseDefinition.type


jedi/api/classes.py
```diff
@@ -154,9 +154,9 @@ def type(self):
         'instance'
         >>> defs[3]
         'function'
-
+n
```
davidhalter: I don't suppose you wanted this `n` :)

Otherwise, thanks!

cc-a:
\*ahem\*, what `n`?

davidhalter:
Thanks!

PR #1537, opened 2020-04-02
Replace the deprecated function "params" with "get_signatures()[...].params"

gorogoroumaru:
The function "get_signature_param_names" in jedi/jedi/api/completion.py uses the deprecated function "params" in BaseName class.
So, I replaced it with "get_signatures()[...].params" in BaseName class.

### DIFF
--- a/jedi/api/completion.py
+++ b/jedi/api/completion.py
@@ -33,7 +33,7 @@ class ParamNameWithEquals(ParamNameWrapper):
 def get_signature_param_names(signatures):
     # add named params
     for call_sig in signatures:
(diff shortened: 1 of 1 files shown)


davidhalter:
Hey @gorogoroumaru Thanks for trying to contribute :) I don't think that you saw that the params function is not deprecated in `CallSignature`, it's just deprecated in `BaseName`. Do you agree?

davidhalter:
?

gorogoroumaru:
I agree with you.
Sorry for late reply.

PR #1546, opened 2020-04-15
Use explicit environment on subprocess

### LINKED ISSUE #1540: Jedi has wrong sys.path when started from pyenv virtual environment
### The Problem
When working from a pyenv virtual environment and attempting to get completions for a conda envrionment, jedi does not establish the correct sys.path, resulting in failed completion. Working from an anaconda environment and attemping to get completions for a pyenv environment works properly.

### MWE summary
* create two virtual environments
  * Anaconda environment with `jedi` and `scipy`
  * Pyenv envrionment with `jedi` and `pandas`
* Use jedi in one environment and get completion from the other environment

### Example 1: use conda; completions from pyenv; successful
```
$ conda create -n conda-jedi python=3.7.7 jedi scipy
$ conda activate conda-jedi
(conda-jedi) $ python
Python 3.7.7 (default, Mar 26 2020, 10:32:53) 
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
>>> import jedi
>>> env = jedi.create_environment('/Users/rclary/.pyenv/versions/pyenv-jedi/bin/python', safe=False)
>>> env
<Environment: 3.7.7 in /Users/rclary/.pyenv/versions/3.7.7/envs/pyenv-jedi>
>>> env.get_sys_path()
['/usr/local/mdsplus/pydevices', '/Users/rclary/.pyenv/versions/3.7.7/Python.framework/Versions/3.7/lib/python37.zip', '/Users/rclary/.pyenv/versions/3.7.7/Python.framework/Versions/3.7/lib/python3.7', '/Users/rclary/.pyenv/versions/3.7.7/Python.framework/Versions/3.7/lib/python3.7/lib-dynload', '/Users/rclary/.pyenv/versions/3.7.7/envs/pyenv-jedi/lib/python3.7/site-packages']
>>> jedi.Script(r"""import pa\n""", line=1, column=len('import pa'), environment=env).completions()
[<Completion: pandas>, <Completion: parser>, <Completion: parso>, <Completion: pathlib>]
>>> 
```

### Example 2: use pyenv; completions in conda; unsuccessful
```
$ pyenv virtualenv 3.7.7 pyenv-jedi
$ pyenv activate pyenv-jedi
(pyenv-jedi) $ pip install jedi pandas
(pyenv-jedi) $ python
Python 3.7.7 (default, Apr  6 2020, 11:49:38) 
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
>>> import jedi
>>> env = jedi.create_environment('/Users/rclary/anaconda3/envs/conda-jedi/bin/python', safe=False)
>>> env
<Environment: 3.7.7 in /Users/rclary/.pyenv/versions/3.7.7/envs/pyenv-jedi>
>>> env.get_sys_path()
['/usr/local/mdsplus/pydevices', '/Users/rclary/anaconda3/envs/conda-jedi/lib/python37.zip', '/Users/rclary/anaconda3/envs/conda-jedi/lib/python3.7', '/Users/rclary/anaconda3/envs/conda-jedi/lib/python3.7/lib-dynload', '/Users/rclary/.pyenv/versions/3.7.7/envs/pyenv-jedi/lib/python3.7/site-packages']
>>> jedi.Script(r"""import sc\n""", line=1, column=len('import sc'), environment=env).completions()
[<Completion: sched>, <Completion: scripts>]
>>> 
```

Notice that for the unsuccessful case, the environment description incorrectly indicates the pyenv-jedi environment and the `get_sys_path()` returns the wrong site-package path. The result is missing `<Completions: scipy>`

mrclary:
Resolves #1540.

Use an explicit environment for subprocess to ensure that existing environment variables are not inherited. This ensures more reliable results.

### DIFF
--- a/jedi/inference/compiled/subprocess/__init__.py
+++ b/jedi/inference/compiled/subprocess/__init__.py
@@ -181,6 +181,15 @@ class CompiledSubprocess(object):
             os.path.dirname(os.path.dirname(parso_path)),
             '.'.join(str(x) for x in sys.version_info[:3]),
         )
+        # Use explicit envionment to ensure reliable results (#1540)
+        env = {}
+        if os.name == 'nt':
+            # if SYSTEMROOT (or case variant) exists in environment,
+            # ensure it goes to subprocess
+            for k, v in os.environ.items():
+                if 'SYSTEMROOT' == k.upper():
+                    env.update({k: os.environ[k]})
+                    break  # don't risk multiple entries
         process = GeneralizedPopen(
             args,
             stdin=subprocess.PIPE,
@@ -188,7 +197,8 @@ class CompiledSubprocess(object):
             stderr=subprocess.PIPE,
             # Use system default buffering on Python 2 to improve performance
             # (this is already the case on Python 3).
-            bufsize=-1
+            bufsize=-1,
+            env=env
         )
         self._stderr_queue = Queue()
         self._stderr_thread = t = Thread(
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -56,6 +56,7 @@ Code Contributors
 - Shane Steinert-Threlkeld (@shanest) <ssshanest@gmail.com>
 - Tim Gates (@timgates42) <tim.gates@iress.com>
 - Lior Goldberg (@goldberglior)
+- Ryan Clary (@mrclary)
 
 And a few more "anonymous" contributors.
 


mrclary:
Tested locally on my system, under both conda and pyenv environments, `TestSetupReadline.test_import` has one failure due to `assert len(difference) < 22` where `len(difference)` is 47 for the conda environment and 45 for th pyenv environment. Both environments have only `parso==0.7.0` and `pytest` explicitly installed (Python 3.7.7, MacOS 10.14.6).

I'm not sure what purpose this assertion serves, since you've increased it recently from 21...
Nevertheless, all other tests pass locally for me.

Note that I do not know what effects there will be on Windows systems. In the [subproccess.Popen](https://docs.python.org/3/library/subprocess.html#popen-constructor) documentation, I found:

> Note: If specified, env must provide any variables required for the program to execute. On Windows, in order to run a side-by-side assembly the specified env must include a valid SystemRoot.

davidhalter:
> Note that I do not know what effects there will be on Windows systems

That's fine I guess, the Windows CI passes. If it doesn't work for people, I guess I will hear from them :).

> Tested locally on my system, under both conda and pyenv environments, TestSetupReadline.test_import has one failure due to assert len(difference) < 22 where len(difference) is 47 for the conda environment and 45 for th pyenv environment. Both environments have only parso==0.7.0 and pytest explicitly installed (Python 3.7.7, MacOS 10.14.6).

This is test is there to ensure that the actual completions for os are not too different from typeshed's definitions. It would probably need a bit of love to make it pass on virtualenvs as well. (Just raising the numbers is not really what I want).

jedi/inference/compiled/subprocess/__init__.py
```diff
@@ -181,14 +181,20 @@ def _get_process(self):
             os.path.dirname(os.path.dirname(parso_path)),
             '.'.join(str(x) for x in sys.version_info[:3]),
         )
+        # Use explicit envionment to ensure reliable results (#1540)
+        env = {}
+        if os.name == 'nt':
+            # Windows requires SYSTEMROOT
+            env.update(SYSTEMROOT=os.environ.get('SYSTEMROOT'))
```
davidhalter, line 188: Minor issue: While SYSTEMROOT is probably defined most of the time, if it is not we should just not set it. The problem is essentially that it could be empty. If it is we cannot set it:
```
>>> os.environ['asdf'] = None
  File "/usr/lib/python3.5/os.py", line 730, in __setitem__
    value = self.encodevalue(value)
  File "/usr/lib/python3.5/os.py", line 798, in encode
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not NoneType
```

mrclary: Another issue: Windows environment variables are case insensitive, so it also be `SYSTEMROOT`, `SystemRoot`, or maybe `systemroot`. I'll put in some robustness and if none are present, I'll refrain from setting.

davidhalter:
Thanks again for you work here and the good reasoning :).

PR #1549, opened 2020-04-17
fix #1548

### LINKED ISSUE #1548: Use the `project` variable before setting the default value.
This function use `project` variable before setting the default value causing an `AttributeError` in jedi v0.17.01 when the user is using the old api.

For example, this code will raise an `AttributeError` with jedi v0.17.0:
https://github.com/neoclide/coc-python/blob/6a5104d18ae5ebf7d9b84af7c4a34c6ced62aee8/pythonFiles/completion.py#L571-L574

https://github.com/davidhalter/jedi/blob/8aaa8e00446f801c023bb4b056f51ac6f3d99e99/jedi/api/__init__.py#L131-L133

haoqixu:
Reorder the code to use `project` after setting the default value.

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -128,15 +128,6 @@ class Script(object):
         # An empty path (also empty string) should always result in no path.
         self.path = os.path.abspath(path) if path else None
 
-        # TODO deprecate and remove sys_path from the Script API.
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks!

PR #1553, opened 2020-04-18
Fix construction of nested generic tuple return types

PeterJCLaw:
Previously the construction of some return types missed populating the generic parameters when they were known.

Notably this overlooked tuples and callables, leading to incomplete information for signatures like the following:

``` python
def func(the_list: List[T]) -> List[Tuple[T]]:
    return [(x,) for x in the_list]

def list_func_t_passthrough(the_list: List[Callable[..., T]]) -> List[Callable[..., T]]:
    return the_list
```

This PR fixes both of these.

Fixing this also showed up a an issue where the concrete type parameter of the return value within a callable in the argument to the function currently being evaluated was not always detected correctly:

``` python
list_of_funcs_2 = [bar]  # type: List[Callable[..., int]]

for b in list_func_t_passthrough(list_of_funcs_2):
    #? int()
    b(None, x="x") 
```

This PR also fixes this issue.

### DIFF
--- a/jedi/inference/gradual/base.py
+++ b/jedi/inference/gradual/base.py
@@ -99,7 +99,7 @@ class DefineGenericBase(LazyValueWrapper):
         for generic_set in self.get_generics():
             values = NO_VALUES
             for generic in generic_set:
-                if isinstance(generic, (GenericClass, TypeVar)):
+                if isinstance(generic, (DefineGenericBase, TypeVar)):
                     result = generic.define_generics(type_var_dict)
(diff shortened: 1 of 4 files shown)


PeterJCLaw:
Hrm, looks like the other issue this uncovers is a little more complex than I thought. I might see if I can fix that separately.

PeterJCLaw:
Turns out the other issue is actually related.

davidhalter:
Thanks for working on this!

PR #1554, opened 2020-04-19
Fix handling of nested tuple arguments

PeterJCLaw:
Discovered while exploring the issues in #1553, we previously weren't correctly handling nested generic tuples in arguments:

```python
list_of_int_tuples = [(x,) for x in list_of_ints]  # type: List[Tuple[int]]

for b in list_tuple_t_to_tuple_list_t(list_of_int_tuples):
    #? int()
    b[0]
```

This PR fixes that.


jedi/inference/gradual/typing.py
```diff
@@ -327,6 +327,9 @@ def infer_type_vars(self, value_set, is_class_value=False):
         from jedi.inference.gradual.annotation import merge_pairwise_generics, merge_type_var_dicts
         from jedi.inference.gradual.base import GenericClass
 
+        if is_class_value:
+            value_set = value_set.execute_annotation()
```
PeterJCLaw, line 331: I'm not actually sure I understand what `execute_annotation` is meant to do -- it seems to both be a way to get the return type from `Callable`s _and_ a way to do something very similar to `get_annotated_class_object` where that mechanism isn't supported.
As a result, I'm not completely sure that this is the right fix here, however this it does work.

davidhalter: `execute_annotation` is basically for this:

```
def foo() -> int:
    pass
```

`int` here is an annotation and needs to be executed to have the right value in Jedi. This is what the function does, it instantiates things. You could also do that with `execute(arguments)`, but there you would need to provide arguments (which is not what we are able to provide in annotations).

davidhalter: However I kind of understand why you are doing this. At the same time I feel like it kind of shows that there are some limits to our infer type vars logic. I feel like inferring type vars should probably work with classes, but I might also be wrong. It's just that this looks a bit like an issue with the basic design.

I clearly do not know how to fix this. Maybe `is_class_value` should always be `False`. I think an iterative approach here makes sense. At some point there is going to be a pattern and we can refactor.

PeterJCLaw: Yeah, I'd mostly assumed that `execute_annotation` was intended for getting return types as you suggest, yet if that's the intent then I'm a little confused by the implementation in [`TypingValueWithIndex`](https://github.com/davidhalter/jedi/blob/912fe68069e163d8f05262e59464dd65faea2e11/jedi/inference/gradual/typing.py#L103) which handles some type constructs rather differently (the mapping at the bottom).

In the case of this code, it's the behaviour there for `Tuple` which I'm trying to access. Should that logic be moved out, perhaps to a `get_annotated_class_object` implementation? (If so, then the usage of `execute_annotation` [within the `Tuple` branch of `TypingClassValueWithIndex.infer_type_vars`](https://github.com/davidhalter/jedi/blob/912fe68069e163d8f05262e59464dd65faea2e11/jedi/inference/gradual/typing.py#L225) should probably also change.

PeterJCLaw: I've pulled out the logic into a `get_annotated_class_object` in ce1ac38c (with subsequent follow-ups to use that).

davidhalter: > which handles some type constructs rather differently (the mapping at the bottom).

This is totally ok IMO. `Tuple` really is a tuple instance and not a class. The same is true for all the others. `execute_annotation` actually does what it says, it makes `Tuple[int]` to a `Tuple` instance. The naming is probably a bit unfortunate. Should we call it `TupleInstance` instead?

davidhalter: BTW just to make it clear: `get_annotated_class_object` returns **always** a class, while `execute_annotation` returns an instance - or generally executes (which is kind of a feature, but we could also disable that). I think that's the big difference. 

There are a couple of issues with the names, e.g. `get_annotated_class_object` should probably be called `get_annotated_class` and `Tuple` should probably be called `TupleInstance` to make clearer what they actually are. But that's where the issues stem from.

PeterJCLaw: Ah, I think I'm maybe understanding this a bit better now.

Is it the case that Jedi has (semi-)separate hierarchies to represent:
- instance types inferred from the code -- e.g: from `foo = [4]` jedi knows that `foo` is an instance of `list` containing an `int`
- types inferred from annotations -- e.g: from `foo: List[int]` jedi knows that `foo` has type `List[int]`
- instantiated types from annotations -- e.g: from `foo: List[int]` jedi can "execute" the annotation to know that `foo` will be an instance of `list` containing some `int`s

From what I recall, for `list`, these end up using a combination of `_GenericInstanceWrapper` and `TreeInstance` around `<GenericClass: <ClassValue: <Class: list@896-946>>[S{...}]>`.

However for tuples (and other things which use the `TypingClassValueWithIndex`), they specialise their "instance" forms using the `Tuple` (maybe now `TupleInstance`) classes rather than using a `GenericClass` around their `ClassValue` representation.

I guess the bit I'm missing is the relationship between the two instance forms of tuples -- the `ClassValue` form and the `Tuple` form. I can see that for the other types which `TypingClassValueWithIndex` handles it's (mostly) because they don't have a natural form (e.g: `Union`s don't exist at runtime, only as annotations), though tuples do. Is it because jedi needs a way to represent whether or not the tuple is homogeneous?

Anyway, I'm happy to back out the changes I made to use a `get_annotated_class_object` method -- now I understand a bit more I can see why that doesn't make sense.

jedi/inference/gradual/typing.py
```diff
@@ -136,6 +136,15 @@ def execute_annotation(self):
     def gather_annotation_classes(self):
         return ValueSet.from_sets(self._generics_manager.to_tuple())
 
+    def get_annotated_class_object(self):
+        if self._tree_name.value == 'Tuple':
+            return Tuple(
+                self.parent_context,
+                self._tree_name,
+                generics_manager=self._generics_manager,
+            )
```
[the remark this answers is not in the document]
PeterJCLaw: I'd misunderstood the relations between the various classes here. I'm going to back this out.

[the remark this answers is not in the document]
PeterJCLaw: 55facaaf3d36c1e896f721645a255461c63ba396 backs this out.

jedi/inference/gradual/typing.py
davidhalter: You are definitely on the right track! It's not quite easy to understand, because there's a lot of different values.

Like you mention, there's indeed a different concept for `[1]` and `List[int]`. The latter is obviously a `class TypeAlias(LazyValueWrapper)`, which is itself a wrapper around the stub `builtins.list[int]` (which is I think a `GenericClass`, but I might be wrong). The actual object will probably be a `TreeInstance()` that takes the class as its argument. If you define something with literals there's a `class SequenceLiteralValue()` that wraps basically around the `TreeInstance` that I just described.

However note that this is all done by composition, not inheritance and it's lazy. This makes sense, especially historically (I might not do it exactly the same way if I would implement Jedi from the ground up). The lazyness was put there to avoid calculating the generics when it's not necessary. In Jedi `[1, 2, 3, 4, 5][3]` actually returns an `int` literal `4` without any generics calculation. `Sequence._get_wrapped_value` is never called, because we can just call `SequenceLiteralValue().py__simple_getitem__(4)`. However if we want to access `[1, 2, 3, 4, 5].pop(3)` it will return an `int` instance. This happens, because in that case the generics of the list literal is calculated and from there we just use typeshed to get the correct types.

So kind of a TL;DR; `List[int]` is represented as something like `TreeInstance(GenericClass(ClassValue(builtins.list), Generics<int>))`. `[1]` is represented as `
SequenceLiteralValue([1])`, which kind of lazily resolves to `SequenceLiteralValue(TreeInstance(GenericClass(ClassValue(builtins.list), Generics<int>))`. Yeah, this stuff is pretty crazy haha :).

I think this still kind of works like intended. BUT for tuples this is kind of an issue...

> I guess the bit I'm missing is the relationship between the two instance forms of tuples -- the ClassValue form and the Tuple form. [...]
> Is it because jedi needs a way to represent whether or not the tuple is homogeneous?

I just never thought about this case really. You make a very very good observation. The end result is that `tuple` and `Tuple` have different tree nodes. `Tuple` is a very special case in typeshed, because unlike `List`, `Dict`, `Set`, etc. It's **not** a `TypeAlias`, it's its own class.

Reading your stuff and thinking about it for a bit, we should probably do one of the following:
- Make the `Tuple().tree_node` use `tuple` instead of `Tuple`.
- Make `Sequence()._get_wrapped_value` somehow return a `Tuple`. However this would probably lead to strange cases where goto would land on `Tuple` instead of `tuple`. I'm not really sure how to do this exactly, because Tuple is kind of a weird wrapper around `tuple()` and itself not very flexible, but we could probably refactor that.
- Combine those two approaches and use tree_node `builtins.tuple()` and use a modified version of `Tuple` so we have a common ancestor in our composition chain.

I think the first proposal is probably the best and easiest (I'm not sure the other ones are even feasible). It would probably look like this:

```
class Tuple(BaseTypingValueWithGenerics):
    [...]
    @property                 
    def tree_node(self):      
        return self._wrapped_value.tree_node
```

I don't really understand why this is not already the default behavior...

davidhalter: I think I understand a bit better now. I didn't really understand what you meant with `TypingClassValueWithIndex`. I think I understand what you mean: That comparing `TypingClassValueWithIndex(Tuple)[int]` with tuple[int] doesn't work.

I feel like we should catch this case in `DefineGenericBase.is_same_class`. Everything else seems to crash at some points (with gotos on stubs and stuff).

PeterJCLaw: Sorry for the belated response here. Thanks for the details around the various representations -- that's really useful.

I don't think that changing `DefineGenericBase.is_same_class` is going to solve the issue this PR is trying to solve (though it may be what's needed in #1560). Having put some `print`s in `DefineGenericBase.is_same_class` and running the new tests on this branch, I can't see anywhere that we compare the tuple forms in that method. Perhaps I'm missing something about what you're suggesting here?

davidhalter: Hmmm, that makes sense. I'll try to look at it on Friday. Too much to other things are happening at the moment.

BTW: I'm pretty happy with the patch now. It's more like can we get rid of `execute_annotation()` or not.

davidhalter:
I feel like this is in a good shape and does exactly what it should. I'm trying to work out some stuff locally, but that doesn't really have to anything with this branch.

stonebig:
hi @davidhalter , when do you expect to release a patch-release 0.17.1 ?

PR #1558, opened 2020-04-23
Api classes typos

joshbax189:
Some minor docstring typos and fix an accidental use of is_builtin_module as a property not a method.
(In deprecated method, so I didn't add a test.)

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -94,7 +94,11 @@ class BaseName(object):
 
     @property
     def module_path(self):
-        """Shows the file path of a module. e.g. ``/usr/lib/python2.7/os.py``"""
+        """
+        Shows the file path of a module. e.g. ``/usr/lib/python2.7/os.py``
+
+        :rtype: str or None
+        """
         module = self._get_module_context()
         if module.is_stub() or not module.is_compiled():
             # Compiled modules should not return a module path even if they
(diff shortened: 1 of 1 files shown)


jedi/api/classes.py
```diff
@@ -734,7 +738,7 @@ def desc_with_module(self):
             DeprecationWarning,
             stacklevel=2
         )
-        position = '' if self.in_builtin_module else '@%s' % self.line
+        position = '' if self.in_builtin_module() else '@%s' % self.line
```
davidhalter, line 741: This mistake was made in 0cb23dcfa2e54481c3d96c9fb281641a89a86cd4. That's from 2014. So IMO we should probably just remove the position. It was never applied anyway and nobody asked for it. Just remove this whole line and we should be good :)

joshbax189: Aha! It was referenced from emacs-jedi, but with that observation it is completely replaced by full_name.

joshbax189:
Hi David! I had noticed that you have been working on jedi for a long time and were still actively working on it: I thought it must be a quality project. I was toying with refactoring using v0.16 and noticed that quite a few open issues in other refactoring tools could be addressed in only a few lines using jedi. But you closed that issue before I had finished playing around with edge cases :D So instead I thought I would update emacs-jedi to use the new API.

davidhalter:
> But you closed that issue before I had finished playing around with edge cases :D 

So you mean you were actually implementing refactoring, while I was finishing it? Haha, that's so unlucky, because it took me so long to finally tackle it :)

davidhalter:
Good stuff, thanks! Happy to see you around here in the future as well :)

PR #1559, opened 2020-04-24
Add support for generic optional parameters

PeterJCLaw:
This adds support for things like:

``` python
optional_float = 42.  # type: Optional[float]
list_of_optional_floats = [optional_float]  # type: List[Optional[float]]

def optional_t_to_list_t(x: Optional[T]) -> List[T]:
    return [x] if x is not None else []

for xz in optional_t_to_list_t(optional_float):
    #? float()
    xz

def list_optional_t_to_list_t(the_list: List[Optional[T]]) -> List[T]:
    return [x for x in the_list if x is not None]

for xy in list_optional_t_to_list_t(list_of_optional_floats):
    #? float()
    xy
```

Discovered while exploring alternatives for #1554.
(Not sure how I missed this previously!)


davidhalter:
You really are on a hunt here to find all those cases :)

I'll put this PR on a hold until we figure out things in #1554. It might affect this PR as well.

jedi/inference/gradual/typing.py
```diff
@@ -143,6 +143,42 @@ def _create_instance_with_generics(self, generics_manager):
...
+        if annotation_name == 'Optional':
+            if is_class_value:
+                # This only applies if we are comparing something like
+                # List[Optional[int]] with Iterable[Optional[int]]. First, Jedi
+                # tries to match List/Iterable. After that we will land here,
+                # because is_class_value will be True at that point. Obviously
+                # we also compare below that both sides are `Optional`.
+                for element in value_set:
+                    element_name = element.py__name__()
+                    if element_name == 'Optional':
```
PeterJCLaw, line 166: It's since occurred to me that this isn't strictly correct -- it's completely valid to pass a non-optional value into a function which expects an optional one. I'll need to add some tests for this to see if this works for that case, though I suspect it won't.

There is a slight gotcha here, in that `mypy` won't actually allow this in all cases due to covariance rules. For example, you can't pass `List[int]` to `List[Optional[int]]` even though you can pass it to `Sequence[Optional[int]]`.

I _suspect_ jedi might just be able to ignore that though. It'll mean jedi will offer completions where `mypy` might complain about the types, but I'm not sure if that's a big deal. (IMO that's preferable to not offering completions where they could be valid).

davidhalter: > mypy won't actually allow this in all cases due to covariance rules

Hmm why is that? Isn't the former a covariant of the latter and should therefore be OK? It feels quite stupid to pass that. (I'm assuming we are talking about a case where you pass the former to `def foo(x: List[Optional[int]]): ...`).

> I suspect jedi might just be able to ignore that though. It'll mean jedi will offer completions where mypy might complain about the types, but I'm not sure if that's a big deal. (IMO that's preferable to not offering completions where they could be valid).

100% agree. Mypy and Jedi have different goals. I also think that the case you mention would never result in an issue, because it's valid typing.

PeterJCLaw: Covariance rules are a little complicated to grok. Essentially it comes down to this sort of behaviour:

``` python
def cannot_accept_non_optional(items: List[Optional[int]]) -> None:
    items.append(None)

def can_accept_non_optional(items: Sequence[Optional[int]]) -> None:
    # Here items is immutable (from a type system perspective), so while
    # the code inside the function needs to cope with `None`s, they can't
    # leak out
    print(len(items))

items = [42]  # type: List[int]

cannot_accept_non_optional(items) # bad
can_accept_non_optional(items) # ok

for item in items:
    print(item * 2)
```

davidhalter: Oh, thanks a lot for these examples! I understand. I just never even thought of passing a mutable list :). I don't really write code like this, so I guess the rule makes sense if it's not immutable. It's even invariant as far as I can see.

davidhalter:
Since I merged #1572, I'm happy to look at this now. I guess it makes sense however to try to get tests passing again with the new code.

jedi/inference/gradual/typing.py
PeterJCLaw: > It's since occurred to me that this isn't strictly correct -- it's completely valid to pass a non-optional value into a function which expects an optional one. I'll need to add some tests for this to see if this works for that case, though I suspect it won't.

7c6c62a5 adds said tests. After updating this branch for the recent removal of `is_class_value` (i.e; removing that branch) it turns out that those tests already pass :)

davidhalter:
Thanks!

Issue #1560, opened 2020-04-25
No completions for method returning class generic type when __init__ contains nested tuple type


PeterJCLaw:
(Discovered while exploring a issue with `NamedTuple`s)
``` python
from typing import List, Tuple, TypeVar, Generic, NewType

T = TypeVar('T')


class That(Generic[T]):
    # It doesn't seem to matter where the `T` comes in this signature - just
    # somewhere so that it's valid from an annotation perspective.
    def __init__(self, items: List[Tuple[str, T]]) -> None:
        pass

    # I don't think it matters that this is `T` specifically - it just needs to
    # be something non-concrete, so that jedi is forced to evaluate the class
    # rather than being able to short-circuit that by looking only at the annotation.
    def get(self) -> T:
        pass

inst = That([("abc", 2)])

# No completions here, but should have completions for `int`
inst.get().
```

PeterJCLaw:
Where I've got to so far with this is that jedi rejects the signature of the `__init__` method when trying to resolve the creation of `inst`, due to `[("abc", 2)]` not looking enough like `List[Tuple[str, T]]`.

Specifically we end up in:
 - `ExecutedParamName.matches_signature` for the `items` parameter and are comparing the annotation to the given type, which (via `is_sub_class_of`) calls:
- `DefineGenericBase.is_same_class` for the list types, which recurses to match the generics thus calling:
- `DefineGenericBase.is_same_class` for the tuple types, which returns `False` because the tuple types don't match (as compared by `self.tree_node != other.tree_node`).

At the bottom level we're comparing two similar but not sufficiently equal types:

- `self`: `<GenericClass: <ClassValue: <Class: tuple@870-895>>[S{<ClassValue: <Class: str@410-557>>}, S{<ClassValue: <Class: int@138-220>>}]>`
- `other`: `TypingClassValueWithIndex(Tuple<LazyG>[S{<ClassValue: <Class: str@410-557>>}, S{<TypeVar: T>}])`

I might carry on with this at some point, however I feel I'm hitting a similar issue here to the one discussed on https://github.com/davidhalter/jedi/pull/1554#discussion_r410908948. Not because this is nested tuples per-se, but rather that jedi has two representations for a very similar concept and I've not really understood why that's the case.

davidhalter:
> At the bottom level we're comparing two similar but not sufficiently equal types:

[...]

PR #1561, opened 2020-04-26
Support accessing the py__class__ of a NewType

PeterJCLaw:
The test here is a bit contrived, the actual place I found this was in using a `NewType` as a type within a `NamedTuple`. However due to https://github.com/davidhalter/jedi/issues/1560 that currently also fails for other reasons. This still feels useful to fix on its own though.

### DIFF
--- a/test/completion/pep0484_typing.py
+++ b/test/completion/pep0484_typing.py
@@ -283,6 +283,18 @@ def testnewtype2(y):
     y
     #? []
     y.
+
+# The type of a NewType is equivalent to the type of its underlying type.
+MyInt = typing.NewType('MyInt', int)
+x = type(MyInt)
+#? type.mro
+x.mro
+
+PlainInt = int
+y = type(PlainInt)
+#? type.mro
+y.mro
+
 # python > 2.7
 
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks!

test/completion/pep0484_typing.py
```diff
@@ -283,6 +283,18 @@ def testnewtype2(y):
     y
     #? []
     y.
+
+# The type of a NewType is equivalent to the type of its underlying type.
```
PeterJCLaw, line 287: (for lack of a better place to ask this) Is there a better place to put these tests? It's obviously really convenient to keep adding to these `pep0484_*.py` files, though they're not always the easiest to ensure isolation between cases for.

Is there somewhere else I should be considering when adding tests of this kind in future?

PeterJCLaw:
Thanks for the speedy merge :) I'll follow up on the other PRs discussions tomorrow.

test/completion/pep0484_typing.py
davidhalter: No, at the moment there is no such thing. I feel like adding the tests there is fine. It hasn't really caused any issues in the past.

PR #1564, opened 2020-04-30
Reduce noise in signatures of compiled params

Endilll:
Hello again.
Things are definitely improved since my [previous PR](https://github.com/davidhalter/jedi/pull/1441).

Given my initial example:
`clip: <class 'vapoursynth.VideoNode'>, width: typing.Union[int, NoneType]=None`
Upstream Jedi and Python 3.8 got better:
`clip: VideoNode, width: typing.Union[int, NoneType]=None`
It'll receive further improvements in Python 3.9 since [my PR](https://github.com/python/cpython/pull/19714) got merged:
 `clip: VideoNode, width: typing.Optional[int]=None`

Here I have a pair of commits that pushes it even further.
1. Removing `typing.` prefix as `inspect.py` [does](https://github.com/python/cpython/blob/3209cbd99b6d65aa18b3beb124fac9c792b8993d/Lib/inspect.py#L1224):
 `clip: VideoNode, width: Optional[int]=None`
2. Hiding default value for Optional as it doesn't help anybody:
`clip: VideoNode, width: Optional[int]`

I remember you saying that a bit of noise in annotations is intentional, but you haven't got into detail, so I decided to bring everything on.

As for implementation of default value hiding, I'm not happy myself with using regex, but I haven't found a way to extract `__args__` from `annotation` even after using `create_from_access_path()`.


jedi/inference/compiled/access.py
```diff
@@ -487,7 +487,7 @@ def needs_type_completions(self):
     def _annotation_to_str(self, annotation):
         if isinstance(annotation, type):
             return str(annotation.__name__)
-        return str(annotation)
+        return str(annotation).replace('typing.', '')
```
FFY00, line 490: This is wrong, `my_custom_module.typing.Something` is now `my_custom_module.Something`.

Endilll: I missed the fact that `inspect` first check if symbol came from `typing`. Fixed.

Endilll:
As far as I can see, CI fails because `re` in 2.7 doesn't have `fullmatch()` function. Since 0.17 was the last version to support 2.7, I suppose I don't need to fix that.

davidhalter:
> Hiding default value for Optional as it doesn't help anybody:

This is not true. An optional can still be mandatory.

> Removing typing. prefix as inspect.py does:

I'll think about this. I'm probably fine with this change.

Endilll:
> An optional can still be mandatory.

Can you elaborate, please?

Endilll:
Ah, now I get it. Sounds strange to me, but it's definitely possible.

jedi/inference/compiled/access.py
```diff
@@ -487,7 +487,10 @@ def needs_type_completions(self):
     def _annotation_to_str(self, annotation):
         if isinstance(annotation, type):
             return str(annotation.__name__)
-        return str(annotation)
+        s = str(annotation)
+        if getattr(annotation, '__module__', None) == 'typing':
```
davidhalter, line 491: Can you add an `and s.startswith('typing.')` and add a `1` as a param to replace? That way it will not break in the future.

Endilll: Done

Endilll:
But it bothers me that this implementation fails to work correctly for perfectly reasonable `typing.Optional[typing.Union[int, float]]`, removing only the outer prefix. I'll share my idea in a bit.

Endilll:
Why to reimplement `inspect` logic if we can reuse it? If there're concerns about `typing.` substrings not pointing to `typing` module being removed, we better report it upstream.

This also reverts a side effect of #1508 which I think was overlooked. Generating annotation string using `__name__` strips all the information about module it comes from, as well as qualified name. This side effect is depicted in the description: `clip: <class 'vapoursynth.VideoNode'>` became `clip: VideoNode` in upstream Jedi. `inspect.formatannotation()` got this right, and even leaves the option to strip module name (just it, and not qualified name): `inspect.formatannotation(p.annotation, p.annotation.__module__)`.

Endilll:
Adding to concerns I mentioned above, I tried to craft code that supposed to be broken by aggressive `replace()` while working on test, but wasn't able to. If you can craft it, and it will be reasonable enough to go to CPython folks with, I don't mind submitting PR that stops breaking it.

FFY00:
The problem is that there may be context you are lacking in the original code, eg. X condition will never happen in the original code, while it might here.

jedi/inference/compiled/access.py
```diff
@@ -498,7 +493,7 @@ def get_signature_params(self):
                 default_string=repr(p.default),
                 has_annotation=p.annotation is not p.empty,
                 annotation=self._create_access_path(p.annotation),
-                annotation_string=self._annotation_to_str(p.annotation),
+                annotation_string=inspect.formatannotation(p.annotation),
```
FFY00, line 496: Where does this function come from? I can't find anything in the documentation.

Endilll:
> in the original code

Is there any for compiled functions?

> We still should evaluate it on our own terms. "They do it, so we can do it" is not valid here.

I agree that we shouldn't follow `inspect` blindly. But my point is that if we see issues with its approach that survived 3.5 years of usage (that particular line with `str.replace()`), we need to let CPython devs know of it as well. I don't think hand-waving will hold in a discussion with them, neither it should be sufficient in our discussion here.

Hence I don't want to make any changes based on this concern until more strong evidences arise. I tried to find them myself, but to not avail. Probably I'm not up to the task, but fortunately there are at least 3 people here.

One more note is that the stronger concern is, the more it needs a test. Which means we need code to reason about.

jedi/inference/compiled/access.py
davidhalter: Please return an empty string for Python 2. I don't want this to break in that case.

> I can't find anything in the documentation.

Yeah, I'm also a bit worried by that. Especially because `inspect.formatargspec` (the function that uses it) is deprecated. However Python is usually very stable about functions that have no leading underscore, especially for micro versions. I think it's fine to use this.

Also note that I'm not a big fan of the function at all. It's buggy (it does just replace `typing.`) and not well maintained. I just think that it's not going to matter for most people.

davidhalter:
> The typing code may be used as a starting point but it should be nothing more than that. We still should evaluate it on our own terms. "They do it, so we can do it" is not valid here.

That's actually a very good point that I haven't thought of. The fact that the function is not documented means basically they can do pretty much whatever with it. The only reason why I'm kind of ok with using it is because I know how slow CPython moves in these regards.

davidhalter:
Just to make it clear, I'm waiting for @Endilll to make sure it doesn't crash in Python 2. (Please also resolve the merge conflict.

davidhalter:
Thanks again!

Issue #1566, opened 2020-05-01
Add ".venv" to `jedi.inference.references._IGNORE_FOLDERS`


pappasam:
".venv" is a popular virtual environment folder name; `project.search` gets really mucked up when it isn't ignored.

I've gotten around this issue in https://github.com/pappasam/jedi-language-server with the following line:

```python
import jedi.inference.references
# TODO: remove this once Jedi ignores '.venv' folders by default
# without this line, Jedi will search in '.venv' folders
jedi.inference.references._IGNORE_FOLDERS = (
    *jedi.inference.references._IGNORE_FOLDERS,
    ".venv",
)
```

We can also consider making this `_IGNORE_FOLDERS` a jedi configuration. I can make it configurable in my language server for end users, but I figure we probably want an official jedi API that isn't quite as "private" as the above. What do you think?

davidhalter:
Thanks for looking into this. I think configuration options like this for venvs are probably the wrong idea.

I think it might be better to just remove venv's from the equation, if there's a `bin/activate`. I'm however also happy to just add the `.venv` to the `_IGNORE_FOLDERS` upstream.

davidhalter:
And btw: The above API is private. It may change at any time.

pappasam:
Understood. And that’s exactly why I opened this issue!

PR #1570, opened 2020-05-05
Add .venv to _IGNORE_FOLDERS

pappasam:
".venv" is a popular virtual environment folder name; project.search gets really mucked up when it isn't ignored.

This tiny PR resolves https://github.com/davidhalter/jedi/issues/1566


davidhalter:
Thanks!

As I said before, maybe we should improve the logic to detect virtualenvs, but for now this is fine.

If other people complain I might implement that, but I guess this is definitely not wrong.

Issue #926, opened 2017-05-23
Jedi cache directory on Windows (AppData\Roaming)


gandhis1:
The default jedi cache directory is the roaming folder. In my company, we have restrictions on the file size of the Roaming folder (since I believe these settings are synced across computers). The jedi cache really isn't a fundamental user setting that needs to be synced across devices, so it really should be in the local app data folder. So I propose here that the default windows path use the environment variable `%LocalAppData%` instead.

davidhalter:
I have no idea about typical windows folders whatsoever. Any other opinions?

gandhis1:
Here's an excerpt from stackoverflow (https://superuser.com/questions/21458/why-are-there-directories-called-local-locallow-and-roaming-under-users-user):

> Roaming is the folder that would be synchronized with a server if you logged into a domain with a roaming profile (enabling you to log into any computer in a domain and access your favorites, documents, etc. Firefox stores its information here, so you could even have the same bookmarks between computers with a roaming profile.
> 
> Local is the folder that is specific to that computer - any information here would not be synchronized with a server. This folder is equivalent in Windows XP to C:\Documents and Settings\User\Local Settings\Application Data.
> 
> LocalLow is the same folder as local, but it has a lower integrity level. Internet Explorer 8, for example, can only write to the locallow folder (when protected mode is on).

Based on this, I would think Local is the way to go.

davidhalter:
Can you fix it in parso and create a pull request (jedi's future parser)? https://github.com/davidhalter/parso

davidhalter:
This was changed.

PR #1575, opened 2020-05-10
This fixes two issues with the caching on Windows / APPDATA

mmaeusezahl:
The problem is that Jedi starts to fill the organizational storage space in %APPDATA% on Windows. In my case (with just minimal use) I have a 400MB Jedi folder that gets synced across our organizational network each time I log in (I am using Visual Studio Code which in turn uses Jedi).

### Fixes 

* The cache directory should really be %LOCALAPPDATA%
 * ~ is not a meaningful directory on Windows. It should really be
   os.path.expanduser('~'). To be honest it is probably always safe to
   assume that os.getenv('LOCALAPPDATA') executes to something sensible
   on any Windows system that hasn't been tampered with.

### Reasoning and a historical side note:

This issue was already reported once here: https://github.com/davidhalter/jedi/issues/926

It got then partially fixed, but only for parso in a different place (thinking of which, the second issure should probably also get ported there as well): https://github.com/davidhalter/parso/pull/1

Someone also temporarily fixed it in [pythonVSCode](https://github.com/DonJayamanne/pythonVSCode) (now maintained as [microsoft/vscode-python](https://github.com/microsoft/vscode-python)): https://github.com/DonJayamanne/pythonVSCode/pull/1035

But it got lost in https://github.com/microsoft/vscode-python/commit/20a7bd5ab03adac8bc63022a13049aaf748c64bc when Jedi was introduced as an external dependency much rather than pulling it in as external code.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -57,6 +57,7 @@ Code Contributors
 - Tim Gates (@timgates42) <tim.gates@iress.com>
 - Lior Goldberg (@goldberglior)
 - Ryan Clary (@mrclary)
+- Max Mäusezahl (@mmaeusezahl) <maxmaeusezahl@googlemail.com>
 
 And a few more "anonymous" contributors.
 
--- a/jedi/settings.py
+++ b/jedi/settings.py
@@ -69,8 +69,8 @@ Adds an opening bracket after a function for completions.
 # ----------------
 
 if platform.system().lower() == 'windows':
-    _cache_directory = os.path.join(os.getenv('APPDATA') or '~', 'Jedi',
-                                    'Jedi')
+    _cache_directory = os.path.join(os.getenv('LOCALAPPDATA') or 
+                                    os.path.expanduser('~'), 'Jedi', 'Jedi')
 elif platform.system().lower() == 'darwin':
     _cache_directory = os.path.join('~', 'Library', 'Caches', 'Jedi')
 else:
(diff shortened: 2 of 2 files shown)


davidhalter:
Good points. That was really my fault. I simply forgot that Jedi has also a separate path (it's debatable if that makes sense).

Thanks for the research and this fix.

I'm merging, because I believe that everybody forgot about this. It would still be nice if @micbou could verify that this is correct.

davidhalter:
I also patched it in parso: 15403fd998c2e60f9f9dc7def819dfc0ff0926f8. I don't think it matters much since as you say `os.getenv('LOCALAPPDATA')` should always be something meaningful. However now they should both be very similar.

mmaeusezahl:
Maybe it would be sensible to use the [appdirs](https://pypi.org/project/appdirs/) package in the first place instead of having roughly the same (error prone to future Windows cache locations) code in two code-bases?

Thanks for the quick merge!

davidhalter:
That's not really an option, because I don't want dependencies for something as small as that.

mmaeusezahl:
Alright I try to keep it that way too (usually) 👍

Looking at the code of appdirs, the main difference seems to be that it tries to call a low-level Windows function instead of os.getenv. I can't really tell when this would be necessary -- it's a thing that Python would hardly support anyway.

Issue #1576, opened 2020-05-10
Feature: end_line and end_column for api.classes.BaseName


pappasam:
## Feature

When receiving a `Name`, there are use-cases where it's helpful to know the `end_line` and `end_column` associated with a symbol.

Name already supports `line` and `column`, so it doesn't seem like too much of a stretch to add these two new attributes.

## Use case

In [jedi-language-server](https://github.com/pappasam/jedi-language-server), these values would be useful to support symbol range selection (used to create text objects based on symbols found in the document, including Functions, Classes, whatever LSP defines).

I currently support this by accessing a hidden variable, and the result isn't perfect. The `end_column` is normally `0` and does not cover the full range of characters (eg, it'll omit the return statement of a function). I therefore hack it by adding a huge number in my return value. It'd be nice if Name's public API supported this so it doesn't need to be a hack.

```python
def _document_symbol_range(name: Name) -> Range:
    """Get accurate full range of function

    Thanks https://github.com/CXuesong from
    https://github.com/palantir/python-language-server/pull/537/files for the
    inspiration!

    Note: I add tons of extra space to make dictionary completions work. Jedi
    cuts off the end sometimes before the final function statement. This may be
    the cause of bugs at some point.
    """
    _name = (
        name._name.tree_name.get_definition()  # pylint: disable=protected-access
    )
    (start_line, start_column) = _name.start_pos
    (end_line, end_column) = _name.end_pos
    return Range(
        start=Position(start_line - 1, start_column),
        end=Position(end_line - 1, end_column + 20000),
    )
```

See here for this function in context: https://github.com/pappasam/jedi-language-server/blob/v0.13.1/jedi_language_server/jedi_utils.py#L87

davidhalter:
> I currently support this by accessing a hidden variable, and the result isn't perfect. The end_column is normally 0 and does not cover the full range of characters (eg, it'll omit the return statement of a function). I therefore hack it by adding a huge number in my return value. It'd be nice if Name's public API supported this so it doesn't need to be a hack.

Can you give me an example?

----

In general, how is your approach holding up? Is it good enough?

It's also unclear to me where you want the end position to be. Is it at the end of the last symbol of a function? At the end of the last comment? At the start of the next symbol?

pappasam:
@davidhalter I just did a proof of concept and it seems like [coc](https://github.com/neoclide/coc.nvim) is the main culprit of the weird ranges, not Jedi for the most part

[...]

PR #1584, opened 2020-05-16
Add BaseName.definition_[start,end]_position

pappasam:
Provides two public (property) methods getting the (row, column) of the start / end of the definition range. Rows start with 1, columns start with 0.

Resolves https://github.com/davidhalter/jedi/issues/1576

Note: I cannot run tests locally so, unfortunately, I didn't write any. See comment in issue mentioned above.

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -231,6 +231,37 @@ class BaseName(object):
             return None
         return start_pos[1]
 
+    @property
+    def definition_start_position(self):
+        """
+        The (row, column) of the start of the definition range. Rows start with
+        1, columns start with 0.
+
+        :rtype: Tuple[int, int]
+        """
+        definition = self._name.tree_name.get_definition()
+        if definition is None:
+            return self._name.start_pos
+        return definition.start_pos
+
+    @property
+    def definition_end_position(self):
+        """
+        The (row, column) of the end of the definition range. Rows start with
+        1, columns start with 0.
+
+        :rtype: Tuple[int, int]
+        """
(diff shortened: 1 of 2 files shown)


pappasam:
I’ll write tests soon, please don’t merge until then!

pappasam:
@davidhalter I've added the tests!

davidhalter:
Thanks again!

I ended up changing a few things that I didn't notice and wanted you to know:

- I'm not using properties anymore. For most things in APIs properties are a bad idea, because you cannot modify their behavior with params anymore. 2d17b813138f400ef948d42cb411563af34077c2
- Used parametrize for tests, I like that better for tests in Python fa6194c0a9c5907e2c8fe046009d3aea2b9d53a4
- There was an AttributeError for some cases (and the return value can be None -> so it's an Optional). 8fdf16b316aedf6f581ef97ac9d63c843cf25898

pappasam:
@davidhalter ah, gotcha, glad you caught those!

That said, the change in 8fdf16b may introduce a bug. The way I read it, if `self._name.tree_name` is None, the line `definition = self._name.tree_name.get_definition()` would raise a missing attribute exception, no?

We'd probably want the check to go:

```python
if self._name.tree_name is None:
    return None
definition = self._name.tree_name.get_definition()
if definition is None:
    return self._name.tree_name.end_pos
```

Also, we'd probably also want to do this exact same check for `get_definition_start_position`, right?

davidhalter:
Very good point. I think we finally arrived at a good solution :).

PR #1586, opened 2020-05-18
Support more Django model fields

PeterJCLaw:
Adds support for a few more Django model field types.

There are still a few field types left, but I think this is still useful.

At some point it might be worth exploring a better understanding of the inheritance chains (so that things like `EmailField` and `URLField` fall out of `CharField` without explicit listing), but just listing them works for now.


davidhalter:
Awesome, thanks!

PR #1589, opened 2020-05-20
Import FileNotFoundError from jedi._compatibility

cc-a:
Found this to be preventing calls to get_type_hint in some cases for Python 2.


davidhalter:
Thanks!

PR #1590, opened 2020-05-21
Get references in the current module only

muffinmad:
Trying to implement response to the [`textDocument/documentHighlight`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentHighlight) LSP request:

> For programming languages this usually highlights all references to the symbol scoped to this file

One posible solution is to filter out Names returned by the `Script.get_references`. But it's not very effective: Jedi will do amazing job collecting the symbol references across all related files and then most of the results will be thrown away!

This PR propose `all_scopes` parameter for the `get_references` method. If `False`, collect references in the current module only.

Thanks!

### DIFF
--- a/jedi/inference/references.py
+++ b/jedi/inference/references.py
@@ -25,7 +25,7 @@ easily 100ms for bigger files.
 """
 
 
-def _resolve_names(definition_names, avoid_names=(), all_scopes=True):
+def _resolve_names(definition_names, avoid_names=()):
     for name in definition_names:
         if name in avoid_names:
             # Avoiding recursions here, because goto on a module name lands
@@ -37,7 +37,7 @@ def _resolve_names(definition_names, avoid_names=(), all_scopes=True):
             # names when importing something like `import foo.bar.baz`.
             yield name
 
-        if all_scopes and name.api_type == 'module':
+        if name.api_type == 'module':
             for n in _resolve_names(name.goto(), definition_names):
                 yield n
 
@@ -49,17 +49,16 @@ def _dictionarize(names):
     )
 
 
-def _find_defining_names(module_context, tree_name, all_scopes=True):
-    found_names = _find_names(module_context, tree_name, all_scopes=all_scopes)
+def _find_defining_names(module_context, tree_name):
+    found_names = _find_names(module_context, tree_name)
 
-    if all_scopes:
-        for name in list(found_names):
-            # Convert from/to stubs, because those might also be usages.
-            found_names |= set(convert_names(
-                [name],
-                only_stubs=not name.get_root_context().is_stub(),
-                prefer_stub_to_compiled=False
-            ))
+    for name in list(found_names):
+        # Convert from/to stubs, because those might also be usages.
+        found_names |= set(convert_names(
+            [name],
+            only_stubs=not name.get_root_context().is_stub(),
+            prefer_stub_to_compiled=False
+        ))
 
     found_names |= set(_find_global_variables(found_names, tree_name.value))
     for name in list(found_names):
@@ -67,15 +66,15 @@ def _find_defining_names(module_context, tree_name, all_scopes=True):
                 or name.tree_name.parent.type == 'trailer':
             continue
         found_names |= set(_add_names_in_same_context(name.parent_context, name.string_name))
-    return set(_resolve_names(found_names, all_scopes=all_scopes))
+    return set(_resolve_names(found_names))
 
 
-def _find_names(module_context, tree_name, all_scopes=True):
+def _find_names(module_context, tree_name):
     name = module_context.create_name(tree_name)
-    found_names = set(name.goto(all_scopes=all_scopes))
+    found_names = set(name.goto())
     found_names.add(name)
 
-    return set(_resolve_names(found_names, all_scopes=all_scopes))
+    return set(_resolve_names(found_names))
 
 
 def _add_names_in_same_context(context, string_name):
@@ -114,7 +113,7 @@ def _find_global_variables(names, search_name):
                     yield n
 
 
-def find_references(module_context, tree_name, all_scopes=True):
+def find_references(module_context, tree_name, only_in_module=False):
     inf = module_context.inference_state
     search_name = tree_name.value
 
@@ -122,20 +121,20 @@ def find_references(module_context, tree_name, all_scopes=True):
     # certain cases, we want both sides.
     try:
         inf.flow_analysis_enabled = False
-        found_names = _find_defining_names(module_context, tree_name, all_scopes=all_scopes)
+        found_names = _find_defining_names(module_context, tree_name)
     finally:
         inf.flow_analysis_enabled = True
 
     found_names_dct = _dictionarize(found_names)
 
     module_contexts = [module_context]
-    if all_scopes:
+    if not only_in_module:
         module_contexts.extend(
             m for m in set(d.get_root_context() for d in found_names)
             if m != module_context and m.tree_node is not None
         )
     # For param no search for other modules is necessary.
-    if not all_scopes or any(n.api_type == 'param' for n in found_names):
+    if only_in_module or any(n.api_type == 'param' for n in found_names):
         potential_modules = module_contexts
     else:
         potential_modules = get_module_contexts_containing_name(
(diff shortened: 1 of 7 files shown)


davidhalter:
Thanks for this! :)

I think before I get into the details, I think we have to clarify a few cases.

- What about `foo().bar`, where bar is defined in another file
- What about keyword params where the function is defined in another file

I think your implementation ignores these cases. Regardless if we do too much work, I think your implementation of `all_scopes` should throw any name away that is not part of the current file. Otherwise it's just confusing for users, because some names get filtered, while others are not filtered.

muffinmad:
> * What about `foo().bar`, where bar is defined in another file

Do you mean something like this?

```python
from datetime import datetime

d1 = datetime.now()

d2 = datetime.now()
```

With cursor under `now`, both `now` must be highlighted. By "highlighted" I mean that `get_references(3, 14, all_scopes=False)` must return `[(3, 14), (5, 14)]`.

> * What about keyword params where the function is defined in another file

Can we please have some example code? ;-)

> I think your implementation ignores these cases. Regardless if we do too much work, I think your implementation of `all_scopes` should throw any name away that is not part of the current file.

Yes, that's the idea: find references only in the current file.

> Otherwise it's just confusing for users, because some names get filtered, while others are not filtered.

Maybe it must be another method and not the `get_references` param? Or param name must be changed to reflect that references of a variable in the file will be listed but not in the project.

davidhalter:
> Do you mean something like this?

Yes to the datetime example.

> With cursor under now, both now must be highlighted. By "highlighted" I mean that get_references(3, 14, all_scopes=False) must return [(3, 14), (5, 14)].

That's of course true, but it's important that get_references doesn't return the definition, which is in a different file.

> Can we please have some example code? ;-)

Something like

```
def x(foo):
    return foo

x(foo=3)
```

> Maybe it must be another method and not the get_references param?

I like the param. It's essentially a minor filter.

> Or param name must be changed to reflect that references of a variable in the file will be listed but not in the project.

I would probably suggest something like `only_in_script=False` or something similar. But we can decide this later, I'm more interested in what works and what doesn't at the moment :).

PeterJCLaw:
A quick thought on the API design (for when that discussion continues): are there other scopes which clients might want to ask Jedi about?
For example, might it be useful to be able to ask Jedi for all the references in the current package? In the current Project?
If so, maybe an enum of expanding scopes might be more suitable than a named boolean? (I realise this is a little complicated given the Python 2 support, but if the plan is to drop that anyway it might be a way to go)

muffinmad:
Thanks @davidhalter ! Added test to cover the cases you mention. Proposed implementation not working as expected yet.

davidhalter:
> A quick thought on the API design (for when that discussion continues): are there other scopes which clients might want to ask Jedi about?
> For example, might it be useful to be able to ask Jedi for all the references in the current package? In the current Project?

Good point! I kind of forgot about this, but I agree. The default scope at the moment is the `project`. But it might be interesting to have a `global` scope which includes libraries/stdlib, a `package` scope or even a `range` scope that will just return references for that range. However, I think we can decide this at the end.

> If so, maybe an enum of expanding scopes might be more suitable than a named boolean?

Or we could just use a string. I'm open to both options.

muffinmad:
Now all the get-local-references tests are passed, but I'm certainly lost in the implementation.

@davidhalter can you please give me some advices towards the right way to implement this?

Thanks!

davidhalter:
Sorry for the delay, but I might need some more days to review this. It's big PR so it will take some time for me to review it.

jedi/inference/references.py
```diff
@@ -113,25 +114,28 @@ def _find_global_variables(names, search_name):
                     yield n
 
 
-def find_references(module_context, tree_name):
+def find_references(module_context, tree_name, all_scopes=True):
```
davidhalter, line 117: I think we should use `only_in_module=False`

jedi/inference/references.py
```diff
@@ -113,25 +114,28 @@ def _find_global_variables(names, search_name):
...
     inf = module_context.inference_state
     search_name = tree_name.value
 
     # We disable flow analysis, because if we have ifs that are only true in
     # certain cases, we want both sides.
     try:
         inf.flow_analysis_enabled = False
-        found_names = _find_defining_names(module_context, tree_name)
+        found_names = _find_defining_names(module_context, tree_name, all_scopes=all_scopes)
```
davidhalter, line 125: I disagree with this. it should always find the defining names in all files. We just shouldn't return them. We would probably need to filter them in the end.

jedi/inference/references.py
```diff
@@ -140,10 +144,15 @@ def find_references(module_context, tree_name):
...
+    gotos = set()
+    if not all_scopes:
+        for name in found_names_dct.values():
+            gotos |= set(name.goto())
+
     non_matching_reference_maps = {}
     for module_context in potential_modules:
         for name_leaf in module_context.tree_node.get_used_names().get(search_name, []):
-            new = _dictionarize(_find_names(module_context, name_leaf))
+            new = _dictionarize(_find_names(module_context, name_leaf, all_scopes=all_scopes))
```
davidhalter, line 155: I think this all_scopes is also not needed and with that gone, you can probably remove quite a lot of code that modifies `goto`.

jedi/inference/references.py
```diff
@@ -140,10 +144,15 @@ def find_references(module_context, tree_name):
             search_name,
         )
 
+    gotos = set()
+    if not all_scopes:
+        for name in found_names_dct.values():
+            gotos |= set(name.goto())
```
davidhalter, line 150: With the modification removed below, I think you won't need the `gotos` variable anymore. The filter below will probably be as simple as `.get_root_context() == module_context` (where module_context is the original param that was not modified by the loop variable).

test/test_api/test_usages.py
```diff
@@ -15,3 +15,78 @@ def get(include):
...
+    assert len(int_places) == 2
+    assert int_places == r(11, 22)
+
+    baz_places = r(11, 4)
+    assert len(baz_places) == 2
+    assert baz_places == r(12, 11)
+
+    script = Script('from datetime', project=project)
+    places = r(1, 5)
+    assert len(places) == 1
```
davidhalter, line 73: While this test does everything it should, we should probably rewrite it with `pytest.mark.parametrize`. There's a lot of example in the Jedi codebase and I'm happy to help if you don't know exactly how to do that. It's way easier to debug things with pytest if something like the 4th test fails.

davidhalter:
Good changes overall. There is mostly one thing that needs changes (everything else are details): The goto should probably not be modified.

jedi/inference/references.py
pappasam: From an LSP perspective, this function would be used for [documentHighlight](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentHighlight), which is generally called on `CursorHold` in Vim:

```vim
autocmd CursorHold * silent call CocActionAsync('highlight')
```

Ideally, module-only references wouldn't need to do any unnecessary searching, which would almost certainly contribute to unreasonably high CPU usage with the documentHighlight use-case.

davidhalter: This is not an additional search. This is the search for the actual definition, which is important. Otherwise you won't be able to e.g. find all references of str in your file.

pappasam: Ahh, that makes sense.

tox.ini
```diff
@@ -41,3 +41,5 @@ commands =
 commands = flake8 --extend-ignore F401 {posargs:jedi}
 deps =
 extras = qa
+[pycodestyle]
+max-line-length=100
```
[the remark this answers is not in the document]
muffinmad: OK. But can we have the `pycodestyle` section alongside with the `flake8` section in this case? Or this change is for different PR?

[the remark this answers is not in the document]
davidhalter: Oh now I realized that it's not the same "directive". If it's possible to use in `setup.cfg` I'm fine with that. If it's not we can also keep it in tox.ini. But since I want to get rid of tox eventually I'd rather not have it in that file.

muffinmad:
@davidhalter Thanks for the very helpful review!

test/test_api/test_usages.py
```diff
@@ -15,3 +18,45 @@ def get(include):
...
+    ('', [(7, 4), (8, 11)]),
+    ('', [(7, 22), (11, 22)]),
+    ('', [(11, 4), (12, 11)]),
+    ('from datetime', [(1, 5)]),
+    ('''from datetime import datetime
+d1 = datetime.now()
+d2 = datetime.now()
+''', [(2, 14), (3, 14)]),
+    ('''from jedi import Script
+Script(code='')
```
davidhalter, line 37: Can you double that? To make sure Jedi finds both params?

davidhalter:
Great set of changes, thanks!

muffinmad:
> Great set of changes, thanks!

Thank you!

Can we have 0.17.1 anytime soon? ;-)

davidhalter:
I want to improve Django support a bit, but after that: sure!

PR #1593, opened 2020-05-22
Basic support for custom Django object managers

### LINKED ISSUE #1588: Support custom Django object managers
The new Django support unfortunately lost support for custom object managers (previously only they were offered as completions).

For example:
``` python
from django.db import models

class ThingManager(models.Manager):
    def more_things(self):
        return self.all()

class Thing(models.Model):
    name = models.TextField()

    objects = ThingManager()

Thing.objects.<tab>
```

Ideally this would offer all the usual manager completions, plus `more_things`.

Bonus points if Jedi can get this right for relations which reference `Thing` too.

PeterJCLaw:
This fixes #1588.

For the moment this support is limited to just `Model.objects` replacements and does not use the custom manager for `ForeignKey` related managers.

Since the latter were previously not supported anyway, I think this is a reasonable limitation for now.

### DIFF
--- a/jedi/plugins/django.py
+++ b/jedi/plugins/django.py
@@ -110,15 +110,47 @@ def _create_manager_for(cls, manager_cls='BaseManager'):
 
 
 def _new_dict_filter(cls):
-    filters = cls.get_filters(is_instance=True, include_metaclasses=False)
+    def get_manager_name(filters):
+        for f in filters:
+            names = f.get('objects')
+            if not names:
+                continue
+
+            # Found a match. Either the model has a custom manager, or we're
+            # now in django.db.models.Model. If the latter we need to use
+            # `_create_manager_for` because the manager we get from the
+            # stubs doesn't work right.
+
+            name = names[0]  # The first name should be good enough.
+
+            parent = name.get_defining_qualified_value()
+            if parent.py__name__() == 'Model':
+
+                django_models_model, = cls.inference_state.import_module(
+                    ('django', 'db', 'models', 'base'),
+                ).py__getattribute__('Model')
+                if django_models_model == parent:
+                    # Don't want to use the value from the Django stubs, but
+                    # we have found the point where they'd take precedence.
+                    break
+
+            return name
+
+        manager = _create_manager_for(cls)
+        if manager:
+            return manager.name
+
+    filters = list(cls.get_filters(is_instance=True, include_metaclasses=False))
     dct = {
         name.string_name: DjangoModelName(cls, name)
-        for filter_ in reversed(list(filters))
+        for filter_ in reversed(filters)
         for name in filter_.values()
     }
-    manager = _create_manager_for(cls)
-    if manager:
-        dct['objects'] = manager.name
+
+    manager_name = get_manager_name(filters)
+    if manager_name:
+        dct['objects'] = manager_name
+
     return DictFilter(dct)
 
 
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
Yup, will add some more tests.

> We could override `py__get__` probably to make all of this custom `objects` code not necessary anymore:

That, or more simply `py__getattribute__` would probably work. Either approach feels like a larger change though. Maybe something to consider as part of wider support for custom managers -- for example handling them as the related manager.

> I think descriptors is the way how this should be done (and it's probably how Django implements it anyway).

It turns out that Django does the majority of its model class munging as a metaclass which assigns plain objects onto the model. There are definitely getters involved in fields, however the other aspects of a model seem to be just done in the `ModelBase` metaclass. I'm not sure this impacts how jedi implements it though.

PeterJCLaw:
> It turns out that Django does the majority of its model class munging as a metaclass which assigns plain objects onto the model. There are definitely getters involved in fields, however the other aspects of a model seem to be just done in the `ModelBase` metaclass.

Nah, I'm wrong on this. There is some metaclass involvement, but `BaseManager` does indeed set things up with a descriptor.

PeterJCLaw:
> Yup, will add some more tests.

I added the tests and it turns out they fail.

Having made some changes to get a generic instance of the `TagManager` when it's assigned into `objects`, that generic-ness doesn't end up translated into the method return values.

I'm guessing that this is a result of the `TagManager` type not itself being generic and thus the generic-ness not properly propagating into the parent type. However this leads to a conundrum -- how to modify the apparent parent type of `TagManager` without actually modifying the code. (Users unfortunately cannot make their code be `class TagManager(models.Manager[T])` because the real `models.Manager` doesn't actually inherit from `Generic` and so they'll get an error at import-time).

Perhaps we do need something akin to `BaseManagerInstanceWrapper`, which could know about this?
Alternatively, is there a way to mark which level the generic-ness should apply to?

davidhalter:
> > We could override py__get__ probably to make all of this custom objects code not necessary anymore:

> That, or more simply py__getattribute__ would probably work. Either approach feels like a larger change though. Maybe something to consider as part of wider support for custom managers -- for example handling them as the related manager.

You could probably make it work with `py__getattribute__`, but the descriptor is just a way nicer solution, since it will work if people define something like `other_objects = SomeManager()`. It will also not need special code to treat cases like `objects = 3`. Descriptors are just a very natural way to get the calling instance.

davidhalter:
> Having made some changes to get a generic instance of the TagManager when it's assigned into objects, that generic-ness doesn't end up translated into the method return values.

Yeah, this kind of sucks... I'm not sure what the correct way of action is here.

Would it be a general improvement to make it possible to understand stuff like:

```
class Foo(Generic[T]):
    def x(self) -> T: pass

class Bar(Foo): pass

Bar[int]().x()   # should maybe infer to int?!
```

I think this is not unreasonable at all. Might even be somewhere in all the typing PEPs.

I definitely think that this is something useful and might make sense for quite a lot of cases. If it's against the intention of the PEPs we should obviously not do it, but I think this would be a nice addition.

davidhalter:
BTW: Consider this merged. I'm still playing locally with getters to get this working in a more natural way.

PeterJCLaw:
Apologies for the silence here. I started on the test cases you suggested -- around making sure that `Tag.objects.first()` returned something sensible but didn't complete that.

I can't recall how far I got, though I do think it showed up an issue with the type of the returned things. It might be best to hold off merging this until that's sorted?

I've a feeling that it came down to Jedi (somewhat reasonably) not knowing that `TagManager` should return `Tag`s.

PeterJCLaw:
Replaced by #1612.

Issue #1594, opened 2020-05-22
KeyError: 'not in'


d-ryzhikov:
I'm using jedi in vim via deoplete-jedi.
The error happens when the cursor is after a dot after webtest's [TestResponse](https://docs.pylonsproject.org/projects/webtest/en/latest/testresponse.html) (when I'm trying to access its attribute `response.json`).

Traceback:
```
[deoplete] Traceback (most recent call last):
  File "/home/dmitry/.vim/bundle/deoplete.nvim/rplugin/python3/deoplete/child.py", line 194, in _gather_results
    result = self._get_result(context, source)
  File "/home/dmitry/.vim/bundle/deoplete.nvim/rplugin/python3/deoplete/child.py", line 254, in _get_result
    ctx['candidates'] = source.gather_candidates(ctx)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/sources/deoplete_jedi/profiler.py", line 37, in wrapper
    return func(self, *args, **kwargs)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/sources/deoplete_jedi.py", line 189, in gather_candidates
    completions = self.get_completions(script)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/sources/deoplete_jedi/profiler.py", line 37, in wrapper
    return func(self, *args, **kwargs)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/sources/deoplete_jedi.py", line 137, in get_completions
    return script.completions()
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/vendored/jedi/jedi/api/__init__.py", line 291, in completions
    return self.complete(*self._pos, fuzzy=fuzzy)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/vendored/jedi/jedi/api/helpers.py", line 482, in wrapper
    return func(self, line, column, *args, **kwargs)
  File "/home/dmitry/.vim/bundle/deoplete-jedi/rplugin/python3/deoplete/vendored/jedi/jedi/api/__init__.py", line 275, in complete
    return self._complete(line, column, **kwargs)
[...]

PR #1595, opened 2020-05-23
Explicitly handle `a not in b` operator comparison

PeterJCLaw:
This avoids a `KeyError` from operator_to_magic_method lookup for this case. Jedi probably could check for `__contains__` here, however as it doesn't do so for `in` checks I'm following that lead for now.

Fixes https://github.com/davidhalter/jedi/issues/1594.

### DIFF
--- a/jedi/inference/syntax_tree.py
+++ b/jedi/inference/syntax_tree.py
@@ -641,7 +641,7 @@ def _infer_comparison_part(inference_state, context, left, operator, right):
             _bool_to_value(inference_state, True),
             _bool_to_value(inference_state, False)
         ])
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
@davidhalter while fixing this I spotted that Jedi currently returns `NO_VALUES` for `in` (and now `not in`) operations - can you recall if there was a reason for these not to return `{True, False}`? Alternatively do you think Jedi should only return `{True, False}` if a `__contains__` method is found?

davidhalter:
I think it should probably return `bool()` in both cases (or True/False, either should be fine).

PR #1599, opened 2020-05-28
Upgrade setup.py's version parsing for 3.8+

isidentical:
Previous usage was deprecated due to accessing `.s` attribute of `Constant`.


davidhalter:
Great! Thanks.

PR #1600, opened 2020-05-28
fix typo of extract_function doc

yuan-xy:


jedi/api/__init__.py
```diff
@@ -751,7 +751,6 @@ def extract_function(self, line, column, **kwargs):
             global_var = 3
 
             def x():
-                foo = 3.1
```
davidhalter, line 754: I think it's the other way around: `foo = 3.1` should be part of `def x` below.

PR #1601, opened 2020-05-29
add test case to fix code example in doc

yuan-xy:


jedi/api/__init__.py
```diff
@@ -759,10 +759,11 @@ def x():
             global_var = 3
 
             def bar(foo):
-                return foo + 1 + global_var
+                return int(foo + 1 + global_var)
 
             def x(foo):
```
davidhalter, line 764: Can you remove this foo in the param, it shouldn't be there.

yuan-xy: I can't understand why.
before refactor:
```
            global_var = 3

            def x():
                foo = 3.1
                x = int(foo + 1 + global_var)
```
after refactor, as show in test, the result is:
```
            global_var = 3

            def bar(foo):
                return int(foo + 1 + global_var)

            def x(foo):
                foo = 3.1
                x = bar(foo)
```

davidhalter: No it's not, in your own test you have `def x():` both before and after

yuan-xy:
I see :)

PR #1603, opened 2020-05-29
fix line_len with \r\n

yuan-xy:


jedi/api/helpers.py
```diff
@@ -470,7 +470,7 @@ def wrapper(self, line=None, column=None, *args, **kwargs):
         line_string = self._code_lines[line - 1]
         line_len = len(line_string)
         if line_string.endswith('\r\n'):
-            line_len -= 1
+            line_len -= 2
```
davidhalter, line 473: This change is not correct IMO, because as you can see below the same line also ends with `\n` and you would therefore remove 3. I agree though that the code is unnecessarily complicated. So feel free to just make the line below an `elif` and I will merge.

yuan-xy:
you are right. I misunderstood it, and the 'elif' is added.

Issue #1608, opened 2020-06-02
Python 3.9 test failures


mgorny:
When running the test suite on Python 3.9.0b1, the following tests fail:

```python
==================================================================== FAILURES =====================================================================
__________________________________________________________ TestSetupReadline.test_import __________________________________________________________

self = <test.test_utils.TestSetupReadline testMethod=test_import>

    def test_import(self):
        s = 'from os.path import a'
        assert set(self.complete(s)) == {s + 'ltsep', s + 'bspath'}
        assert self.complete('import keyword') == ['import keyword']
    
        import os
        s = 'from os import '
        goal = {s + el for el in dir(os)}
        # There are minor differences, e.g. the dir doesn't include deleted
        # items as well as items that are not only available on linux.
        difference = set(self.complete(s)).symmetric_difference(goal)
        difference = {
            x for x in difference
            if all(not x.startswith('from os import ' + s)
                   for s in ['_', 'O_', 'EX_', 'MFD_', 'SF_', 'ST_'])
        }
        # There are quite a few differences, because both Windows and Linux
        # (posix and nt) librariesare included.
>       assert len(difference) < 22
E       AssertionError: assert 26 < 22
E        +  where 26 = len({'from os import CLD_KILLED', 'from os import CLD_STOPPED', 'from os import POSIX_SPAWN_CLOSE', 'from os import POSIX_SPAWN_DUP2', 'from os import POSIX_SPAWN_OPEN', 'from os import P_DETACH', ...})

test/test_utils.py:87: AssertionError
_____________________________________________________________ test_infer_on_generator _____________________________________________________________

Script = functools.partial(<class 'jedi.api.Script'>, environment=<Environment: 3.9.0 in /tmp/jedi/.tox/py39>)

    def test_infer_on_generator(Script):
        def_, = Script('def x(): yield 1\ny=x()\ny').infer()
>       assert def_.name == 'Generator'
E       AssertionError: assert '_SpecialGenericAlias' == 'Generator'
E         - _SpecialGenericAlias
E         + Generator

test/test_api/test_api.py:139: AssertionError
========================================= 2 failed, 3570 passed, 46 skipped, 5 xfailed in 114.37 seconds ==========================================
```

davidhalter:
I guess with your fix only the stuff with `_SpecialGenericAlias` remains.

mgorny:
[...]

PR #1609, opened 2020-06-02
Exclude more Linux constants in test_import

mgorny:
The list of differences have grown again in Python 3.9.  Instead of
increasing the allowed count let's filter out more Linux-specific
constants.  This probably makes it possible to reduce allowed
len(difference) too.

See #1608


davidhalter:
Makes sense. Thanks!

I still have no idea why they are flooding the os module, but that's really a different topic :)

PR #1612, opened 2020-06-12
Support for custom Django object managers

davidhalter:
Includes #1593 

This uses `py__get__`, which seems to work very well.

@PeterJCLaw This got way more complicated than I initially imagined. I think most changes here are internal changes in Jedi to make this possible, but it's still a bit worrying that this was so complicated.

Feel free to not really review this deeply. Just wanted to let you review it, because you essentially wrote the code that we depend on.

### DIFF
--- a/jedi/plugins/django.py
+++ b/jedi/plugins/django.py
@@ -1,16 +1,18 @@
 """
 Module is used to infer Django model fields.
 """
+from jedi._compatibility import Parameter
 from jedi import debug
+from jedi.inference.cache import inference_state_function_cache
 from jedi.inference.base_value import ValueSet, iterator_to_value_set, ValueWrapper
-from jedi.inference.filters import DictFilter, AttributeOverwrite, publish_method
-from jedi.inference.names import NameWrapper
+from jedi.inference.filters import DictFilter, AttributeOverwrite
+from jedi.inference.names import NameWrapper, BaseTreeParamName
 from jedi.inference.compiled.value import EmptyCompiledName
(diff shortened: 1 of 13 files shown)


davidhalter:
I think only a2108de2c00d62bc138ef409a244e1f951c0adc9 is important. The CI is still failing so give me a bit of time

davidhalter:
@PeterJCLaw Should be good now.

jedi/plugins/django.py
```diff
@@ -109,25 +131,160 @@ def _create_manager_for(cls, manager_cls='BaseManager'):
...
+        return func(cls, metaclasses, is_instance)
+    return wrapper
+
 
-        return func(cls, metaclasses)
+def tree_name_to_values(func):
+    def wrapper(inference_state, context, tree_name):
+        result = func(inference_state, context, tree_name)
+        if tree_name.value in ('create', 'filter', 'exclude', 'update', 'get',
+                               'get_or_create', 'update_or_create'):
```
PeterJCLaw, line 176: It took me a while to work out what the common theme was for these methods. Would it be worth pulling them out to a named constant perhaps or adding a comment here about the intent of this branch?

Maybe `FIELD_NAMES_AS_KWARGS_METHODS` or `SIGNATURE_MATCHES_MODEL_METHODS`?

davidhalter: Good point. Will do and then merge! :)

I tend to not comment enough anyway since for a long time I was the only one editing.

Issue #1607, opened 2020-06-01
No attribute completions for instance of a decorated class


eugenhu:
When I run the following,
```python
import jedi

code = """from typing import TypeVar, Type

T = TypeVar('T', Type)

def my_decorator(x: T) -> T:
    return x

@my_decorator
class MyClass:
    my_attribute = 123

def my_function(x: MyClass) -> int:
    return x."""

script = jedi.Script(code)
completion = script.complete(13, 13)

print(completion)
```
(trying to find completions after `return x.`), I get
```
[]
```
whereas I would expect jedi to suggest `my_attribute`, and it seems strange that there are no completions at all.

If I change `T = TypeVar('T', Type)` to just `T = TypeVar('T')`, then the output looks about right
```
[<Completion: my_attribute>, <Completion: __annotations__>, <Completion: __class__>, <Completion: __delattr__>, <Completion: __dict__>, <Completion: __dir__>, <Completion: __doc__>, <Completion: __eq__>, <Completion: __format__>, <Completion: __getattribute__>, <Completion: __hash__>, <Completion: __init__>, <Completion: __init_subclass__>, <Completion: __module__>, <Completion: __ne__>, <Completion: __new__>, <Completion: __reduce__>, <Completion: __reduce_ex__>, <Completion: __repr__>, <Completion: __setattr__>, <Completion: __sizeof__>, <Completion: __slots__>, <Completion: __str__>]
```
However `T = TypeVar('T', bound=Type)` also does not work.

I am using Python 3.8.2 and I've tried this both on Jedi v0.17.0 and the current latest master.

davidhalter:
Decorators are not real

The problem is that what you're doing is basically:

```
def my_decorator(x: T<type>) -> T<type>:
    return x
```

There is absolutely no guarantee that the return type is `MyClass`, it could as well be `str`.

To avoid this you will probably need to do this:

```
def my_decorator(x: Type[T]) -> Type[T]:
    return x
```

In general typing of decorators is pretty annoying, especially for functions.

eugenhu:
I see, yeah I think I get it, using
```
def my_decorator(x: Type[T]) -> Type[T]:
    return x
```
instead works.

Thanks.

eugenhu:
[...]

PR #1613, opened 2020-06-13
Handle passing functions and classes through a TypeVar

### LINKED ISSUE #1425: Autocomplete broken when typing a function decorator
Hi there,

I discovered lately that autocompletion got broken when typing a function decorator, here is the minimal example:

```python
from typing import TypeVar

R = TypeVar("R")

def dec(fn : R) -> R:
    return fn

@dec
def my_func(a, b, c):
    ...

def __main__():
    my_func()
```
**Actual:**
params_with_typevar
As you can see the signature of the function fallback to **FunctionType** instead of the actual type.

**Expected:**
params_without_typevar
I removed the return type of the decorator to get this work, but it's not suitable for my use case.

Thanks for your help !

PeterJCLaw:
This fixes #1425 and #1607 by persisting the original underlying function or class when we process a `TypeVar` they are passed into.

### DIFF
--- a/jedi/inference/gradual/type_var.py
+++ b/jedi/inference/gradual/type_var.py
@@ -1,6 +1,6 @@
 from jedi._compatibility import unicode, force_unicode
 from jedi import debug
-from jedi.inference.base_value import ValueSet, NO_VALUES
+from jedi.inference.base_value import ValueSet, NO_VALUES, ValueWrapper
 from jedi.inference.gradual.base import BaseTypingValue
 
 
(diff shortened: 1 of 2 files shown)


test/completion/pep0484_generic_passthroughs.py
```diff
@@ -142,3 +160,66 @@ def get_first(self):
...
+
+
+def will_be_decorated(var: complex) -> float:
+    pass
+
+
+is_decorated = decorator(will_be_decorated)
+
+#? will_be_decorated
+is_decorated
```
PeterJCLaw, line 225: I'd actually wanted to validate the signature here, but I don't know how to do that in these tests?

davidhalter: It's not really posible here unfortunately. We could actually make it possible, but until then we should be using the other tests. I sometimes use keyword param tests like

```
#? ['var=']
is_decorated(var)
```

to validate the signature at least partially (and keep things together).

jedi/inference/gradual/type_var.py
```diff
@@ -108,8 +108,24 @@ def execute_annotation(self):
...
-        return {annotation_name: value_set.py__class__()}
+        return {annotation_name: ValueSet(iterate())}
 
     def __repr__(self):
         return '<%s: %s>' % (self.__class__.__name__, self.py__name__())
+
+
+class TypeWrapper(ValueWrapper):
+    def __init__(self, wrapped_value, original_value):
+        super().__init__(wrapped_value)
```
davidhalter, line 127: Since this code is coming from me it's my bad, but the CI is failing. It should be `super(TypeWrapper, self)`.

We will finally get rid off Python 2 in the next few weeks, but until that point we still have to make it work :).

PeterJCLaw: D'oh. Of course. 7e637c5e

test/completion/pep0484_generic_passthroughs.py
PeterJCLaw: aha, thanks 4f11f20e..

davidhalter:
Awesome!

PR #1614, opened 2020-06-14
Support passing values through decorators from factories

PeterJCLaw:
This builds on the approach taken in https://github.com/davidhalter/jedi/pull/1613 but applies it to type vars themselves so that their type var nature is preserved when a function returns `Callable[[T], T]` and the `T` has an upper bound.

### DIFF
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -14,7 +14,7 @@ from jedi.inference.cache import inference_state_method_cache
 from jedi.inference.base_value import ValueSet, NO_VALUES
 from jedi.inference.gradual.base import DefineGenericBaseClass, GenericClass
 from jedi.inference.gradual.generics import TupleGenericManager
-from jedi.inference.gradual.type_var import TypeVar, TypeVarWrapper
+from jedi.inference.gradual.type_var import TypeVar
 from jedi.inference.helpers import is_string
 from jedi.inference.compiled import builtin_from_name
 from jedi.inference.param import get_executed_param_names
@@ -281,7 +281,7 @@ def infer_return_for_callable(arguments, param_values, result_values):
 
     return ValueSet.from_sets(
         v.define_generics(all_type_vars)
-        if isinstance(v, (DefineGenericBaseClass, TypeVar, TypeVarWrapper))
+        if isinstance(v, (DefineGenericBaseClass, TypeVar))
         else ValueSet({v})
         for v in result_values
     ).execute_annotation()
--- a/jedi/inference/gradual/annotation.py
+++ b/jedi/inference/gradual/annotation.py
@@ -14,7 +14,7 @@ from jedi.inference.cache import inference_state_method_cache
 from jedi.inference.base_value import ValueSet, NO_VALUES
 from jedi.inference.gradual.base import DefineGenericBaseClass, GenericClass
 from jedi.inference.gradual.generics import TupleGenericManager
-from jedi.inference.gradual.type_var import TypeVar
+from jedi.inference.gradual.type_var import TypeVar, TypeVarWrapper
 from jedi.inference.helpers import is_string
 from jedi.inference.compiled import builtin_from_name
 from jedi.inference.param import get_executed_param_names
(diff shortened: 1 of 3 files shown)


davidhalter:
I'm not really sure what this is for. Can you give me an example that worked differently before?

PeterJCLaw:
Ah, sorry, I'd assumed that the test cases would cover that.

Essentially this fixes the same issue as #1613 fixed, but at one layer removed in the type system.

Consider a decorator factory:
``` python
from typing import TypeVar, Callable, Any
TCallable = TypeVar('TCallable', bound=Callable[..., Any])

def decorator_with_args(name: str) -> Callable[[TCallable], TCallable]:
    def decorator(fn: TCallable) -> TCallable:
        # ... do something with `name` and `fn`
        return fn
    return decorator


@decorator_with_args(name='my_new_behaviour')
def my_new_behaviour(blah: float) -> None:
    pass


# ./sith.py run infer demo.py 17 5
my_new_behaviour
```

On `master` sith infers `typing.Callable`. With these changes, sith finds the definition of the function.

davidhalter:
There are still some test failures.

I'm still trying to understand what this does and if it's the "optimal" solution for this issue. It's pretty complicated :).

PeterJCLaw:
In terms of whether this is optimal or not, I'm not sure. I definitely wasn't expecting this to be the solution, however the more I've thought about it the more I like this.

My current thinking is that this enables the resolution of the type-var to maintain its type-var nature when it falls back onto its upper bound, which previously it couldn't do. Previously once the upper bound was evaluated there was effectively a loss of information and thus generality in its behaviour -- it then behaved as though the original definition had been that upper bound. In many cases (perhaps even most cases) that works fine because there is no later valid reference to the type-var, however in the case of a callable return type, the (unresolved) type-var must maintain its type-var nature.

Does that help?

davidhalter:
BTW: I decided to do a release. I'm happy to do a release pretty soon again if you want it. I'm generally open to do more point releases, it's just that mostly people don't really care too much about new releases :).

PeterJCLaw:
The failures here turn out to be a result of changes in how `Callable` is implemented in `typing` in Python 3.7+. Since we don't have anything to offer on a symbol which is of type `Callable` (as distinct from, say, an actual function or a `Protocol`) I've just dropped those tests for now. The tests of interest were (happily) already passing.

We might need to fix the slightly mis-handling of `Callable` which this uncovered at some point, but I don't it as an issue right now.

Aside: a possible completion we could offer on a `Callable` might be to offer the types of the parameters even if we cannot offer the names. This might also be something which Jedi could benefit from learning about anyway for positional-only parameters.

davidhalter:
> Consider a decorator factory:

Since I'm still trying to understand when this doesn't work :) I ran the decorator factory example. 

I don't get a `Callable` as a result in 3.7/3.8 (where I primarily tested), this only happens in 3.6. Is this another bug (that we should maybe tackle first)?

davidhalter:
I think I just fixed my original issue with the Python version differences. The fix was minor.

Now back again to the original problem :)

PeterJCLaw:
Ah, glad the `Callable` fix was that simple.

The symptom this PR is aiming to fix is the difference between the decorated function being inferred as merely `Callable` and being inferred as the original function (and thus having a useful name, parameters etc.). This is essentially the same symptom as was fixed in #1613, albeit for a slightly different definition of the decorator.

Does that help?

davidhalter:
Well I kind of understand what you're trying to solve, I just don't really understand why this doesn't already work, but that's really something where I have to dig into the code :).

davidhalter:
After a long time of research and thinking about this, I finally found an alternative (my gut always told me this should be simple):

```
diff --git a/jedi/inference/gradual/type_var.py b/jedi/inference/gradual/type_var.py
index 502eaf1f..1af5870e 100644
--- a/jedi/inference/gradual/type_var.py
+++ b/jedi/inference/gradual/type_var.py
@@ -102,7 +102,7 @@ class TypeVar(BaseTypingValue):
         else:
             if found:
                 return found
-        return self._get_classes() or ValueSet({self})
+        return ValueSet({self})
 
     def execute_annotation(self):
         return self._get_classes().execute_annotation()
```

While you definitely have a point with wrapping the upper bound, we simply don't have to get rid of TypeVars in `define_generics`. This is very similar to your solution, where the TypeVar nature is preserved.

What do you think?

PeterJCLaw:
> While you definitely have a point with wrapping the upper bound, we simply don't have to get rid of TypeVars in `define_generics`. This is very similar to your solution, where the TypeVar nature is preserved.
> 
> What do you think?

Huh, yeah, that's quite a lot simpler. I'm not sure why I didn't try this first and it passes all the tests I'd got, so lets go with that :+1: .

I've rebased to pick up the `Callable` fix and switched to the simpler solution.

davidhalter:
No worries, it only seems like an obvious fix in hindsight. It took me a long time (hours) to come up with this.

PR #1618, opened 2020-06-20
Add Python 3.8 to the tox env list

PeterJCLaw:
Since Python 3.8 is tested on CI, we should have it in the tox.ini too.

Should we have 3.9 there as well?


davidhalter:
> Should we have 3.9 there as well?

I wouldn't say no to that. I just hope that it's easy enough to get it working (it's not installed by default).

Issue #1617, opened 2020-06-19
Jedi has wrong sys.path when target environment has non-standard library location


mrclary:
@davidhalter,
Okay, when I submitted #1540, the problem arose when the host environment was a pyenv environment and the target environment was a conda environment. The solution was that the target environment should not inherit system environment variables from the host environment, i.e. the `CompiledSubprocess` should spawn a clean environment. Well, this works as it should when the target environment python libraries are in a standard relative location, i.e. `../lib` relative to the the executable. Unfortunately, this does not work for target environments where the python libraries are not in a standard location, such as for python embedded in MacOS applications (see [Spyder Mac App](https://github.com/spyder-ide/mac-application))

Python provides a mechanism for handling this, namely by setting the `PYTHONHOME` environment variable to the base directory of the libraries. This, along with other potential `PYTHON...` environment variables, were causing the problem outlined in #1540. Essentially, the wrong python environment variables were being inherited by the subprocess and mucking things up. In the absence of `PYTHONHOME`, it will find the libraries in the standard location.

So, I still think that starting a python subprocess with a clean environment is the correct paradigm, I now think that the solution provided for #1540 in #1546 is incomplete and should include a mechanism for passing explicit environment variables to the subprocess. `Popen` already does this, of course, with the `env` keyword arg. I propose modifying `CompiledSubprocess` to accept an `env` keyword argument that will be passed to the subprocess. The default should be an empty dictionary, but the user has the option to pass in envrionment variables explicitly, if needed.

What are your thougthts?

davidhalter:
I guess this is a good proposal. Since this is not very important to me, are you willing to do this in a PR?

I would probably not name it `env`, but `environment_variables` instead. But that's really just a minor detail.

mrclary:
I'll be happy to submit the PR.
I suggested `env` only because that is the keyword used by `Popen`.
I'm not a fan of long names; how about `env_vars`?

davidhalter:
That's fine.

davidhalter:
Fixed by #1619

PR #1619, opened 2020-06-20
Provide option to pass explicit environment variables to Environment and CompiledSubprocess

mrclary:
`subprocess.Popen` permits passing explicit environment variables to the subprocess. This PR permits this option for `Environment` and `CompiledSubprocess`.

Addresses issue #1617

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -61,8 +61,9 @@ class Environment(_BaseEnvironment):
     """
     _subprocess = None
 
-    def __init__(self, executable):
+    def __init__(self, executable, env_vars={}):
         self._start_executable = executable
+        self._env_vars = env_vars
         # Initialize the environment
         self._get_subprocess()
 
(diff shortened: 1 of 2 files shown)


jedi/api/environment.py
```diff
@@ -363,18 +366,18 @@ def get_system_environment(version):
     raise InvalidPythonEnvironment("Cannot find executable python%s." % version)
 
 
-def create_environment(path, safe=True):
+def create_environment(path, env_vars={}, safe=True):
```
davidhalter, line 369: One more small thing: can you create it like this (the current implementation would break the API BTW:

```
def create_environment(path, safe=True, **kwargs):
    return _create_environment(path, safe, **kwargs)  # TODO Python 2...


def _create_environment(path, safe=True, env_vars={}):
```

I want to make env_vars a keyword argument when I drop Python 2 (very soon).

jedi/inference/compiled/subprocess/__init__.py
```diff
@@ -156,11 +156,21 @@ class CompiledSubprocess(object):
     # Start with 2, gets set after _get_info.
     _pickle_protocol = 2
 
-    def __init__(self, executable):
+    def __init__(self, executable, env_vars={}):
         self._executable = executable
+        self._env_vars = env_vars
```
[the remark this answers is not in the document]
mrclary: Done. Thank you for the feedback. I went ahead and added kwargs for `find_system_environments` and `get_system_environments` so that the possibility is there without breaking the API.

jedi/api/environment.py
mrclary: Done. Thanks for the feedback.

davidhalter:
Thanks!

PR #1630, opened 2020-07-06
Includes PYTHON specific environment variables

### LINKED ISSUE #1629: Jedi should include environment variables that start with PYTHON for subprocess
According to [python's official document](https://docs.python.org/3/using/cmdline.html#environment-variables), there are some environment variables that can influence python's behavior.

I have my custom path for python modules, so on Windows, I run `set PYTHONPATH=<my custom path>` before executing python. Since `_get_process` in jedi\inference\compiled\subprocess\__init__.py uses explicit environment variables `env` that's basically empty (see the code [here](https://github.com/davidhalter/jedi/blob/a3a9ae1a262178f38ae80bd9ee4cb351893fff6b/jedi/inference/compiled/subprocess/__init__.py#L165)), the subprocess does not inherit the PYTHONPATH.

So I suggest jedi include those PYTHON* environment variables to subprocess in order for this to work.

Related issue:
#1540

wingkinl:
fix #1629, environment variables that begin with PYTHON should also go to subprocess

### DIFF
--- a/jedi/inference/compiled/subprocess/__init__.py
+++ b/jedi/inference/compiled/subprocess/__init__.py
@@ -167,9 +167,9 @@ class CompiledSubprocess(object):
             # if SYSTEMROOT (or case variant) exists in environment,
             # ensure it goes to subprocess
             for k, v in os.environ.items():
-                if 'SYSTEMROOT' == k.upper():
-                    self._env_vars.update({k: os.environ[k]})
-                    break  # don't risk multiple entries
+                # no need to worry about multiple entries since Windows ignores case
+                if 'SYSTEMROOT' == k.upper() or k.upper().startswith('PYTHON'):
+                    self._env_vars.update({k: v})
 
     def __repr__(self):
         pid = os.getpid()


davidhalter:
@mrclary You did this in #1619. What do you think about this change?

mrclary:
Well, I have a few concerns.

1. Grabbing environment variables from `os.environ` defeats the purpose of a default clean environment for subprocesses. Indeed, one of the motivations for passing clean environments is that `PYTHON*` system variables can cause unexpected behavior, especially when the host and client environments are different. For windows platforms, `SYSTEMROOT` _must_ be present, so there is an exception for that. The purpose of the `env_vars` keyword argument of `CompiledSubprocess` is to provide the option to explicitly pass environment variables which may be desired for any particular application. This PR would also unwittingly override an explicitly passed `PYTHON*` variable with that which is in `os.environ`.
1. This PR would only affect Windows platforms, so there would be different behavior for *nix platforms. I don't know how much of an issue this would be.

My recommendation is capture your `PYTHON*` variables outside of `CompiledSubprocess`, rather than impose the behavior on all Windows users. 
```python
my_env_vars = {}
for k, v in os.environ.items():
    if k.upper().startswith('PYTHON'):
        my_env_vars.update({k: v})

CompiledSubprocess(executable, env_vars=my_env_vars)
```

All that being said, if Windows environment variables are not case sensitive, then I gladly appreciate the simplification in that regard.

mrclary:
Well, I'm rethinking my comment. Perhaps the best thing to do is to adhere to the behavior of `Popen` as much as possible. `Popen` inherits `os.environ` by default, but accepts `env` keyword argument for explicit environments. Any user desiring a clean environment should explicitly pass `env={}` and Windows users should then shoulder the responsibility of including `SYSTEMROOT` (since `Popen` makes no such accommodation either)

Let me modify this PR with my updated recommendation

mrclary:
@wingkinl @davidhalter , take a look at #1633. Will this do what you need?

wingkinl:
> @wingkinl @davidhalter , take a look at #1633. Will this do what you need?

It works great! Thanks.

davidhalter:
@mrclary How about having the default of `env=None` instead and inherit the environment variables in that case and if somebody sets `env={}` it will be "clean"?

mrclary:
> @mrclary How about having the default of `env=None` instead and inherit the environment variables in that case and if somebody sets `env={}` it will be "clean"?
> 
> I don't want to provide full access to all popen arguments.
No problem.

> It could also be interesting to only give access to a few selected variables by default (SYSTEMROOT, PYTHONPATH and maybe one or two more).
Hmm. Users may expect default `Popen` behavior if `env=None`, i.e. inherit all `os.environ` variables.

davidhalter:
@wingkinl I resolved this with the help of @mrclary in #1633. Now all environment variables are again available by default. If someone provides a dictionary the environment variables will be overwritten.

Thanks for looking into this and trying to find a solution!

Issue #1627, opened 2020-07-02
jedi latest version runtime error for Windows


jdhao:
I am using deoplete-jedi which in turn depends on Jedi. It seems that the latest version of Jedi or master branch of Jedi will error out for Windows.

Here is the MWE to reproduce this issue:

```
import jedi

source1 = '''
import os
os.
'''

script1 = jedi.Script(source1, 3, len('os.'), 'example1.py')
print(script1.completions())
```

Here is the test summary for different combination of platform and jedi version:

| platform | jedi version | error out? |
|----------|--------------|------------|
| Windows  | 0.16.0       | No         |
| Windows  | 0.17.0       | Yes        |
| Linux    | 0.16.0       | No         |
| Linux    | 0.17.0       | No         |

On both Linux and Windows, Python is provided by Anaconda and the version is 3.6.5.

Here is the traceback when error occurs:

```
Traceback (most recent call last):
  File "D:\Anaconda\lib\site-packages\jedi\inference\compiled\subprocess\__init__.py", line 261, in _send
    is_exception, traceback, result = pickle_load(self._get_process().stdout)
  File "D:\Anaconda\lib\site-packages\jedi\_compatibility.py", line 396, in pickle_load
    return pickle.load(file, encoding='bytes')
EOFError: Ran out of input

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\Anaconda\lib\site-packages\jedi\api\environment.py", line 75, in _get_subprocess
    info = self._subprocess._send(None, _get_info)
  File "D:\Anaconda\lib\site-packages\jedi\inference\compiled\subprocess\__init__.py", line 273, in _send
    stderr,
jedi.api.exceptions.InternalError: The subprocess D:\Anaconda\python.exe has crashed (EOFError('Ran out of input',), stderr=).

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test_jedi.py", line 8, in <module>
    script1 = jedi.Script(source1, 3, len('os.'), 'example1.py')
  File "D:\Anaconda\lib\site-packages\jedi\api\__init__.py", line 185, in __init__
    project, environment=environment, script_path=self.path
  File "D:\Anaconda\lib\site-packages\jedi\inference\__init__.py", line 90, in __init__
    self.compiled_subprocess = environment.get_inference_state_subprocess(self)
  File "D:\Anaconda\lib\site-packages\jedi\api\environment.py", line 114, in get_inference_state_subprocess
[...]

PR #1633, opened 2020-07-08
Fix for #1630

mrclary:
See #1630

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -61,7 +61,7 @@ class Environment(_BaseEnvironment):
     """
     _subprocess = None
 
-    def __init__(self, executable, env_vars={}):
+    def __init__(self, executable, env_vars=None):
         self._start_executable = executable
         self._env_vars = env_vars
         # Initialize the environment
@@ -136,7 +136,7 @@ class _SameEnvironmentMixin(object):
         self._start_executable = self.executable = sys.executable
         self.path = sys.prefix
         self.version_info = _VersionInfo(*sys.version_info[:3])
-        self._env_vars = {}
+        self._env_vars = None
 
 
 class SameEnvironment(_SameEnvironmentMixin, Environment):
@@ -379,7 +379,7 @@ def create_environment(path, safe=True, **kwargs):
     return _create_environment(path, safe, **kwargs)
 
 
-def _create_environment(path, safe=True, env_vars={}):
+def _create_environment(path, safe=True, env_vars=None):
     if os.path.isfile(path):
         _assert_safe(path, safe)
         return Environment(path, env_vars=env_vars)
--- a/jedi/inference/compiled/subprocess/__init__.py
+++ b/jedi/inference/compiled/subprocess/__init__.py
@@ -156,20 +156,11 @@ class CompiledSubprocess(object):
     # Start with 2, gets set after _get_info.
     _pickle_protocol = 2
 
-    def __init__(self, executable, env_vars={}):
+    def __init__(self, executable, env_vars=None):
         self._executable = executable
-        self._env_vars = dict(env_vars)
+        self._env_vars = env_vars
         self._inference_state_deletion_queue = queue.deque()
         self._cleanup_callable = lambda: None
-
-        # Use explicit envionment to ensure reliable results (#1540)
-        if os.name == 'nt':
-            # if SYSTEMROOT (or case variant) exists in environment,
-            # ensure it goes to subprocess
-            for k, v in os.environ.items():
-                if 'SYSTEMROOT' == k.upper():
-                    self._env_vars.update({k: os.environ[k]})
-                    break  # don't risk multiple entries
 
     def __repr__(self):
         pid = os.getpid()


jedi/inference/compiled/subprocess/__init__.py
```diff
@@ -156,20 +156,19 @@ class CompiledSubprocess(object):
...
         self._executable = executable
-        self._env_vars = dict(env_vars)
         self._inference_state_deletion_queue = queue.deque()
         self._cleanup_callable = lambda: None
 
-        # Use explicit envionment to ensure reliable results (#1540)
-        if os.name == 'nt':
-            # if SYSTEMROOT (or case variant) exists in environment,
-            # ensure it goes to subprocess
+        if env_vars is None:
```
davidhalter, line 164: Wouldn't `self._env_vars = None` be good enough?

mrclary: Sorry, I don't follow.

mrclary: I thought you wanted specific default environment variables instead of inheriting all of `os.environ`.

davidhalter: Yeah, I'm not sure about that. Need a moment to think about it :) Sorry no action needed right now, I'm much more interested in the discussion.

The question is probably why we make this kind of isolation, when it is not providing a big upside (all variables that are relevant are provided by default).

mrclary: So maybe a recap of what happened will help us.

1. I originally discovered an issue (#1540) where jedi was providing incorrect completions when working with different host and client environments. In particular, inherited environment variables such as `PYTHONHOME`, or `*PYENV*` could cause incorrect `sys.path` for the client. Naturally, I thought that a clean environment would be best. #1546  then changed Jedi's behavior from passing the host `os.environ` to the client (`Popen`'s default behavior) to forcing a clean environment.
1. Except that turned out to cause other issues (#1617) where some circumstances _required_ a specific environment variable (`PYTHONHOME`), _but not necessarily_ inherited from the host `os.environ`. Naturally, I thought that we just need to allow the user to explicitly pass environment variables. #1619 then changed Jedi's behavior from forcing a clean environment to setting a clean environment by default with an option for the user to pass explicit environment variables.
1. Then #1629 revealed my mistake. I should have simply provided the option to explicitly pass environment variables to `Popen`. This would have preserved `Popen`'s (and Jedi's) original behavior as default (client inherits host `os.environ`) and resolved #1540, in which case #1690 and #1629 would not have arisen. In my defense, however, this solution was not possible for me originally since I am working with Jedi through [python-language-server](https://github.com/palantir/python-language-server) which does not provide the option to pass through environment variables to Jedi `Environment`s. This will be remedied by palantir/python-language-server#822.

So, bottom line (IMHO)...just reflect `Popen` behavior: client environments will inherit `os.environ` by default or optionally take user provided environment variables.
- Restores Jedi's behavior pre 0.17.1 and fixes #1629 (and precludes any other similar issues that my exploits may have caused :wink:)
- Resolves issues such as #1540 and #1617 by allowing the user to pass explicit environment variables to the client if necessary.

Also, I think that you are correct that `self._env_vars = None` should be sufficient. If the default behavior is to pass only specific variables from the host `os.environ`, we may be inviting issues similar to #1629.

mrclary: @davidhalter , so should I just set `self._env_vars = None`?

davidhalter: Sorry I'm a bit behind PR's issue. Will answer soon.

davidhalter: Thanks for the good summary. That helps indeed a lot! For #1540, what variables did you try to hide? Or what variables needed to be removed from the environment to be have it perform correctly?

mrclary: For #1540, specifically, it was `PYENV_VERSION`; but related to Jedi I also had issues with `PYTHONHOME`.
However, I've had many experiences in the past where unexplained behavior or a bug is ultimately traced back to unwittingly passing bizarre environment variables to subprocess. These can be particularly malicious since one can't easily reproduce them by duplicating the platform and conda/pip environment (what _else_ is installed?).

davidhalter: Hmm ok. I'm probably in favor of just passing `self._env_vars = None` in the default case. What is your favorite solution?

I think this makes the "normal" case very easy where you want to load the same environment again that you are already running. Since I'm going to be dropping Python 2, there's not really a lot of reasons anymore to load Python versions other than the current one. For people using environment variables (probably a minority, I think it's a terrible hack to use them), it's important that at least the default case works. If a virtualenv doesn't load properly, they will have to dig into that themselves and I don't mind that as much. Your issue was probably the first one where a clean env helped.

There were also a few issues now (#1627 and #1629) that showed that people are annoyed with the current behavior.

Passing just a part of the variables is probably just more confusing than it needs to be. If I'm honest I would probably also like to not pass any environment variables to have a clean environment and let the people figure out their issues themselves. But the reality is that a lot of people are going to come here and complain and I'd rather not have that.

mrclary: I agree, completely. Again, I apologize for the trouble this has caused.

mrclary: Appropriate changes pushed.

mrclary:
I don't know why this PR is not updating with my pushed changes...
Created #1635.

jedi/api/environment.py
```diff
@@ -136,7 +136,7 @@ def __init__(self):
         self._start_executable = self.executable = sys.executable
         self.path = sys.prefix
         self.version_info = _VersionInfo(*sys.version_info[:3])
-        self._env_vars = {}
+        self._env_vars = None
 
 
 class SameEnvironment(_SameEnvironmentMixin, Environment):
```
davidhalter, line 142: You seem to have missed the default on line 382 (`_create_environment`).

mrclary: Thanks for catching that! Correction pushed up.

davidhalter:
Thanks again for looking into that and being patient!

PR #1638, opened 2020-07-21
Update flake8

PeterJCLaw:
This is a precursor to adding type annotations; the latest flake8 has much improved detection of types annotations in various contexts and adds handling for tagged `# type: ignore[reason]` comments.

### DIFF
--- a/test/test_api/test_call_signatures.py
+++ b/test/test_api/test_call_signatures.py
@@ -121,11 +121,8 @@ def test_multiple_signatures(Script):
 
 
 def test_get_signatures_whitespace(Script):
-    s = dedent("""\
-    abs(
-    def x():
-        pass
-    """)
+    # note: trailing space after 'abs'
+    s = 'abs( \ndef x():\n    pass\n'
     assert_signature(Script, s, 'abs', 0, line=1, column=5)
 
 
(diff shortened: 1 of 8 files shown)


test/test_api/test_call_signatures.py
```diff
@@ -122,10 +122,10 @@ def f(a, b):
 
 def test_get_signatures_whitespace(Script):
     s = dedent("""\
-    abs( 
+    abs(
```
davidhalter, line 125: This change should not be removed. It's intentional. It's also why the tests are failing :)

davidhalter:
The rest is obviously fine and I'd be happy to merge this!

test/test_api/test_call_signatures.py
PeterJCLaw: Ah.
I tend to have my editor strip trailing spaces (and spotted a few places where there were unexpected trailing spaces in jedi, which the `.editorconfig` file aims to help prevent), and suspect others might do too.

I realise it's a little less clear, but I wonder if this would be more robust if spelled as a single line string with explicit `\n`. Would you be ok with that?

davidhalter: I'm fine with that. I generally think that automatically stripping of whitespace is a bad idea for Jedi in string literals. This has to do with a fact that it's pretty much necessary to test this kind of stuff.

PeterJCLaw:
Hrm, is `{posargs:jedi}` a tox thing? Should that change to just `jedi` on Travis?

davidhalter:
Yes, that's a tox thing. Probably my fault. Sorry for that.

davidhalter:
Thanks!

PR #1641, opened 2020-07-24
Python 3.6 embeddable doesn't have pydoc_data

PeterJCLaw:
This reinstates the import check for pydoc_data for now.

Specifically I looked in the following:
- [python-3.6.8-embed-amd64.zip](https://www.python.org/ftp/python/3.6.8/python-3.6.8-embed-amd64.zip): missing pydoc_data
- [python-3.7.8-embed-amd64.zip](https://www.python.org/ftp/python/3.7.8/python-3.7.8-embed-amd64.zip): present
- [python-3.8.5-embed-amd64.zip](https://www.python.org/ftp/python/3.8.5/python-3.8.5-embed-amd64.zip): present

Found while adding types; I've not actually tested this beyond inspecting the zips.

### DIFF
--- a/jedi/api/keywords.py
+++ b/jedi/api/keywords.py
@@ -3,7 +3,11 @@ from contextlib import suppress
 
 from jedi.inference.names import AbstractArbitraryName
 
-from pydoc_data import topics as pydoc_topics
+try:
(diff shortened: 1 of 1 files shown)


davidhalter:
Makes sense, thanks!

PR #1646, opened 2020-08-01
Turn print into warning to simplify silencing them.

Carreau:
There are other places, but I want to know whether you would be ok for that...

### DIFF
--- a/jedi/inference/compiled/access.py
+++ b/jedi/inference/compiled/access.py
@@ -1,5 +1,6 @@
 import inspect
 import types
+import traceback
 import sys
 import operator as op
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


davidhalter:
I'm fine with this.

It's pretty hard to debug, because it's usually part of the subprocess (except for InterpreterEnvironment of course), but I still think that these prints are pretty bad for some people, so I'm fine if you use logs there.

davidhalter:
Feel free to push further changes. I want to fix the CI first before I merge this.

Carreau:
actually there was no other places, but I re formated the code, added stack level and move import to the top. I just forgot to install in my env, so I was still seeing some error messages.

davidhalter:
Turns out it's pretty hard to fix the CI for now (because I seem to have a slightly different Python 3.9, which is still beta).

So I'm choosing to not fix it for now and just merge this.

davidhalter:
BTW: You had a Django-stubs submodule change in there which I just reverted. Not sure why I did not see that.

Carreau:
Oh sorry about that !

PR #1662, opened 2020-08-29
Fix #1655 by using platform-correct path handling for relative imports

### LINKED ISSUE #1655: Specifc relative import statement causing issues with VSC highlighting and 'jumping' functionality.
Hi! This is an issue that I've forwarded over from https://github.com/microsoft/vscode-python/issues/13198, following advice. For the full discussion so far, please check the issue above. 

## Environment data

-   VS Code version: 1.47.3 x64 , although also tested on 1.48.0 Insider
-   Extension version (available under the Extensions sidebar): v2020.7.96456
-   OS and version: Windows 10 Pro Verision 1909 64-bit
-   Python version (& distribution if applicable, e.g. Anaconda): Python 3.8.5 64-bit
-   Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A 
-   Relevant/affected Python packages and their versions: N/A
-   Relevant/affected Python-related VS Code extensions and their versions: N/A
-   Value of the `python.languageServer` setting: Jedi

## Steps to reproduce:

Example Directory: 
[VSC_behaviour.zip](https://github.com/microsoft/vscode-python/files/4998136/VSC_behaviour.zip)

1. Have the 'Python' extension installed.
2. Download the Example Directory above, unzip, and open it with Visual Studio Code. You should arrive at this tree:
3. This is the set-up for the testing for 'Actual behaviour' and 'Expected behaviour'.
```
project
├── directory1
│   ├── __init__.py
│   ├── file2.py
│   └── file3.py
├── file1.py
├── run.py
```

## Expected behaviour
I will describe the highlighting as 'blue-highlighting', as that's what's shown in my screenshots due to my theme, but presumably, there will be different colours for others.

1.1. Navigate to `run.py`. Assuming default keybindings, hover your mouse over the `file1` part of `import file1`. It should highlight and underline in blue, shown below, and you should be able to left-click and 'jump' to `file1`. 
image3
1.2. Type `file1.` inside `run.py`, as if you were calling a function from `file1`. You should be auto-suggested the `do_nothing()` function, the behaviour of which I believe comes from the Python extension.
image4

2. Now navigate to `file2.py`. Notice the `from . import file3`. Hovering your mouse over file3 should result in highlighting/underlining/jumping (but it doesn't - see Actual Behaviour). Moreover, writing `file3.` in `file2.py` should suggest the `do_nothing()` function. 

3. Replacing `from . import file3` in `file2.py` with an alternative syntax should also have blue-highlighting/underlining/jumping/function suggestions. For instance, `from .file3 import *` should have this behaviour, as well as `import file3`, although the latter throws an editor error, but still has that highlighting/underlining and 'jumping'.

## Actual behaviour

Steps 1.1 and 1.2 show the expected behaviour (see images above). There is the expected blue-highlighting/underlining/jumping/function suggestions.

However, Step 2 does not show the expected behaviour. There is no blue-highlighting/underlining/jumping/function suggestions.
image6
Perhaps this behaviour is due to my syntax, but my project appears to run fine with this step 2 syntax. This seems frustrating to me, as this syntax seems to be 'legal'. Furthermore, if I manually write `file3.do_nothing()`, the function can run perfectly normally at execution, provided the function is ultimately called from `run.py`, the Python file that's ran for the project. You can see an example of this expected behaviour here. You should see the text 'I'm working!' on terminal. 
[VSC_Behaviour_run.zip](https://github.com/microsoft/vscode-python/files/4998174/VSC_Behaviour_run.zip)

One possible theory I have is that the highlighting/etc. is being missed out due to the nature of the relative import. However, the highlighting/etc. work for another relative imports method, as shown in Step 3 below. Hence, I think that this specific relative import syntax could be causing conflicts with the Python extension for some reason, but I'm not sure why this might be occurring. 

As mentioned above, the alternative import syntaxes in Step 3 do show the expected behaviour. There is highlighting/etc.
image5

Furthermore, within `__init__.py` specifically, I found Step 2 to indeed meet the expected behaviour. Writing: `from . import file3` showed the blue-highlighting/underlining/jumping/function suggestions. This makes me confused. Now this syntax apparently does show support, but only within that `__init__.py` file, not within `file2.py`, for instance.
image7

However, as I mentioned above, the relative import that's missing that highlighting/etc. still runs in the project with expected output, with the relative imports working, as can be seen in VSC_Behaviour_run.zip above.
I was recommended to post the issue here, just to let `jedi`, but apparently it could be due to the Python extension's use of the Jedi API. The original issue is here: https://github.com/microsoft/vscode-python/issues/13198

Thank you!

PeterJCLaw:
This fixes an issue which showed up on Windows in 0.17.x where the VSCode Python extension passes paths which have their drive letter normalised to lower case. Since paths are (mostly) case insensitive on Windows, that shouldn't matter. However Jedi's handling was `==` on `str`s, meaning that completions based on relative imports on windows didn't work.

On master this same issue had started appearing due to the partial migration to use `Path` instead of `str` (which will avoid such issues entirely once complete), as `Path` instances were being compared to `str` instances without first converting.

We fix both by clarifying which is used here and then doing the proper conversions.

### DIFF
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -10,6 +10,7 @@ statements like ``from datetim`` (cursor at the end would return ``datetime``).
 """
 import os
 from pathlib import Path
+from typing import List, Optional, Tuple
 
 from parso.python import tree
 from parso.tree import search_ancestor
@@ -123,7 +124,11 @@ def _add_error(value, name, message):
         debug.warning('ImportError without origin: ' + message)
 
 
-def _level_to_base_import_path(project_path, directory, level):
+def _level_to_base_import_path(
+    project_path: Path,
+    directory: Path,
+    level: int,
+) -> Tuple[Optional[List[str]], Optional[str]]:
     """
     In case the level is outside of the currently known package (something like
     import .....foo), we can still try our best to help the user for
@@ -131,23 +136,23 @@ def _level_to_base_import_path(project_path, directory, level):
     """
     for i in range(level - 1):
         old = directory
-        directory = os.path.dirname(directory)
+        directory = directory.parent
         if old == directory:
             return None, None
 
     d = directory
-    level_import_paths = []
+    level_import_paths: List[str] = []
     # Now that we are on the level that the user wants to be, calculate the
     # import path for it.
     while True:
         if d == project_path:
-            return level_import_paths, d
+            return level_import_paths, str(d)
         dir_name = os.path.basename(d)
         if dir_name:
             level_import_paths.insert(0, dir_name)
-            d = os.path.dirname(d)
+            d = d.parent
         else:
-            return None, directory
+            return None, str(directory)
 
 
 class Importer:
@@ -202,7 +207,7 @@ class Importer:
                     directory = os.path.dirname(path)
 
                 base_import_path, base_directory = _level_to_base_import_path(
-                    project_path, directory, level,
+                    project_path, Path(directory), level,
                 )
                 if base_directory is None:
                     # Everything is lost, the relative import does point
--- a/test/examples/issue1655/subdir/file3.py
+++ b/test/examples/issue1655/subdir/file3.py
@@ -1,0 +1,2 @@
+def do_nothing_file3():
+    pass
--- a/test/test_inference/test_imports.py
+++ b/test/test_inference/test_imports.py
@@ -544,6 +544,10 @@ def test_level_to_import_path(
     project_path: str,
     result: Tuple[Optional[List[str]], Optional[str]],
 ) -> None:
+    # Convert for path separators being different on Windows
+    subdirs, root = result
+    result = subdirs, None if root is None else str(Path(root))
+
     assert imports._level_to_base_import_path(
         Path(project_path),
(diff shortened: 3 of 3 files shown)


davidhalter:
Hey Peter
Sorry for not looking into this sooner. I was in my honeymoon for a few weeks :) I'm pretty positive on merging this, but I want to do some more research, because I'm not sure if I detected an issue on Linux as well. So just give me a bit of time.

I had the time to fix the CI at least. So at least that should work again.

PeterJCLaw:
> Sorry for not looking into this sooner. I was in my honeymoon for a few weeks :)

No worries; congrats!

> I'm pretty positive on merging this, but I want to do some more research, because I'm not sure if I detected an issue on Linux as well. So just give me a bit of time.
> 
> I had the time to fix the CI at least. So at least that should work again.

:+1:

PeterJCLaw:
While I suspect we're going to want some of these changes for a fuller move to using `Path`s, I don't think we want this PR as-is any longer (at least not now the issue it was aiming to solve is closed).

PR #1663, opened 2020-08-30
Tidyups

PeterJCLaw:
Various tidyups found while exploring adding type annotations.

Not sure where I got to with adding annotations, however on re-reviewing these changes they seem useful anyway.


davidhalter:
Thanks!

PR #1669, opened 2020-09-14
make contextualized_node an optional kwarg in ReversedObject

mvanderkamp:
In all other py__iter__ definitions that I found, this argument is optional. It also often seems to not be passed around. I'm not sure why it was deemed mandatory here despite not being used.


davidhalter:
Very good point, thanks!

PR #1684, opened 2020-10-19
Relative imports should work even if they are not within the project

davidhalter:
See comments in code.

This is another potential solution for #1655 (and an answer to #1662 as well).

@PeterJCLaw I was a bit confused in the beginning why this was happening, but I understand it better now. This was an issue on all platforms. We could of course still use #1662 or at least parts of it, but I guess this is the more important part.

### DIFF
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -252,7 +252,7 @@ class Importer:
                 # relative imports the wrong way, we might end up here, where
                 # the `fixed_sys_path == project.path` in that case we kind of
                 # use the project.path.parent directory as our path. This is
-                # usually not a problem, except if imorts in other places are
+                # usually not a problem, except if imports in other places are
                 # using the same names. Example:
                 #
                 # foo/                       < #1
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -267,6 +267,10 @@ class Importer:
                 # `__init__.py` (#2) and not as `foo.py` (#3). This is usually
                 # not an issue, because this case is probably pretty rare, but
                 # might be an issue for some people.
+                #
+                # However for most normal cases where we work with different
+                # file names, this code path hits where we basically change the
+                # project path to an ancestor of project path.
                 from jedi.inference.value.namespace import ImplicitNamespaceValue
                 import_path = (os.path.basename(self._fixed_sys_path[0]),)
                 ns = ImplicitNamespaceValue(
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -245,7 +245,38 @@ class Importer:
         )
 
     def follow(self):
-        if not self.import_path or not self._infer_possible:
+        if not self.import_path:
+            if self._fixed_sys_path:
+                # This is a bit of a special case, that maybe should be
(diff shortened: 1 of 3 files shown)


test/test_api/test_usages.py
```diff
@@ -3,7 +3,7 @@
 
 def test_import_references(Script):
```
PeterJCLaw, line 4: This test appears to be failing on CI.

jedi/inference/imports.py
```diff
@@ -245,7 +245,38 @@ def _sys_path_with_modifications(self, is_completion):
         )
 
     def follow(self):
-        if not self.import_path or not self._infer_possible:
+        if not self.import_path:
+            if self._fixed_sys_path:
+                # This is a bit of a special case, that maybe should be
```
PeterJCLaw, line 250: The description here of why the code is needed/is special feels out of step with the test which validates the behaviour. Perhaps I'm missing something, but is there more to this code than just the niche case described here?

PeterJCLaw (commented):
Looks like this improves some cases, though to be honest I'm not sure I completely follow how this code works (I've only had a fairly quick look).
I'll have another play with #1655 soon (which I think we're going to want the `str`/`Path` fixing from anyway) and have another look at this afterwards.

jedi/inference/imports.py
davidhalter: No it's simply that.

It's just a special case, because it breaks the basic idea of something being on the sys path.

test/test_api/test_usages.py
```diff
@@ -3,7 +3,7 @@
...
     s = Script("from .. import foo", path="foo.py")
-    assert [usage.line for usage in s.get_references(line=1, column=18)] == [1]
```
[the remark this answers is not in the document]
davidhalter: It's intentional.

(I just touched this, because there was an issue with the test when I refactored some things. After that it was not there anymore, but I still thought that the change made it more understandable - and functionally it's exactly the same).

test/test_api/test_usages.py
davidhalter: Test is passing again.

jedi/inference/imports.py
PeterJCLaw: Ah sorry. I think if this code were here just for the special case, which this comment suggests is to do with the same name appearing at several levels in the hierarchy, then on reading this comment alone I would assume that it would only be activated by that special case and that the test to validate it would exercise that case. However the test doesn't actually seem to hit that case (the directory names it chooses differ at each level), yet this code path is taken.

Clearly the test that is present is a useful test and this codepath is useful in making it work, so perhaps this comment could be tweaked to clarify that this codepath is here for all cases of namespace packages, rather than _just_ for the case of matching names?

davidhalter: I guess my comment just wasn't good enough. Because having the same names is just one case. Will try to improve :)

davidhalter: I added a bit to the comment.

Once I merge this, I will probably make a release.

davidhalter:
@PeterJCLaw I merged, because I think this was at least good enough. Please let me know if you think it's still not good enough and I will fix it for the next release.

PR #1704, opened 2020-12-05
add Completion.get_completion_prefix_length

### LINKED ISSUE #1687: How to figure out completion prefix without `completion.complete` in fuzzy mode?
Hi,

I'm trying to make IPython use fuzzy completion, however it doesn't work right now because `Completion.complete` [is None for fuzzy completions](https://github.com/davidhalter/jedi/commit/41a6591d88efcbf29832844e877b863a49d1d8ff).
From what I can see, IPython needs this field [to compute which prefix should be replaced](https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L1956) in the completion input. For example, if the current text is `os.ab` and there's an `abort(` completion, IPython needs to know that after accepting the completion the new text is `os.abort(` and not `os.ababort(` or something else.

Is there a way to get the same information with fuzzy completions?

Thanks

infokiller:
fixes #1687

I will add tests after we're good with the naming/API

### DIFF
--- a/jedi/api/classes.py
+++ b/jedi/api/classes.py
@@ -749,6 +749,24 @@ class Completion(BaseName):
 
         return super().type
 
+    def get_completion_prefix_length(self):
+        """
+        Returns the length of the prefix being completed.
+        For example, completing ``isinstance``::
+
+            isinstan# <-- Cursor is here
+
+        would return 8, because len('isinstan') == 8.
+
+        Assuming the following function definition::
+
+            def foo(param=0):
+                pass
+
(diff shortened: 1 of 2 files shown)


davidhalter:
Can you make it a method with `get_` and remove the `@property`? Otherwise LGTM.

davidhalter:
Let me know once you're finished.

infokiller:
Add tests, you can review now.

davidhalter:
Awesome, thanks!

PR #1706, opened 2020-12-06
Catch 'PermissionError' for unreadable directories

ColdGrub1384:
Hi. The master branch of Jedi raises a `PermissionError` on iOS. I added two try excepts on code that assumes a directory is readable. That may be useful on other platforms too. That was the error:

```Traceback (most recent call last):
  File "/private/var/containers/Bundle/Application/2FB8EF9C-46B5-4A69-8D4F-0D1115D2BED0/Pyto.app/Lib/_codecompletion.py", line 59, in suggestForCode
    script = jedi.Script(code, path=path)
  File "/private/var/containers/Bundle/Application/2FB8EF9C-46B5-4A69-8D4F-0D1115D2BED0/Pyto.app/site-packages/jedi/api/__init__.py", line 145, in __init__
    project = get_default_project(None if self.path is None else self.path.parent)
  File "/private/var/containers/Bundle/Application/2FB8EF9C-46B5-4A69-8D4F-0D1115D2BED0/Pyto.app/site-packages/jedi/api/project.py", line 422, in get_default_project
    if probable_path is None and _is_potential_project(dir):
  File "/private/var/containers/Bundle/Application/2FB8EF9C-46B5-4A69-8D4F-0D1115D2BED0/Pyto.app/site-packages/jedi/api/project.py", line 369, in _is_potential_project
    if path.joinpath(name).exists():
  File "/var/mobile/Containers/Data/Application/4A635FFF-9668-44D3-9A19-6D54C2D289FF/Library/python38/pathlib.py", line 1370, in exists
    self.stat()
  File "/var/mobile/Containers/Data/Application/4A635FFF-9668-44D3-9A19-6D54C2D289FF/Library/python38/pathlib.py", line 1176, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 1] Operation not permitted: '/private/var/mobile/Library/Mobile Documents/setup.py'
```

`/private/var/mobile/Library/Mobile Documents` is the directory where all iCloud Drive files are stored, but it's sandboxed so apps can only read their own container which are located in this directory.

Sorry, I don't think iOS is a priority here, but other computers with strict permissions could also raise this error so I think it would be better anyway to handle those cases.

### DIFF
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -366,8 +366,11 @@ class Project:
 
 def _is_potential_project(path):
     for name in _CONTAINS_POTENTIAL_PROJECT:
-        if path.joinpath(name).exists():
-            return True
+        try:
+            if path.joinpath(name).exists():
+                return True
+        except PermissionError:
+            continue
     return False
 
 
--- a/jedi/api/project.py
+++ b/jedi/api/project.py
@@ -369,7 +369,7 @@ def _is_potential_project(path):
         try:
             if path.joinpath(name).exists():
                 return True
-        except PermissionError:
+        except OSError:
             continue
     return False
 
--- a/jedi/inference/sys_path.py
+++ b/jedi/inference/sys_path.py
@@ -174,7 +174,7 @@ def _get_parent_dir_with_file(path: Path, filename):
         try:
             if parent.joinpath(filename).is_file():
                 return parent
-        except PermissionError:
+        except OSError:
             continue
     return None
 
--- a/jedi/inference/sys_path.py
+++ b/jedi/inference/sys_path.py
@@ -171,8 +171,11 @@ def _get_paths_from_buildout_script(inference_state, buildout_script_path):
 
 def _get_parent_dir_with_file(path: Path, filename):
     for parent in path.parents:
-        if parent.joinpath(filename).is_file():
-            return parent
+        try:
+            if parent.joinpath(filename).is_file():
+                return parent
+        except PermissionError:
+            continue
     return None
 
 
--- a/test/test_inference/test_sys_path.py
+++ b/test/test_inference/test_sys_path.py
@@ -108,3 +108,13 @@ def test_transform_path_to_dotted(sys_path_, module_path, expected, is_package):
     module_path = os.path.abspath(module_path)
     assert sys_path.transform_path_to_dotted(sys_path_, Path(module_path)) \
         == (expected, is_package)
+
+
+@pytest.mark.parametrize(
+    'path,filename,expected', [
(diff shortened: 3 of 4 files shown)


jedi/api/project.py
```diff
@@ -366,8 +366,11 @@ def __repr__(self):
 
 def _is_potential_project(path):
     for name in _CONTAINS_POTENTIAL_PROJECT:
-        if path.joinpath(name).exists():
-            return True
+        try:
+            if path.joinpath(name).exists():
+                return True
+        except PermissionError:
```
davidhalter, line 372: Just open for discussion: Shouldn't we use OSError here? I'm pretty sure that we do not care here if we can read it. If it's not readable, it should probably be ignored.

jedi/inference/sys_path.py
```diff
@@ -171,8 +171,11 @@ def _get_paths_from_buildout_script(inference_state, buildout_script_path):
 
 def _get_parent_dir_with_file(path: Path, filename):
     for parent in path.parents:
-        if parent.joinpath(filename).is_file():
-            return parent
+        try:
+            if parent.joinpath(filename).is_file():
+                return parent
+        except PermissionError:
+            continue
```
davidhalter, line 178: Same here.

davidhalter (commented):
Thanks for the patch, I guess it makes sense. If you write a test that's obviously a plus, but I'm fine here without them as well.

ColdGrub1384:
I just added some tests

jedi/api/project.py
ColdGrub1384: I just replaced `PermissionError` by `OSError`

test/test_inference/test_sys_path.py
```diff
@@ -108,3 +108,13 @@ def test_transform_path_to_dotted(sys_path_, module_path, expected, is_package):
...
+
+
+@pytest.mark.parametrize(
+    'path,filename,expected', [
+        (Path(__file__).parents[1], "setup.py", Path(__file__).parents[2]),
+        (Path(__file__).parents[2], os.path.basename(__file__), None)
+    ]
+)
+def test_get_parent_dir_with_file(path, filename, expected):
+    assert sys_path._get_parent_dir_with_file(path, filename) == expected
```
davidhalter, line 120: Since `os.path.basename(__file__),` might be part of a directory ancestor, we should not test like this. Most of the time this will not be an issue for people, but if it is it's very confusing.

IMO this function is tested well enough with other means and we can probably just omit this test.

ColdGrub1384: Sorry, I removed it

Issue #1702, opened 2020-12-05
KeyError in cache -> AttributeError -> UncaughtAttributeError 


Carreau:
Can give a minimal working example if needed but it might take me some time to  narrow down. Hoping the tb make sens on its own. (also can't figure out how to properly format that in the details tag.... so also in a [gist](https://gist.github.com/Carreau/6285fc4026f5ba8c756be2a9f16fe155).

For now I'm likely to just put in a try/except/pass.

Using latest master (69750b9b) 

<details>
<summary>Loooooong traceback</summary>
<pre>
Traceback (most recent call last):
  File "/Users/bussonniermatthias/dev/jedi/jedi/cache.py", line 110, in wrapper
    return dct[key]
KeyError: ((), frozenset())

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bussonniermatthias/dev/jedi/jedi/cache.py", line 110, in wrapper
    return dct[key]
KeyError: ((), frozenset())

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/utils.py", line 67, in wrapper
    return func(*args, **kwds)
  File "/Users/bussonniermatthias/dev/jedi/jedi/cache.py", line 112, in wrapper
    result = method(self, *args, **kwargs)
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/base_value.py", line 340, in _wrapped_value
    return self._get_wrapped_value()
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/value/iterable.py", line 203, in _get_wrapped_value
    TupleGenericManager(self._cached_generics())
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/cache.py", line 44, in wrapper
    rv = function(obj, *args, **kwargs)
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/value/iterable.py", line 195, in _cached_generics
    return self._get_generics()
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/value/iterable.py", line 335, in _get_generics
    return tuple(x.infer().py__class__() for x in self.py__iter__())
  File "/Users/bussonniermatthias/dev/jedi/jedi/inference/value/iterable.py", line 335, in <genexpr>
    return tuple(x.infer().py__class__() for x in self.py__iter__())
[...]

PR #1707, opened 2020-12-07
Add tests for #1702, for a rare numpydoc syntax.

Carreau:
It looks like numpydoc, and things like masked array docstrings use a
syntax that make jedi crash:

    fill_value : {var}, optional
            Value used internally for the masked values.
            If ``fill_value`` is not None, it supersedes ``endwith``.

Here we add a test that we do not crash jedi, and then a commit that only lookup children the node have a children attribute.


davidhalter:
Awesome, thanks!

PR #1734, opened 2021-01-20
Allow tweaking Interpreter sys_path

aivarannamaa:
Currently it is not easy to tweak `Interpreter`'s  sys_path. It is not clear how to make `InterpreterEnvironment` and giving a `project` argument gives duplicate named argument error.

I must say I liked the pre-0.18 approach better (with explicit `sys_path` parameter). I understand that the new approach is meant to be cleaner for common cases, but I'm trying to offer autocompletion for a MicroPython interpreter using module stubs in a directory unrelated to any CPython interpreter.

With the changes proposed in this PR I seem to be able to achieve my goal. I wouldn't mind the custom `InterpreterEnvironment` approach either, but I didn't understand how to make one and what else to take into account with this approach. If you added an API for making an `InterpreterEnvironment` with extra sys_path directories, then I would like this even more.

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -719,12 +719,12 @@ class Interpreter(Script):
         else:
             if not isinstance(environment, InterpreterEnvironment):
                 raise TypeError("The environment needs to be an InterpreterEnvironment subclass.")
-        
+
         if project is None:
             project = Project(Path.cwd())
 
         super().__init__(code, environment=environment, project=project, **kwds)
-            
+
         self.namespaces = namespaces
         self._inference_state.allow_descriptor_getattr = self._allow_descriptor_getattr_default
 
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -707,7 +707,7 @@ class Interpreter(Script):
     """
     _allow_descriptor_getattr_default = True
 
-    def __init__(self, code, namespaces, **kwds):
+    def __init__(self, code, namespaces, *, project=None, **kwds):
         try:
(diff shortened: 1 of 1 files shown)


davidhalter:
I agree that there's probably something wrong with `Interpreter`. I never really thought about that case, because I was always thinking about how to optimize the `Script` case.

> I must say I liked the pre-0.18 approach better (with explicit sys_path parameter). I understand that the new approach is meant to be cleaner for common cases, but I'm trying to offer autocompletion for a MicroPython interpreter using module stubs in a directory unrelated to any CPython interpreter.

I can see why this feels worse for you. However in most cases people are wrong in patching the sys path anyway and they should probably use a better code base. In your case this is probably not true, because you don't target CPython.

jedi/api/__init__.py
```diff
@@ -720,8 +720,12 @@ def __init__(self, code, namespaces, **kwds):
...
                 raise TypeError("The environment needs to be an InterpreterEnvironment subclass.")
 
-        super().__init__(code, environment=environment,
-                         project=Project(Path.cwd()), **kwds)
+        if "project" in kwds:
+            super().__init__(code, environment=environment, **kwds)
+        else:
+            super().__init__(code, environment=environment,
+                             project=Project(Path.cwd()), **kwds)
+
```
davidhalter, line 728: Just use the signature `__init__(self, code, namespaces, *, project=None, **kwds)` 

and use an 

```
if project is None:
    project = Project(Path.cwd())
```

I feel like this is cleaner.

davidhalter:
Thanks!

PR #1741, opened 2021-01-30
BUGFIX: endless loop in pytest plugin

sfavazza:
I use `elpy` in emacs 26.3 on Windows 10 Pro. One of the dependencies for the code completion is Jedi.

Recently I am working with projects involving the use of the `pytest` Python package and I observed a strange behavior that cause my editor to freeze.

I managed to attach a debugger to the Python process launched by `elpy` than I realized that the Jedi package got stuck in an endless loop, the section responsible for this issue is in `plugins/pytest.py::_iter_pytest_module`.

The infinite loop is caused by the search of a `conftest.py` file until the drive identifier letter (`C:` in my case). Then the search endlessly continues looking for the `C:/conftest.py`.

This MR propose a solution to this inconvenient by stopping the search once a parent is proposed twice.

### DIFF
--- a/jedi/plugins/pytest.py
+++ b/jedi/plugins/pytest.py
@@ -129,6 +129,10 @@ def _iter_pytest_modules(module_context, skip_own_module=False):
     if file_io is not None:
         folder = file_io.get_parent_folder()
         sys_path = module_context.inference_state.get_sys_path()
+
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks!

Issue #1791, opened 2021-07-22
get_signature() for functools.wraps()-decorated function contains parameter list of undecorated function


smheidrich:
Consider the following minimal reproducible example:

```python
from functools import wraps

def f(x: int, y: float):
  "docstring of f"
  pass

@wraps(f)
def g(*args, **kwargs):
  f(*args, **kwargs)

g()
```

We try to obtain and print the signature for the call to `g()` at the end via `get_signature()`:

```python
from jedi import Script

with open("example.py") as f:
  scr = Script(f.read())

sigs = scr.get_signatures(11, 2)

for sig in sigs:
  print(sig.to_string())
```

This yields `f(*args, **kwargs)`. So the function's name was correctly obtained from the wrapped function (`f` instead of `g`), but the parameter list is that of `g`. In my opinion, it would be better if the parameter list was also taken from the decorated function, as this is the behavior of e.g. Python's own `inspect.signature()` (which yields `<Signature (x: int, y: float)>` for `g`).

I tested this with Jedi 0.18 from PyPI as well as the current revision from the git repo (41652507b3de645a8a4329299e8e40bff68c8da7) and Python 3.9.2.

## Behavior in Jedi 0.15

This might be a regression compared to Jedi 0.15. Slightly modified version of the signature-obtaining code above:

```python
from jedi import Script

with open("example.py") as f:
  scr = Script(f.read(), line=11, column=2)

sigs = scr.call_signatures()

for sig in sigs:
  print(sig.to_string())
```

 After installing via `pip install jedi==0.15 parso==0.7`, this yields (correctly, in my opinion) `f(x: int, y: float)`.

smheidrich:
It works fine for something like this, which I guess would be more typical usage for `functools.wraps()`:

```python
from functools import wraps

def f(x: int, y: float):
  "docstring of f"
  pass

def do_wrap(func):
  @wraps(func)
  def g(*args, **kwargs):
    return func(*args, **kwargs)
  return g

g = do_wrap(f)
g()
```

But this gets the parameter list right even if you remove `functools.wraps`, i.e. just by looking at where `*args, **kwargs` are actually going. This should work for the original example above as well, shouldn't it?

davidhalter:
> . In my opinion, it would be better if the parameter list was also taken from the decorated function, as this is the behavior of e.g. Python's own inspect.signature() (which yields <Signature (x: int, y: float)> for g)

I'll consider it.

[...]

PR #1793, opened 2021-07-24
Fix module-scope passthrough function signatures

PeterJCLaw:
This ensures `*args, **kwargs` lookthrough works at module scope as well as at function scope, meaning that passthrough signatures will be found for top level functions.

Fixes https://github.com/davidhalter/jedi/issues/1791.

### DIFF
--- a/jedi/inference/star_args.py
+++ b/jedi/inference/star_args.py
@@ -22,7 +22,15 @@ def _iter_nodes_for_param(param_name):
     from jedi.inference.arguments import TreeArguments
 
     execution_context = param_name.parent_context
-    function_node = execution_context.tree_node
+    # Walk up the parso tree to get the FunctionNode we want. We use the parso
+    # tree rather than going via the execution context so that we're agnostic of
+    # the specific scope we're evaluating within (i.e: module or function,
+    # etc.).
+    # - .tree_name is a Name
+    # - .parent is a Param
+    # - .parent is a PythonNode(parameters)
+    # - .parent is the FunctionNode we want.
+    function_node = param_name.tree_name.parent.parent.parent
     module_node = function_node.get_root_node()
     start = function_node.children[-1].start_pos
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
Locally I'm seeing `test_import_references` failing both on this branch and on `master` (even on a fresh clone of the repo), yet it's not failing on `master` CI.

jedi/inference/star_args.py
```diff
@@ -22,7 +22,15 @@ def _iter_nodes_for_param(param_name):
...
-    function_node = execution_context.tree_node
+    # Walk up the parso tree to get the FunctionNode we want. We use the parso
+    # tree rather than going via the execution context so that we're agnostic of
+    # the specific scope we're evaluating within (i.e: module or function,
+    # etc.).
+    # - .tree_name is a Name
+    # - .parent is a Param
+    # - .parent is a PythonNode(parameters)
+    # - .parent is the FunctionNode we want.
+    function_node = param_name.tree_name.parent.parent.parent
```
davidhalter, line 33: Oh wow, that's the reason. IMO we should probably use `tree.search_ancestor(param_name.tree_name, 'funcdef', 'lambdef')`. That's probably more stable. If something ever changes (or maybe if lambdas are involved or something similar, some assumptions might break here.

Not sure about lambdas anyway.

davidhalter (commented):
Looks really good. I think this one thing should probably be changed.

jedi/inference/star_args.py
PeterJCLaw: `tree.search_ancestor` definitely sounds more reliable! I'll look at using that.

Yeah, previously the module-scope things were getting `Module` as their `function_node`.

PeterJCLaw: ab2eb570a81922546f7c870f2d3c34878a55f642

davidhalter:
Thanks!

PR #1794, opened 2021-07-25
Fix quoted generic annotations

PeterJCLaw:
It turns out that quoted annotations that contained generics weren't being handled. This PR changes the handling for quoted annotations so that they're resolved to tree nodes upfront rather than getting a different codepath later on.

This means that things like this will now have proper inference:
``` python
def foo(x: 'Tuple[T]') -> 'List[T]':
    return list(x)
```
Previously the value of `T` wouldn't have been handled, but now will be.


davidhalter:
Good catch!

PR #1795, opened 2021-08-05
inspect now raises OSError for objects without source file

frenzymadness:
This partially fixes compatibility with Python 3.10rc1. I've found this issue while debugging IPython, where the inspect module change caused a similar issue, see https://github.com/ipython/ipython/issues/13079 

CPython issue: https://bugs.python.org/issue44648

### DIFF
--- a/test/test_api/test_interpreter.py
+++ b/test/test_api/test_interpreter.py
@@ -711,3 +711,24 @@ def test_negate():
     assert x.name == 'int'
     value, = x._name.infer()
     assert value.get_safe_value() == -3
+
+
+def test_complete_not_findable_class_source():
+    class TestClass():
+        ta=1
+        ta1=2
+
+    # Simulate the environment where the class is defined in
+    # an interactive session and therefore inspect module
+    # cannot find its source code and raises OSError (Py 3.10+) or TypeError.
+    TestClass.__module__ = "__main__"
+    # There is a pytest __main__ module we have to remove temporarily.
+    module = sys.modules.pop("__main__")
+    try:
+        interpreter = jedi.Interpreter("TestClass.", [locals()])
+        completions = interpreter.complete(column=10, line=1)
+    finally:
+        sys.modules["__main__"] = module
+
+    assert "ta" in [c.name for c in completions]
+    assert "ta1" in [c.name for c in completions]
--- a/jedi/inference/compiled/mixed.py
+++ b/jedi/inference/compiled/mixed.py
@@ -187,7 +187,7 @@ def _find_syntax_node_name(inference_state, python_object):
     try:
         python_object = _get_object_to_check(python_object)
         path = inspect.getsourcefile(python_object)
-    except TypeError:
+    except (OSError, TypeError):
         # The type might not be known (e.g. class_with_dict.__weakref__)
         return None
     path = None if path is None else Path(path)


davidhalter:
Is there a way you can test and reproduce this? I don't like catching random errors.

frenzymadness:
I see that this piece of code is executed many times during the testing but the exception is not raised in any of them. Based on `test_all_completions_dups` in https://github.com/ipython/ipython/blob/770024afeaa5bd66910cf89b9667c4967b65004e/IPython/core/tests/test_completer.py#L436 and `_jedi_matches` method in https://github.com/ipython/ipython/blob/770024afeaa5bd66910cf89b9667c4967b65004e/IPython/core/completer.py#L1399 I've been able to create this simple reproducer.

I have to admit that I have no idea whether this is allowed usage from the jedi point of view or not but it seems to be the way how IPython uses it.

```
Python 3.10.0rc1 (default, Aug  3 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)] on linux
>>> import jedi
>>> 
>>> class TestClass():
...     ta=1
...     ta1=2
... 
>>> 
>>> interpteter = jedi.Interpreter("TestClass.", [globals()])
>>> interpteter.complete(column=10, line=1)
  File "/home/lbalhar/Software/jedi/jedi/cache.py", line 110, in wrapper
    return dct[key]
KeyError: ((), frozenset())

  File "/home/lbalhar/Software/jedi/jedi/api/helpers.py", line 487, in wrapper
    return func(self, line, column, *args, **kwargs)
  File "/home/lbalhar/Software/jedi/jedi/api/__init__.py", line 214, in complete
    return completion.complete()
  File "/home/lbalhar/Software/jedi/jedi/api/completion.py", line 170, in complete
    cached_name, completion_names = self._complete_python(leaf)
  File "/home/lbalhar/Software/jedi/jedi/api/completion.py", line 281, in _complete_python
    cached_name, n = self._complete_trailer(dot.get_previous_leaf())
  File "/home/lbalhar/Software/jedi/jedi/api/completion.py", line 382, in _complete_trailer
    values = infer_call_of_leaf(inferred_context, previous_leaf)
  File "/home/lbalhar/Software/jedi/jedi/inference/helpers.py", line 79, in infer_call_of_leaf
    return context.infer_node(leaf)
  File "/home/lbalhar/Software/jedi/jedi/inference/context.py", line 224, in infer_node
    return infer_node(self, node)
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 156, in infer_node
    return _infer_node_if_inferred(context, element)
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 169, in _infer_node_if_inferred
    return _infer_node_cached(context, element)
  File "/home/lbalhar/Software/jedi/jedi/inference/cache.py", line 44, in wrapper
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 174, in _infer_node_cached
    return _infer_node(context, element)
  File "/home/lbalhar/Software/jedi/jedi/debug.py", line 81, in wrapper
    return func(*args, **kwargs)
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 82, in wrapper
    return func(context, *args, **kwargs)
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 184, in _infer_node
    return infer_atom(context, element)
  File "/home/lbalhar/Software/jedi/jedi/inference/syntax_tree.py", line 304, in infer_atom
    return context.py__getattribute__(atom, position=position)
  File "/home/lbalhar/Software/jedi/jedi/inference/context.py", line 77, in py__getattribute__
    values = ValueSet.from_sets(name.infer() for name in names)
  File "/home/lbalhar/Software/jedi/jedi/inference/base_value.py", line 426, in from_sets
    for set_ in sets:
  File "/home/lbalhar/Software/jedi/jedi/inference/context.py", line 77, in <genexpr>
    values = ValueSet.from_sets(name.infer() for name in names)
  File "/home/lbalhar/Software/jedi/jedi/cache.py", line 112, in wrapper
  File "/home/lbalhar/Software/jedi/jedi/inference/compiled/mixed.py", line 137, in infer
    return _create(self._inference_state, compiled_value, module_context)
  File "/home/lbalhar/Software/jedi/jedi/inference/cache.py", line 44, in wrapper
  File "/home/lbalhar/Software/jedi/jedi/inference/compiled/mixed.py", line 274, in _create
    result = _find_syntax_node_name(inference_state, python_object)
  File "/home/lbalhar/Software/jedi/jedi/inference/compiled/mixed.py", line 189, in _find_syntax_node_name
    path = inspect.getsourcefile(python_object)
  File "/usr/lib64/python3.10/inspect.py", line 817, in getsourcefile
    filename = getfile(object)
  File "/usr/lib64/python3.10/inspect.py", line 785, in getfile
    raise OSError('source code not available')
OSError: source code not available
```

If this usage makes sense, I can create a new test for it.

davidhalter:
Sorry for the delay. I was on vacation.

> If this usage makes sense, I can create a new test for it.

Yes, please! Though I guess this might already be in some kind of a test. You could probably just run `test/test_api/test_interpreter.py` and check if that fails without your change.

frenzymadness:
> Yes, please! Though I guess this might already be in some kind of a test. You could probably just run `test/test_api/test_interpreter.py` and check if that fails without your change.

Hmm, there is something strange. When I test the minimal reproducer manually in Python CLI without this fix, it raises the error:
```
Python 3.10.0rc1 (default, Aug  3 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)] on linux
>>> import jedi
>>> class TestClass():
...     ta=1
...     ta1=2
... 
>>> interpteter = jedi.Interpreter("TestClass.", [globals()])
>>> interpteter.complete(column=10, line=1)
...
OSError: source code not available
```
But if I try to prepare the same code as a test:
```python
def test_completion_dups():
    class TestClass():
        ta=1
        ta1=2

    interpreter = jedi.Interpreter("TestClass.", [locals()])
    completions = interpreter.complete(column=10, line=1)
```
It works fine and does not raise the error. So it might already be covered somewhere in tests but there is a difference in CLI usage which causes the failure.

Do you have an idea where the difference might be and how can I reproduce the error in tests?

davidhalter:
The reason is probably pretty simple: Jedi finds `TestClass`, because it knows the source. If you want to fix that, set it to `TestClass.__name__ = 'NOT_FINDABLE'` or something like this.

frenzymadness:
Thanks for the hint and sorry for the stupid question. It has not been that easy in the end because there is a `__main__` pytest module but I've managed to write a test from the reproducer.

davidhalter:
Oh gosh, this is indeed quite a crazy edge case... There's indeed an `if object.__module__ == '__main__':`. Doesn't make much sense IMO, but usually this is also not an issue. Thanks for the fix!

test/test_api/test_interpreter.py
```diff
@@ -711,3 +711,24 @@ def test_negate():
...
+        ta=1
+        ta1=2
+
+    # Simulate the environment where the class is defined in
+    # an interactive session and therefore inspect module
+    # cannot find its source code and raises OSError (Py 3.10+) or TypeError.
+    TestClass.__module__ = "__main__"
+    # There is a pytest __main__ module we have to remove temporarily.
+    module = sys.modules.pop("__main__")
+
```
davidhalter, line 727: Is there a way you can add a try finally where you reset the module at the end? IMO that would be great to avoid subsequent test failures, because of this side effect.

davidhalter:
> Thanks for the hint and sorry for the stupid question.

Not stupid at all, you couldn't know this :).

test/test_api/test_interpreter.py
```diff
@@ -711,3 +711,26 @@ def test_negate():
...
+    # an interactive session and therefore inspect module
+    # cannot find its source code and raises OSError (Py 3.10+) or TypeError.
+    TestClass.__module__ = "__main__"
+    # There is a pytest __main__ module we have to remove temporarily.
+    module = sys.modules.pop("__main__")
+    try:
+        interpreter = jedi.Interpreter("TestClass.", [locals()])
+        completions = interpreter.complete(column=10, line=1)
+    except:
+        raise
```
davidhalter, line 731: You can just delete those two lines, they do nothing.

frenzymadness: Done

davidhalter:
Thanks!

PR #1800, opened 2021-08-26
Improve completion for pytest fixtures

Boerde:
In the following code jedi would provide completions (see last line) for a generator object. But pytest will return the object "A" in the fixture not a generator. 
So I reordered the jedi code to first check if jedi has to handle a pytest generator here.

```python
import pytest


class A():
    def my_test(self):
        pass


@pytest.fixture
def my_a() -> A:
    yield A()


def test_a(my_a: A):
    my_a.
```

### DIFF
--- a/jedi/inference/value/function.py
+++ b/jedi/inference/value/function.py
@@ -344,7 +344,8 @@ class BaseFunctionExecutionContext(ValueContext, TreeContextMixin):
                     GenericClass(c, TupleGenericManager(generics)) for c in async_classes
                 ).execute_annotation()
         else:
-            if self.is_generator():
+            # If there are annotations, prefer them over anything else.
+            if self.is_generator() and not self.infer_annotations():
                 return ValueSet([iterable.Generator(inference_state, self)])
             else:
                 return self.get_return_values()
--- a/jedi/plugins/pytest.py
+++ b/jedi/plugins/pytest.py
@@ -43,6 +43,9 @@ def infer_anonymous_param(func):
             return function_context.get_return_values()
 
     def wrapper(param_name):
+        # parameters with an annotation do not need special handling
+        if param_name.annotation_node:
+            return func(param_name)
         is_pytest_param, param_name_is_function_name = \
             _is_a_pytest_param_and_inherited(param_name)
         if is_pytest_param:
--- a/test/completion/generators.py
+++ b/test/completion/generators.py
@@ -309,3 +309,8 @@ def annotation2() -> Iterator[float]:
 next(annotation1())
 #? float()
 next(annotation2())
+
+
+# annotations should override generator inference
+#? float()
+annotation1()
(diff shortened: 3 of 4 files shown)


davidhalter:
Good catch.

However, I think this is the wrong approach. The code you modified looks correct IMO. Jedi should prefer annotations to anything else. The question is: Why is the annotation still returned as a generator? This should not be the case. The annotation would need to be `Generator<...>` if a generator was returned. So I think that's where the bug is.

Boerde:
You are right. It is also possible to trigger the bug by a simple snippet. No pytest fixtures needed.
```python
def my_b() -> int:
    yield 4

my_b().
```

Narrowed it down in the code and tried again to fix it + added a testcase.

PeterJCLaw:
I'm not sure that the behaviour proposed here is correct. The code in question _does_ return a generator (whether there's a pytest fixture decoration or not, the function body defined is a generator), so the annotations in the examples here aren't correct.

The expected annotation for a function which is a generator is like this:
``` python
# true definition, which allows for complex generators
def generator() -> Generator[int, None, None]:
    yield 4

# accepted and common shorthand
def iterator() -> Iterator[int]:
    yield 4
```

The annotation applies only to the function body as defined, regardless of what decoration there may be. For example, even though `contextlib.contextmanager` returns a context manager, the functions it decorates should be annotated as iterators:

``` python
@contextlib.contextmaanger
def context_manager() -> Iterator[int]
    yield 4

with context_manager() as num:
    reveal_type(num)  # mypy: revealed type is builtins.int
    assert num == 4
```

I'm not entirely sure what `pytest.fixture`'s signature is, though I would expect it to behave like `contextlib.contextmanager` does (for example) rather than expecting the author to use mismatched annotations.

I would encourage fixing the annotation in the code to return the correct type and see if the annotations on `pytest.fixture` translate it according to the behaviour you describe. It's not clear to me that changing jedi to handle annotations which do not match the code is a useful direction to go.

davidhalter:
Hmm while I kind of agree @PeterJCLaw, shouldn't annotations always be preferred over the code? That's at least how I understand them. A type checker should typically flag what you wrote about, but this is not Jedi's job.

So @Boerde and you are both right in a way: Jedi should just use the annotations and ask no further questions and mypy should flag that piece of code. I also suspect that Jedi just ignores the annotation entirely if a generator is used. I think it's good to always trust annotations, because this makes it possible to have very weird decorators and Python magic that can exist while a type checker can still do the work.

Boerde:
Added two more commits. Through the comment of @PeterJCLaw  I noticed the annotation of the fixture should not even be necessary as the parameter itself already has an annotation which should "abort" special handling of parameters which share the same name as a pytest.fixture
This would already fix my issue with completion in my first code example.
```python
def test_a(my_a: A):
```
Should be enough for jedi to determine the type of "my_a". Jedi should not even check the return type of the fixture with the name "my_a"


Regarding annotations of generators, when searching for a solution of my problem, I stumbled over some comments in jedi code which suggested to prefer annotations when given.

davidhalter:
I guess it should be fine. Thanks!

And sorry for the long delay. I kind of forgot about this PR. Feel free to ping me next time :)

PR #1784, opened 2021-07-02
Fix bug finding implicit namespace packages

bbugyi200:
This PR fixes #1759.

PR #1805, opened 2021-10-02
fix(import): support for nested namespace packages

kirat-singh:
If multiple directories in sys.path provide a nested namespace package, then jedi would only visit the first directory which
contained the package.  Fix this by saving the remaining path list in the ImplicitNamespaceValue and add a test for it.

Added test namespace packages to demonstrate the failure:

`
namespace1/
namespace1/pkg1
namespace1/pkg1/pkg2
namespace1/pkg1/pkg2/mod1.py
namespace2
namespace2/pkg1
namespace2/pkg1/pkg2
namespace2/pkg1/pkg2/mod2.py
`

If both namespace1 and namespace2 are added to sys path, then jedi sees only mod1 without the fix.  After the fix it correctly sees both mod1 and mod2.

### DIFF
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -422,11 +422,9 @@ def import_module(inference_state, import_names, parent_module_value, sys_path):
             # The module might not be a package.
             return NO_VALUES
 
-        for path in paths:
-            # At the moment we are only using one path. So this is
-            # not important to be correct.
+        for i, path in enumerate(paths):
             if not isinstance(path, list):
-                path = [path]
+                path = paths[i:]
             file_io_or_ns, is_pkg = inference_state.compiled_subprocess.get_module_info(
                 string=import_names[-1],
                 path=path,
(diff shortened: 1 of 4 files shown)


kirat-singh:
Whoops this is a duplicate of #1784 I'm happy to withdraw this if the other one is approved.

davidhalter:
I guess the `isinstance` there is kind of weird and does not make sense. I guess just change it to

```
        file_io_or_ns, is_pkg = inference_state.compiled_subprocess.get_module_info(
            string=import_names[-1],
            path=paths,
            full_name=module_name,
            is_global_search=False,
        )
        if is_pkg is None:
            return NO_VALUES
```

The for loop here is removed and not needed anymore. I do not know why it was there in the first place. Probably some refactorings that made it unnecessary.

davidhalter:
#1784 is a duplicate, but it changes some other things as well, so we might as well merge both and resolve the merge conflict :)

The earlier fix will probably just win ;-).

kirat-singh:
thanks for looking, and agree should have simplified.  changed it as you suggested.

davidhalter:
Thanks!

Issue #1818, opened 2021-11-13
No completions for asynccontextmanager


jerluc:
I have not been able to get any asynccontextmanagers (e.g. `httpx.AsyncClient`) to actually work and show completions with jedi, so I decided to make a simple test case to show how normal contextmanagers work but async ones do not:

```python
import jedi


code = """\
import typing


class Thing:
    def get_num(self) -> int:
        return 10


class OtherThing:
    def __init__(self, thing: Thing):
        self.thing = thing

    def __enter__(self) -> Thing:
        return self.thing

    def __exit__(self, exc_type, value, traceback):
        pass


class AsyncOtherThing:
    def __init__(self, thing: Thing):
        self.thing = thing

    async def __aenter__(self) -> Thing:
        return self.thing

    async def __aexit__(self, exc_type, value, traceback):
        pass


def sync_test():
    with OtherThing(Thing()) as th:
        th.


async def async_test():
    async with AsyncOtherThing(Thing()) as th:
        th.
"""


script = jedi.Script(code=code)

completions = script.complete(line=33, column=11)
assert len(completions) > 0
assert completions[0].complete == "get_num"

completions = script.complete(line=38, column=11)
assert len(completions) > 0
assert completions[0].complete == "get_num"
```

In fact, in the failing case, there are no completions at all that are generated.

I believe this might be some kind of bug, but I'm not exactly sure where I can even start to try debugging this particular case. Any pointers on where to start looking would be much appreciated!

jedi version: 0.18.0
Python version: 3.7.9

davidhalter:
I can reproduce. Thanks for the precise issue!

jerluc:
Thanks for confirming @davidhalter. I did a little more digging and it looks like the type inference itself is what's a bit off:

```python
import jedi


code = """\
import typing


class Thing:
    def get_num(self) -> int:
        return 10


class OtherThing:
    def __init__(self, thing: Thing):
        self.thing = thing

    def __enter__(self) -> Thing:
        return self.thing

    def __exit__(self, exc_type, value, traceback):
        pass


class AsyncOtherThing:
    def __init__(self, thing: Thing):
        self.thing = thing

    async def __aenter__(self) -> Thing:
        return self.thing

    async def __aexit__(self, exc_type, value, traceback):
        pass


def sync_test():
    with OtherThing(Thing()) as th:
        th.


async def async_test():
    async with AsyncOtherThing(Thing()) as th:
        th.
"""


[...]

PR #1819, opened 2021-11-16
Adds support for "async with" via #1818

jerluc:
Per the discussion in #1818, this implements basic support for return values from `__aenter__` on async context managers. To test, I've added both a previously-failing inference and completion integration tests that are now passing with this patch. Also a simple test in my IDE confirms that completions are now showing up for the following code example:

```python
class Thing:
    def get_num(self) -> int:
        return 10


class AsyncOtherThing:
    def __init__(self, thing: Thing):
        self.thing = thing

    async def __aenter__(self) -> Thing:
        return self.thing

    async def __aexit__(self, exc_type, value, traceback):
        pass


async def async_test():
    async with AsyncOtherThing(Thing()) as th:
        th.  # Shows `get_num()` and other internal attributes of `Thing`
```

Thanks @davidhalter for the help!

### DIFF
--- a/jedi/inference/syntax_tree.py
+++ b/jedi/inference/syntax_tree.py
@@ -738,6 +738,13 @@ def tree_name_to_values(inference_state, context, tree_name):
         types = infer_expr_stmt(context, node, tree_name)
     elif typ == 'with_stmt':
         value_managers = context.infer_node(node.get_test_node_from_name(tree_name))
+        if node.parent.type == 'async_stmt':
+            # In the case of `async with` statements, we need to
+            # first get the coroutine from the `__aenter__` method,
+            # then "unwrap" via the `__await__` method
+            enter_methods = value_managers.py__getattribute__('__aenter__')
+            coro = enter_methods.execute_with_values()
+            return coro.py__await__().py__stop_iteration_returns()
         enter_methods = value_managers.py__getattribute__('__enter__')
         return enter_methods.execute_with_values()
     elif typ in ('import_from', 'import_name'):
--- a/test/completion/async_.py
+++ b/test/completion/async_.py
@@ -105,3 +105,22 @@ async def f():
     f = await C().async_for_classmethod()
     #? C()
     f
+
+
+class AsyncCtxMgr:
+    def some_method():
+        pass
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *args):
+        pass
+
+
+async def asyncctxmgr():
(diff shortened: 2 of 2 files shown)


davidhalter:
Now a test with `async with open` fails. Should that test even work? You probably know that better than me. Otherwise we can just set it to `[]`, which means no completions.

davidhalter:
But in general I'm definitely merging, I'm just waiting for the pipeline to be green :)

jerluc:
> Now a test with `async with open` fails. Should that test even work? You probably know that better than me. Otherwise we can just set it to `[]`, which means no completions.

Good catch, I noticed this was failing before my patch so I figured it was just broken. But that said, I'm running a different version of Python locally as compared to the failing test, so I'll try to reproduce locally with the same Python version and add a fix.

jerluc:
Testing things out a bit more (and having tested with multiple versions of Python), I'm inclined to say this is actually and invalid test as Python's builtin `open()` returns a file object which is only a sync contextmanager:

```python
async with open("asdf") as f:
    # This causes an AttributeError: __aexit__ on Python versions 3.7-3.8
    # and causes an AttributeError: __aenter__ on Python versions 3.9-3.10
```

As to why this was working before this patch, I believe it's because the previous code was assuming all `with` statements (sync or async) should lookup the `__enter__` method, which a file object _does_ have. This can be demonstrated by running the following code that has an object which is both a sync and async contextmanager using jedi @ 42c5276e04a29b020b39e48e18be4bc709c599df vs. #1819:

```python
import jedi


code = """\
class AsyncAndSync:
    def __enter__(self) -> int:
        return 10

    def __exit__(self, *args, **kwargs):
        pass

    async def __aenter__(self) -> str:
        return "ASYNC"

    async def __aexit__(self, *args, **kwargs):
        pass


def sync_test():
    with AsyncAndSync() as c:
        c


async def async_test():
    async with AsyncAndSync() as c:
        c
"""


script = jedi.Script(code=code)

infs = script.infer(line=17, column=9)
assert len(infs) > 0
assert infs[0].name == "int", infs[0].name

infs = script.infer(line=22, column=9)
assert len(infs) > 0
assert infs[0].name == "str", infs[0].name  # Incorrectly infers `int` on 42c5276
```

With all of that, I'd be okay to remove the existing `async with open()` test if you're happy with this explanation!

davidhalter:
Yes very happy, you can just remove it and I will merge :)

davidhalter:
Awesome, thanks!

Issue #1801, opened 2021-08-31
No completion found for a contextmanager decorating a class method


timrid:
Hi.

I tried to get an auto completion of a contextmanager. It works for an normal function but not for an class method. What could be the reason for this?

Here is an example for a working version (normal function) and a not working version (class method).

```python
code_not_working = """\
import contextlib
from typing import Iterator

class DatabaseConnection:
    def read(self) -> bytes:
        ...

class Database:
    @contextlib.contextmanager
    def connect(self, *args, **kwargs) -> Iterator[DatabaseConnection]:
        ...

db = Database()
with db.connect() as dbcon:
    dbcon.read()  # this is line 15, col 13 is between letter e and a
"""

code_working = """\
import contextlib
from typing import Iterator

class DatabaseConnection:
    def read(self) -> bytes:
        ...

@contextlib.contextmanager
def db_connect(*args, **kwargs) -> Iterator[DatabaseConnection]:
    ...

with db_connect() as dbcon:
    dbcon.read()  # this is line 13, col 13 is between letter e and a
"""

import jedi

script_not_working = jedi.Script(code=code_not_working, path="")
print(f"{script_not_working.complete(line=15, column=13)=}")

script_working = jedi.Script(code=code_working, path="")
print(f"{script_working.complete(line=13, column=13)=}")
```

The output is

```
script_not_working.complete(line=15, column=13)=[]
script_working.complete(line=13, column=13)=[<Completion: read>]
```

My Setup:
```
Win 10
Python 3.8.10 (the standard Python installation)
Jedi 0.18.0
Parso 0.8.2
```

timrid:
I experimented a bit more and think that the error has nothing to do with the contextmanager directly, but with the processing of the decorator.

For this I simplified the example code a bit, so that less dependencies are used:

```python
code = """\
from typing import Callable

class ClassA: ...

class ClassB:
    def do_something(self): ...

def convert_a_to_b(func: Callable[..., ClassA]) -> Callable[..., ClassB]: ...

# code_working ########################################################
@convert_a_to_b
def get_object() -> ClassA: ...

class_b1 = get_object()
class_b1.do_something()  # this is line 15, col 14 is between letter s and o

# code_not_working #####################################################
class SomeClass:
    @convert_a_to_b
    def get_object(self) -> ClassA: ...

some_class = SomeClass()
class_b2 = some_class.get_object()
class_b2.do_something()  # this is line 24, col 14 is between letter s and o
"""
import jedi
[...]

Issue #1824, opened 2021-12-10
test_completion[lambdas:112]: returns list() instead of int()


bnavigator:
Hi,

I get this new test failure on the openSUSE build when trying to update from 0.18.0 to 0.18.1:

```
[   72s] _________________________ test_completion[lambdas:112] _________________________
[   72s] 
[   72s] case = <IntegrationTestCase: /home/abuild/rpmbuild/BUILD/jedi-0.18.1/test/completion/lambdas.py:112 '[a for a in [1,2] if lambda: 3][0]'>
[   72s] monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f4961af44f0>
[   72s] environment = <SameEnvironment: 3.10.1 in /usr>, has_django = False
[   72s] 
[   72s]     def test_completion(case, monkeypatch, environment, has_django):
[   72s]         skip_reason = case.get_skip_reason(environment)
[   72s]         if skip_reason is not None:
[   72s]             pytest.skip(skip_reason)
[   72s]     
[   72s]         if (not has_django) and case.path.endswith('django.py'):
[   72s]             pytest.skip('Needs django to be installed to run this test.')
[   72s]         repo_root = helpers.root_dir
[   72s]         monkeypatch.chdir(os.path.join(repo_root, 'jedi'))
[   72s] >       case.run(assert_case_equal, environment)
[   72s] 
[   72s] /home/abuild/rpmbuild/BUILD/jedi-0.18.1/test/test_integration.py:47: 
[   72s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[   72s] /home/abuild/rpmbuild/BUILD/jedi-0.18.1/test/run.py:217: in run
[   72s]     return testers[self.test_type](compare_cb, environment)
[   72s] /home/abuild/rpmbuild/BUILD/jedi-0.18.1/test/run.py:265: in run_inference
[   72s]     return compare_cb(self, is_str, should)
[   72s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[   72s] 
[...]

PR #1826, opened 2021-12-12
Make typed decorators work for instance methods

PeterJCLaw:
This feels incomplete when compared to `FunctionMixin.py__get__`, however seems to work at least in the cut-down reported.

I've also manually checked that this fixes the more complete version (i.e: with `contextmanager`).

Fixes https://github.com/davidhalter/jedi/issues/1801.


PeterJCLaw:
Hrm, the failure here looks to be #1824

davidhalter:
Thanks! I agree that this is probably incomplete, but it's still good enough. If people are running into issues, they will definitely tell us :).

PR #1827, opened 2021-12-12
Lambdas in comprehensions need parentheses in Python > 3.8

PeterJCLaw:
Fixes https://github.com/davidhalter/jedi/issues/1824.


Apteryks:
This is not part of a release; perhaps a patch level bump could be tagged?

PR #1828, opened 2021-12-22
Update references.py

sajjadhossanshimanto:
Vscode setting folder added


davidhalter:
Why is that necessary?

sajjadhossanshimanto:
I don't know actually 😓

Issue #1850, opened 2022-04-13
Jedi doesn't support PEP 604 (union types as `X | Y`)


GalaxySnail:
Jedi doesn't support [PEP 604](https://peps.python.org/pep-0604/):

```python
import jedi

code = """\
from typing import Union
data: Union[bytes, bytearray]
data.
"""

script = jedi.Script(code=code)
completions = script.complete(line=3, column=5)
print(completions)
assert completions[0].complete == "append"

code = """\
data: bytes | bytearray
data.
"""

script = jedi.Script(code=code)
completions = script.complete(line=2, column=5)
print(completions)
assert completions == []
```

python version: 3.10.2
jedi version: 0.18.1

davidhalter:
This is fixed on master.

GalaxySnail:
As mentioned in #1851, there are still some cases not supported:

``` python
type_in_string: int | "str"

#? int() str()
type_in_string


from typing import Union

pep604_with_typing_union: int | Union[str, list]

#? int() str() list()
pep604_with_typing_union


from typing import TypeVar
T = TypeVar("T")

def generic_func(arg: T) -> int | str | T:
    pass

#? int() str() bytes()
generic_func(b"hello")
```

So it may be better to let this issue open.

davidhalter:
Due to updating typeshed I had to implement a few more union types. The only case of the above mentioned ones that is currently not supported is:

```
from typing import Union

pep604_with_typing_union: int | Union[str, list]

#? int() str() list()
pep604_with_typing_union
```

I will close, since I don't think that is that important and the actual issue seems to be solved.

PR #1851, opened 2022-04-13
Add a naive implementation for PEP 604

GalaxySnail:
releted to #1850, but this PR will not close it, because there are still many cases not supported:

```python
type_in_string: int | "str"

#? int() str()
type_in_string


from typing import Union

pep604_with_typing_union: int | Union[str, list]

#? int() str() list()
pep604_with_typing_union


from typing import TypeVar
T = TypeVar("T")

def generic_func(arg: T) -> int | str | T:
    pass

#? int() str() bytes()
generic_func(b"hello")
```

I don't know how to fix them.

### DIFF
--- a/jedi/inference/syntax_tree.py
+++ b/jedi/inference/syntax_tree.py
@@ -2,6 +2,7 @@
 Functions inferring the syntax tree.
 """
 import copy
+import itertools
 
 from parso.python import tree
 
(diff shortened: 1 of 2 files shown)


davidhalter:
While I do not think that this is a great way to do this, I think this is definitely good enough as a heuristic and could later be improved. Thanks!

davidhalter:
Sorry for taking this long to review. I kind of forgot about it again after seeing it.

PR #1855, opened 2022-05-13
Add option to show params first in signature order

joelostblom:
(This is more of a discussion/issue, but I wanted to show an example of how it could be implemented and what I meant exactly)

What do you think of adding a way to return the parameters first and keep them in the same order they are specified in the signature? When I use auto completion, I find that this is often what I expect when completing inside a function and I think making it available in Jedi would be helpful for packages such as IPython (e.g. https://github.com/ipython/ipython/pull/13673). Apologies if there is already a way to achieve this that I missed.

### DIFF
--- a/jedi/api/completion.py
+++ b/jedi/api/completion.py
@@ -123,7 +123,7 @@ def complete_param_names(context, function_name, decorator_nodes):
 
 class Completion:
     def __init__(self, inference_state, module_context, code_lines, position,
-                 signatures_callback, fuzzy=False):
+                 signatures_callback, fuzzy=False, params_first=False):
         self._inference_state = inference_state
         self._module_context = module_context
         self._module_node = module_context.tree_node
@@ -137,6 +137,7 @@ class Completion:
         self._signatures_callback = signatures_callback
 
         self._fuzzy = fuzzy
+        self._params_first = params_first
 
     def complete(self):
         leaf = self._module_node.get_leaf_for_position(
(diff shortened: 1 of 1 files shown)


joelostblom:
@davidhalter What do you think of adding an option like this to preserve parameter order from the signature? I am happy to work on it if you think it would be valuable and have some suggestions to improve the approach I took here.

davidhalter:
First off, sorry for not replying for such a long time.

I personally think now that Jedi should not offer another way to sort, people should probably sort themselves. There are lots of different opinions how sorting should work. If it is hard to sort Jedi output at the moment, please recommend additional methods on Jedis basic output classes. I feel like that would make people happier in general.

I will close this PR, thanks for the try though. I just really feel like this is not going to fix sorting in Jedi for most people.

joelostblom:
Thanks for the reply @davidhalter. I actually tried implementing sorting downstream of jedi initially (in ipython), but couldn't figure out a good way of doing this since I only had access to the completions themselves as passed from Jedi and not the order they were specified in the function signature since Jedi resorts them. I am not sure how to proceed there, but I will see if I can think of something. I agree with you that there are many different ways to sort, but thought that "in the order of the function signature" would be a pretty common one that would be useful for many people to have access to.

Could you elaborate on what you mean when you say "please recommend additional methods on Jedis basic output classes"? Does that mean you would be open to sorting in Jedi if it is added in the right place? Where in the code would that be?

davidhalter:
> Does that mean you would be open to sorting in Jedi if it is added in the right place? Where in the code would that be?

Not open to sorting in Jedi, but open to offering the complete set of APIs to be able to sort however you want. The code is generally in `jedi/api/classes.py`.

krassowski:
An ability to provide a custom sort `key` function would go a long way, in the sense that downstream could:
- sort (for the most part, excluding the prefixed completions) and
- avoid sorting twice (because we would use the sort key we want immediately).

Would you be open to such an extension of the `Completion` class API?

davidhalter:
No, just sort again using `list.sort`. It's really not a performance question there.

krassowski:
I'm more worried about worst-case than an average scenario. It's 3% for `import numpy; numpy.<tab>` on my hardware (2.65 ms for sorting out of 88.4 ms total). Naive sort by name (without constructing the tuple in the key) is 827 µs. Doing the basic sort twice will push us into 5.8%. I would say non-negligible, but not terrible either.

Screenshot from 2022-11-13 11-56-09

Anything <100ms is perfect but then, I do have a fast machine and the downstream also has some overhead.

davidhalter:
It's just that numpy is probably the worst case, because it contains so many items. Usually you have like 20 items. Sorting those again should be pretty fast. I can understand that you want to optimize there, but I just feel like those 3% extra really do not matter that much.

In general I'd probably be happy for this change if there was somebody else than me maintaining Jedi. At the moment I really do not have that much time for it (because I'm rewriting Jedi in Rust), so I'm generally fighting off non-essential changes like this one.

PR #1870, opened 2022-08-17
fix autocomplete crash in ycmd

Presburger:
Signed-off-by: Yusheng.Ma <Yusheng.Ma@zilliz.com>


davidhalter:
I like the change. Is there a way we can test that?

davidhalter:
I'm too lazy to write the test, but I want this fixed either way, so I'm merging. Would still make me happy if someone wrote a test about that. Thanks anyway!

Presburger:
> I'm too lazy to write the test, but I want this fixed either way, so I'm merging. Would still make me happy if someone wrote a test about that. Thanks anyway!

I'm find this bug in my vim's logs, It is hard to write a trest about this.

PR #1875, opened 2022-09-01
Fix test where home could be a potential project

marciomazza:
The test would fail in an environment where any of the files in
`jedi.api.project._CONTAINS_POTENTIAL_PROJECT` is present at the home directory.


davidhalter:
Thanks!

Issue #1786, opened 2021-07-07
Discover pytest fixtures through imports


idanarye:
On big projects you sometimes want to split organize your fixtures in many files. This means importing the fixtures into either `conftest.py` or - for better modularity - into the test file itself.

#791 added support for pytest fixtures discovery, but it only looks for fixtures defined directly inside the test file or directly inside `conftest.py`. Is it possible to make it look for fixtures imported into these files, just like it would do with regular completion?

davidhalter:
This is definitely possible. It's mainly a performance question/issue. I would myself be happy to have this changed. We might just want to follow direct imports, but not other things.

In general it's definitely possible to use it for goto, but for completions, we might need to reconsider it, because of potential speed issues (especially for large projects).

It's a bit unfortunate, but I have been making a lot of these kind of decisions to avoid performance issues. However, I'm currently rewriting Jedi in Rust (with a lot of builtin type caching). I hope that that will eventually fix the issue for people like you. But feel free to play around with this issue.

davidhalter:
But: Feel free to work on this and get it working. Now that I'm reading the code I'm pretty sure that `_is_pytest_func` is the issue (in `jedi/plugins/pytest.py`).

idanarye:
I looked at the code, and while I don't claim to fully understand it, it does not look like `_is_pytest_func` is at fault. It's purpose is to detect the test function - a task that does not care about the origin of the fixtures in said test function.

I noticed that:

1. `_iter_pytest_modules` does not yield the modules that are star imported in `conftest.py` or in the test files.
2. `FixtureFilter` does not even look at star imports, and does not follow explicit by-name imports.

I'm not sure which of the two should be fixed. I don't _think_ it's a problem in `_iter_pytest_modules`, because it just lists modules without looking at their context, but then again `FixtureFilter` is only looking at the module's syntax tree so I'm not sure it alone can provide the solution.

davidhalter:
My idea was to allow `_is_pytest_func` to check if it is part of an import. If that is the case, it would be possible to infer the underlying function and check again. That imples that `is_pytest_func` would need additional arguments. 

Star imports are a different beast altogether and I'm pretty sure that it's not a good idea to to follow those for performance reasons.

idanarye:
I'm not sure I understand. 

```python
def complete_param_names(func):
    def wrapper(context, func_name, decorator_nodes):
        module_context = context.get_root_context()
[...]

Issue #791, opened 2016-10-14
Add support for pytest fixtures


blueyed:
Pytest fixtures are decorated functions that can be dynamically requested in test functions, simply by using them as function arguments.

I could imagine Jedi supporting this:

``` python
import pytest


class Obj:
    myattr = 1


@pytest.fixture
def fix():
    return Obj()


def test_fix(fix):
    fix.myattr  # gets not completed

f = fix()
f.myattr  # gets completed
```

In `test_fix` it could be known that `fix` is a `pytest.fixture` and then that information could be used for completion.

What works currently with Jedi is using type annotations for this (`def test_fix(fix: Obj)`), but typically that involves importing `Obj` in / before your test function, since `Obj` gets imported typically only in the fixture function itself.

Two things make this more difficult:
1. fixtures are typically defined outside of the test function (and are not imported), in a pytest plugin or `conftest` module (which is basically a local pytest plugin AFAIK).  For this Jedi probably needs support from pytest to get those, based on the current file / location.
2. `pytest.fixture` has an optional `name` kwarg, which allows to use a different name, which adds an additional layer of redirection to this.

I'd be happy to help out with this, of course - also with regard to adding/improving interfaces in pytest.

Related: #257

cpdean:
would it be possible to implement this in the general case or would exceptions have to be made for pytest's case? nothing about python gives you the hint as to where the fixture is defined, it comes into play as pytest collects tests and finds the fixtures by name as they are defined in the given file, related contest.py, and project dependences that define their own pytest fixtures. 

does Jedi have a plugin system where that sort of stuff could be kept? i feel like this would have to be isolated from Jedi in general, as it seems like you'd have to pull quite a bit in scope to find the right fixture

davidhalter:
This would be awesome. I'm using `py.test` a lot and think this would be really useful.

Is there 

> does Jedi have a plugin system where that sort of stuff could be kept? i feel like this would have to be isolated from Jedi in general, as it seems like you'd have to pull quite a bit in scope to find the right fixture

I agree. However a plugin system would definitely be right for this. However I haven't really thought about this and it feels like alot of work. I just don't know really what good entry points would be for such a plugin.

I guess I would be ok with supporting py.test for now (within Jedi) and eventually moving it to a separate module. At the same time I'm not even sure. I think it's OK to support certain core-python libraries  (like pytest or django) within Jedi. It seems a bit awkward to need to install a lot of plugins (that might not be well maintained).

[...]

PR #1879, opened 2022-09-05
Find external pytest fixtures

marciomazza:
Find `pytest` fixtures from external plugins registered via `setuptools` entry points.

Using `setuptools` entry points is probably the main `pytest` mechanism of plugin discovery.

See https://docs.pytest.org/en/stable/how-to/writing_plugins.html#setuptools-entry-points

Some examples of pytest plugins installed like this are `pytest-django`, `pytest-sugar` and `Faker`.

This extends the functionality of #791
and maybe eliminates the need for #1786.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,4 +1,4 @@
-﻿Main Authors
+Main Authors
 ------------
 
 - David Halter (@davidhalter) <davidhalter88@gmail.com>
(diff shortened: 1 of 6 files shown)


davidhalter:
I like it. Sorry for not noticing earlier. Thanks a lot for such a high quality PR.

davidhalter:
However I doubt that it eliminates the need for and #1786, because at least in my codebases where I use imports for fixtures, no entrypoints are present.

PR #1907, opened 2023-01-31
Add support for pytest fixtures from local pytest plugins.

moser:
I like to structure my pytest fixtures in files separate from `conftest.py`. The [pytest docs](https://docs.pytest.org/en/6.2.x/fixture.html#using-fixtures-from-other-projects) advise to include those modules containing fixtures as plugin modules, by giving the fully qualified names as strings in an array assinged to `pytest_plugins` in the `conftest.py`. 

In this PR, I propose adding the modules from the above-mentioned list to the fixture source modules for better completion and type inference.

I guess that my approach in `_load_pytest_plugins` is a bit crude and could use some feedback and error handling. Can you give me feedback on this?


davidhalter:
Hmm good point. I feel like generally you could just infer instead of goto and if a list is inferred infer the inner list elements. That's probably a bit more "solid".

So the order is `infer()`, probably something like `py__iter__`?, `infer()`? each element and then probably check with `get_str_or_none(value)` if a value is a valid string that you could use.

moser:
@davidhalter
I tried to improve on this. Can you have another look?

davidhalter:
@moser Thanks! This looks generally good. It's still marked as a draft. So you might want to finish whatever you want to do and then I'll merge.

moser:
@davidhalter Ready from my side :-)

jedi/plugins/pytest.py
```diff
@@ -181,7 +181,12 @@ def _iter_pytest_modules(module_context, skip_own_module=False):
             if Path(file_io.path) != module_context.py__file__():
                 try:
                     m = load_module_from_path(module_context.inference_state, file_io)
-                    yield m.as_context()
+                    conftest_module = m.as_context()
+                    plugins_list = m.tree_node.get_used_names().get("pytest_plugins")
+                    if plugins_list:
+                        name = conftest_module.create_name(plugins_list[0])
+                        yield from _load_pytest_plugins(module_context, name)
```
davidhalter, line 188: One last question I guess: Now that we yield the plugins before the conftest module, the fixtures of the plugins will be prioritized in gotos/etc. Should it be this way or the other way around?

davidhalter (approved):

jedi/plugins/pytest.py
moser: Just checked it...
If a plugin and `conftest.py` have a fixture with the same name, pytest will prioritize `conftest.py`. Thus, I would turn it around.

davidhalter:
Thanks a lot Martin, good stuff!

PR #1910, opened 2023-02-05
Replaced tuple passed as sys_path actual argument with list

ghrist8p:
Fixes davidhalter#1909

The sys_path argument is used to temporarily replace the actual sys.path which causes sys.path.insert calls to fail.

This issue can also be fixed by protecting assignments to sys.path by wrapping the rvalue in a list but I am not sure about the implications of that.


davidhalter:
Thanks a lot!

Issue #1518, opened 2020-03-04
Cached properties without triggering the cache?


tamasgal:
In a library I work on, we have several "heavy" properties which look something like this (dummy example):

```python
import time

class Foo:
    def __init__(self):
        self._bar = None
        self._baz = None

    @property
    def bar(self):
        if self._bar is None:
            time.sleep(5)
            self._bar = 23
        return self._bar

    @property
    def baz(self):
        if self._bar is None:
            time.sleep(5)
            self._baz = 23
        return self._baz

```

etc. (dozens of properties).

The problem is that creating an instance and hitting `TAB` will evaluate all the code inside the `if`-blocks since `Jedi` sees that the values are `None`, which means the REPL is hanging for several seconds (up to minutes).

```python
>>> from foo import Foo
>>> f = Foo()
>>> f.<TAB>    # hangs for a few seconds
```

Is there a way to e.g. tell Jedi beforehand about the signature of these properties to prevent this huge lag? I also tried using different implementations of `@cached_property` decorators but they have the same problems.

I also tried with type hints without success:

```python
class Foo:
    def __init__(self):
        self._bar = None
        self._baz = None

    @property
    def bar(self) -> int:
        if self._bar is None:
            time.sleep(5)
            self._bar = 23
        return self._bar

    @property
    def baz(self) -> int:
        if self._bar is None:
            time.sleep(5)
            self._baz = 23
        return self._baz
```

davidhalter:
Thanks for the report.

This is unfortunately not really a bug but a "feature". I was against adding this, but most users wanted this kind of side effect.

There's still a hidden way to toggle this, which might be good enough for you. You can say

```
jedi.Interpreter._allow_descriptor_getattr_default = False
```

That should probably do the work. (No idea how you get that properly into an IPython session).

In fact I put in a ton of effort to avoid these kind of cases and a lot of people in the community complained that their properties weren't working anymore. See also this issue: https://github.com/davidhalter/jedi/issues/1299#issuecomment-520253660

PS: While I think you raise a valid point, I think it's an anti pattern to do a lot of calculations in properties. In general I think functions are better, because they allow you to "change" the functionality in the future while not breaking backwards compatibility. This has proven to be a big issue for Jedi in the past. So just a few thoughts (I know that's not always feasible to touch large code bases).

krpatter-intc:
> jedi.Interpreter._allow_descriptor_getattr_default = False

Can we perhaps get this to be an official and non private setting?

[...]

PR #1911, opened 2023-02-08
Make allow_descriptor_getattr a non-private variable for more official

krpatter-intc:
This is a follow up to:

https://github.com/davidhalter/jedi/issues/1518

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -708,7 +708,6 @@ class Interpreter(Script):
     :param namespaces: A list of namespace dictionaries such as the one
         returned by :func:`globals` and :func:`locals`.
     """
-    _allow_descriptor_getattr_default = True
 
     def __init__(self, code, namespaces, *, project=None, **kwds):
         try:
@@ -729,7 +728,7 @@ class Interpreter(Script):
         super().__init__(code, environment=environment, project=project, **kwds)
 
         self.namespaces = namespaces
-        self._inference_state.allow_descriptor_getattr = self._allow_descriptor_getattr_default
+        self._inference_state.allow_descriptor_getattr = settings.instance_allow_descriptor_getattr
 
     @cache.memoize_method
     def _get_module_context(self):
--- a/jedi/settings.py
+++ b/jedi/settings.py
@@ -143,6 +143,12 @@ This improves autocompletion for libraries that use ``setattr`` or
 ``globals()`` modifications a lot.
 """
 
+instance_allow_descriptor_getattr = True
+"""
+Controls whether descriptors are evaluated when using an Interpreter. This is
+something you might want to control when using Jedi from a Repl (e.g. IPython)
+"""
+
 # ----------------
 # Caching Validity
 # ----------------
--- a/test/test_api/test_interpreter.py
+++ b/test/test_api/test_interpreter.py
@@ -8,6 +8,7 @@ import typing
 import pytest
 
 import jedi
+import jedi.settings
 from jedi.inference.compiled import mixed
 from importlib import import_module
 
@@ -219,7 +220,7 @@ def test__getattr__completions(allow_unsafe_getattr, class_is_findable):
 
 @pytest.fixture(params=[False, True])
 def allow_unsafe_getattr(request, monkeypatch):
-    monkeypatch.setattr(jedi.Interpreter, '_allow_descriptor_getattr_default', request.param)
+    monkeypatch.setattr(jedi.settings,'instance_allow_descriptor_getattr', request.param)
     return request.param
 
 


krpatter-intc:
> Why did you not use the `Interpreter.__init__` instead?

The existing attribute? I guess I was going for single source of truth being the setting. Would you prefer to copy the setting in to the Interpreter at time of __init__? or just make it a public attribute on the __init__ ?

krpatter-intc:
Also, not sure how to get to the Interepeter object from something like ipython, but...if you want it just as attribute on the interpreter object (and not a setting) I can try to figure that part out as well and figure out what I need to do as a PR to that repo after...

davidhalter:
I meant something like:

```
def __init__(self, *, allow_descriptor_getattr=True):
    ...
    self._inference_state.allow_descriptor_getattr = allow_descriptor_getattr
```

The problem with this attribute is that you changed the behavior slightly. `allow_descriptor_getattr` is False by default and only in the case of `Interpreter` (i.e. IPython completions), this is True by default. In the default case with `Script` (i.e. IDE tools, static analysis) this attribute is False for security reasons. I personally do not want to change these defaults.

Therefore a parameter to `__init__` would be an obvious choice, but I feel like after your explanation I understand why you wouldn't want that, because it's not easily patch-able as an IPython user. The only alternative IMO would be to use a default like `settings.interpreter_allow_descriptor_getattr = True`. What do you think? This would get rid of the private variable and be much more inviting to patch.

jedi/api/__init__.py
```diff
@@ -708,7 +708,6 @@ class Interpreter(Script):
     :param namespaces: A list of namespace dictionaries such as the one
         returned by :func:`globals` and :func:`locals`.
     """
-    _allow_descriptor_getattr_default = True
```
krpatter-intc, line 711: I thought default was True, so I didn't realize I had functionality changed the default somehow.

>The only alternative IMO would be to use a default like settings.interpreter_allow_descriptor_getattr

So have the Interpreter consume the new setting on creation *vs. right now it is a bit _global_ and changes any existing interpreter object) ?

That makes sense to me. I think that means as long as I make the jedi setting before starting ipython, I'd be ok.

jedi/inference/__init__.py
```diff
@@ -102,7 +102,6 @@ def __init__(self, project, environment=None, script_path=None):
         self.is_analysis = False
         self.project = project
         self.access_cache = {}
-        self.allow_descriptor_getattr = False
```
krpatter-intc, line 105: woops, I see here there's a false....apologies...

jedi/api/__init__.py
```diff
@@ -708,7 +708,7 @@ class Interpreter(Script):
...
+    _allow_descriptor_getattr_default = settings.allow_descriptor_getattr
```
krpatter-intc, line 711: I went overly conservative this time and even keeps support for folks who may have been using the unofficial method previously

eachimei:
Very excited about this enhancement, looking forward to already be able to use it!

jedi/api/__init__.py
davidhalter: I'm not sure I understand your question. You might need to rephrase.

jedi/api/__init__.py
davidhalter: Just remove this attribute. It's impossible for people to set it, because this is calculated at import time (which means you cannot influence it ever. Just remove this line and use `settings.interpreter_allow_descriptor_getattr` where `_allow_descriptor_getattr_default` is used.

jedi/settings.py
```diff
@@ -143,6 +143,11 @@
 ``globals()`` modifications a lot.
 """
 
+allow_descriptor_getattr = True
```
davidhalter, line 146: Rename this to `instance_allow_descriptor_getattr`.

jedi/settings.py
```diff
@@ -143,6 +143,11 @@
...
+"""
+Controls whether descripters are evaulated
```
davidhalter, line 148: 2 typos: descriptors and evaluated.

Maybe use this sentence? `Controls whether descriptors are evaluated when using an Interpreter. This is something you might want to control when using Jedi from a Repl (e.g. IPython).`

davidhalter (commented):
Thanks! I feel like this at least does not change the existing functionality.

I think it's fine if the current mechanism doesn't work for people anymore. It was never "public".

jedi/settings.py
krpatter-intc: Thank you for the assistance on the help here, doc string updated

jedi/settings.py
krpatter-intc: Done

davidhalter:
Thanks!

krpatter-intc:
Cool, thank you!

eachimei:
Hi @davidhalter , can you please share an ETA for the next release? Looking forward to be able to use the new setting.

PR #1916, opened 2023-02-13
Update mypy

PeterJCLaw:
This includes updating the ignore comments for things which mypy now knows about or now complains about, as well as pulling in some typeshed packages for things outside the standard library.

There's also a small amount of added strictness here now it's possible to add it.

This PR should unblock getting CI for Python 3.11 since the latest version of `mypy` supports a version of `typed-ast` which compatible with 3.11.


davidhalter:
Great, thanks a lot!

Issue #1914, opened 2023-02-12
Add '3.11' among _SUPPORTED_PYTHONS


mcepl:
https://github.com/davidhalter/jedi/blob/b0025ee6bac9a7ed1ec4041f38cfdd9f14dc54d3/jedi/api/environment.py#L20

We should ‘3.11’ here.

davidhalter:
Feel free to PR it. Ideally this would also come with all the tests enabled in CI (if it isn't already the case).

mcepl:
I am a maintainer of more than two thousand packages in openSUSE/SLE, I really cannot afford to have checked out repositories for all of them. And this change is so trivial (and [tests with it run](https://build.opensuse.org/package/show/devel:languages:python/python-jedi)) that I leave it to you as an exercise.

PR #1917, opened 2023-02-13
Support Python 3.11

PeterJCLaw:
This adds support for targetting Python 3.11 via picking up the latest grammar from `parso` (added in 0.8.3) while also validating support for running on 3.11 by adding it to the CI matrix.

Fixes https://github.com/davidhalter/jedi/issues/1914

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        python-version: ["3.10", "3.9", "3.8", "3.7", "3.6"]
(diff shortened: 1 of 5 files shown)


davidhalter:
Great, thanks a lot! This really helps a lot, because I do not really like playing with the CI stuff :)

I'm also pretty surprised that this went over so smooth. I feel like there should be no surprises here. 3.11 should be mostly the same from a Jedi perspective (especially because the parso grammar is the same for 3.10/3.11.

jamagalhaes:
hey @davidhalter can you please release this fix? thanks!

tachikoma-li:
Hi @davidhalter, just wondering when the next release will be available to pick up this feature? Thanks!

davidhalter:
Was released (0.19.0)

PR #1918, opened 2023-02-23
guard autocomplete of property

nadavmisgav:
guard evaluating `@property` methods, taken from [bpo-44752](https://github.com/python/cpython/pull/27401/commits/54c6d9c455abc6316a27f7ad443a86546af5f706) to fix IPython accessing `type` [here ](https://github.com/ipython/ipython/blob/75ecfe932fc9ca3505efbebc016b046ffc7d0d68/IPython/core/completer.py#L2903)

### DIFF
--- a/jedi/inference/compiled/access.py
+++ b/jedi/inference/compiled/access.py
@@ -363,7 +363,14 @@ class DirectObjectAccess:
             # warnings should not be shown. See also GH #1383.
             with warnings.catch_warnings(record=True):
                 warnings.simplefilter("always")
-                return_obj = getattr(self._obj, name)
+
+                # Make sure we do not evaluate @property methods on autocompelte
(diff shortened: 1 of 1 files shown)


davidhalter:
I feel like you are probably using Jedi with IPython. Since there is already logic to avoid Jedi accessing properties (checkout `is_allowed_getattr` in the same file you modified), it would be important to know when this happens? Can you post a reproduction for this issue?

davidhalter:
I just realized that you probably simply use the defaults and therefore properties are executed. This is intentional, if you don't want this, you can change the defaults that were changed here: https://github.com/davidhalter/jedi/pull/1911/files

However, note that the pull request that I linked is on master, so you have to set the defaults a bit different on 0.18.1.

nadavmisgav:
Hi thanks for the reply, `jedi.settings.instance_allow_descriptor_getattr = False` fixed my problem.

nadavmisgav:
would you mind checking https://github.com/ipython/ipython/pull/13951

Issue #1764, opened 2021-03-30
renaming a variable inside a comprehension goes wrong


dimbleby:
eg start with 
```python
foo = [x for x in [1,2,3] if x < 3]
```
and try to rename `x` as `y`.  Get
```python
foo = [y for y in [1,2,3] if x < 3]
```
where you can see that the final `x` survives.

Here's the same example expressed as a testcase, which currently fails (eg if I put it in `test_refactoring.py`):
```
  def test_rename_in_comprehension(Script, dir_with_content):
      script = Script(
          'foo = [x for x in [1,2,3] if x < 3]\n',
          path=dir_with_content.joinpath('some_script.py'),
          project=jedi.Project(dir_with_content),
      )
      refactoring = script.rename(line=1, column=7, new_name='y')
      assert refactoring.get_diff() == dedent('''\
          --- some_script.py
          +++ some_script.py
          @@ -1,2 +1,2 @@
          -foo = [x for x in [1,2,3] if x < 3]
          +foo = [y for y in [1,2,3] if y < 3]
          ''')
```

davidhalter:
Thanks for the report. Since you asked how you could potentially fix that:

The issue is that goto doesn't work on the `x` in the if clause. Once that works, renaming should probably work. This is probably an issue with the context. The context gets created in `jedi.inference.context.TreeContextMixin.create_context` and there should probably be a special case for the `comp_if` node and we should also somehow end up with a CompForContext. The `sync_comp_for` is probably currently it's context, which means that the if part has no access to the variables defined in that context, I guess.

PR #1923, opened 2023-03-12
fix help when in column zero

dimbleby:
see https://github.com/davidhalter/parso/pull/217, https://github.com/davidhalter/parso/pull/218 for discussion - this is a much more limited and targeted fix

### DIFF
--- a/jedi/api/__init__.py
+++ b/jedi/api/__init__.py
@@ -369,6 +369,12 @@ class Script:
         if definitions:
             return definitions
         leaf = self._module_node.get_leaf_for_position((line, column))
+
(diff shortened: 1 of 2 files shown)


davidhalter:
@dimbleby Thanks a lot for the fix. This is exactly right.

If you're interested I'm happy to help you with other issues within Jedi as well. But thanks anyway!

dimbleby:
thanks - jedi (for me via the language server) mostly works great, I've little to complain about!

I did raise #1764 a while back and it still occasionally catches me out.  If you're able to leave a hint there as to what would be needed to deal with it; and if that hint suggests that it's not too hard - then maybe one day I'd feel like taking a look at it...

Issue #1929, opened 2023-04-17
Test failure with attrs 23.1.0


frenzymadness:
It seems that one test fails with the latest attrs 23.1.0, the output is:

```
=================================== FAILURES ===================================
____________ test_attrs_signature[@frozen\nclass X:-start_params1] _____________

Script = functools.partial(<class 'jedi.api.Script'>, environment=<Environment: 3.11.3 in /usr>)
skip_pre_python37 = None, start = '@frozen\nclass X:', start_params = []

    @pytest.mark.parametrize(
        'start, start_params', [
            ['@define\nclass X:', []],
            ['@frozen\nclass X:', []],
            ['@define(eq=True)\nclass X:', []],
            [dedent('''
             class Y():
                 y: int
             @define
             class X(Y):'''), []],
            [dedent('''
             @define
             class Y():
                 y: int
                 z = 5
             @define
             class X(Y):'''), ['y']],
        ]
    )
    def test_attrs_signature(Script, skip_pre_python37, start, start_params):
        has_attrs = bool(Script('import attrs').infer())
        if not has_attrs:
            raise pytest.skip("attrs needed in target environment to run this test")
    
        code = dedent('''
                name: str
                foo = 3
                price: float
                quantity: int = 0.0
    
            X(''')
    
        # attrs exposes two namespaces
        code = 'from attrs import define, frozen\n' + start + code
    
        sig, = Script(code).get_signatures()
>       assert [p.name for p in sig.params] == start_params + ['name', 'price', 'quantity']
E       AssertionError: assert [] == ['name', 'price', 'quantity']
E         Right contains 3 more items, first extra item: 'name'
E         Use -v to get more diff

test/test_inference/test_signature.py:394: AssertionError
=============================== warnings summary ===============================
jedi/api/__init__.py::jedi.api.Interpreter
test/test_utils.py::TestSetupReadline::test_modules
test/test_api/test_interpreter.py::test_complete_raw_function
test/test_api/test_interpreter.py::test_complete_raw_function_different_name
test/test_api/test_interpreter.py::test_complete_raw_module
  <frozen importlib._bootstrap_external>:1528: DeprecationWarning: PathFinder.find_module() is deprecated and slated for removal in Python 3.12; use find_spec() instead

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
[...]

PR #1935, opened 2023-05-26
Teach Jedi that `attrs`' `frozen` decorator also acts like a dataclass

PeterJCLaw:
In 23.1.0 (specifically in https://github.com/python-attrs/attrs/commit/46053d703d99a6e7419fd291cbf546340690d69d) the definition of the `frozen` decorator was tweaked slightly, such that its type stub is separate from that for `define`. This means that Jedi needs to be told about it as a separate member.

I've manually checked that this still works with the prior version of `attrs`.

Fixes https://github.com/davidhalter/jedi/issues/1929


davidhalter:
Great thanks!

PR #1936, opened 2023-05-27
Support `importlib.metadata` entry points for newer python

PeterJCLaw:
`pkg_resources` is deprecated and liable to be dropped at some point.


davidhalter:
Generally I would argue we can just merge this if tests are passing. The tests should test this pretty well.

PR #1937, opened 2023-05-27
Modernise importlib usage

PeterJCLaw:
`find_module` is deprecated in all supported version of Python and is slated for removal in the upcoming 3.12. Happily it seems we can move to the related `find_spec` and just hoist the loader from the spec which that returns. (This is mostly what current `find_module` implementations do anyway).

This is relying on code inspection, some guessing around Python import machinery and existing tests. I'm not sure best to validate this manually.

### DIFF
--- a/jedi/inference/compiled/subprocess/functions.py
+++ b/jedi/inference/compiled/subprocess/functions.py
@@ -2,7 +2,6 @@ import sys
 import os
 import inspect
 import importlib
-import warnings
 from pathlib import Path
 from zipfile import ZipFile
 from zipimport import zipimporter, ZipImportError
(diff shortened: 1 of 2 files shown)


jedi/inference/compiled/subprocess/functions.py
```diff
@@ -167,17 +166,16 @@ def _find_module(string, path=None, full_name=None, is_global_search=True):
...
+            loader = spec.loader
 
     if loader is None and path is None:  # Fallback to find builtins
         try:
-            with warnings.catch_warnings(record=True):
-                # Mute "DeprecationWarning: Use importlib.util.find_spec()
-                # instead." While we should replace that in the future, it's
-                # probably good to wait until we deprecate Python 3.3, since
-                # it was added in Python 3.4 and find_loader hasn't been
-                # removed in 3.6.
```
davidhalter, line 179: That's quite a long time ago, wow :)

davidhalter:
Thanks for cleaning up stuff. I like :)

Issue #1955, opened 2023-09-16
Test failures with Python 3.12


hugovk:
The [second Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105/5?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Python 3.12.0 final is due out in two weeks: [2023-10-02](https://peps.python.org/pep-0693/).

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

---

However, adding 3.12 to the CI like in https://github.com/hugovk/jedi/commit/7e5d82113a10a1eaeb04228a9919322cedc29b24 results in [three failed tests](https://github.com/hugovk/jedi/actions/runs/6207769604):

```pytb
=================================== FAILURES ===================================
________________________ TestSetupReadline.test_import _________________________

self = <test.test_utils.TestSetupReadline testMethod=test_import>

    def test_import(self):
        s = 'from os.path import a'
        assert set(self.complete(s)) == {s + 'ltsep', s + 'bspath'}
        assert self.complete('import keyword') == ['import keyword']
    
        import os
        s = 'from os import '
        goal = {s + el for el in dir(os)}
        # There are minor differences, e.g. the dir doesn't include deleted
        # items as well as items that are not only available on linux.
        difference = set(self.complete(s)).symmetric_difference(goal)
        difference = {
            x for x in difference
            if all(not x.startswith('from os import ' + s)
                   for s in ['_', 'O_', 'EX_', 'MFD_', 'SF_', 'ST_',
                             'CLD_', 'POSIX_SPAWN_', 'P_', 'RWF_',
                             'SCHED_'])
        }
        # There are quite a few differences, because both Windows and Linux
        # (posix and nt) librariesare included.
>       assert len(difference) < 30
E       AssertionError: assert 39 < 30
E        +  where 39 = len({'from os import CLONE_FILES', 'from os import CLONE_FS', 'from os import CLONE_NEWCGROUP', 'from os import CLONE_NEWIPC', 'from os import CLONE_NEWNET', 'from os import CLONE_NEWNS', ...})

test/test_utils.py:88: AssertionError
______________________ test_create_environment_executable ______________________

    def test_create_environment_executable():
[...]

PR #1956, opened 2023-09-16
Support Python 3.12

PeterJCLaw:
Fixes https://github.com/davidhalter/jedi/issues/1955

### DIFF
--- a/jedi/api/environment.py
+++ b/jedi/api/environment.py
@@ -17,7 +17,7 @@ import parso
 
 _VersionInfo = namedtuple('VersionInfo', 'major minor micro')  # type: ignore[name-match]
 
-_SUPPORTED_PYTHONS = ['3.11', '3.10', '3.9', '3.8', '3.7', '3.6']
+_SUPPORTED_PYTHONS = ['3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6']
 _SAFE_PATHS = ['/usr/bin', '/usr/local/bin']
 _CONDA_VAR = 'CONDA_PREFIX'
 _CURRENT_VERSION = '%s.%s' % (sys.version_info.major, sys.version_info.minor)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        python-version: ["3.11", "3.10", "3.9", "3.8", "3.7", "3.6"]
-        environment: ['3.8', '3.11', '3.10', '3.9', '3.7', '3.6', 'interpreter']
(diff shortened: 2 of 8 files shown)


davidhalter:
I think the `test_compiled_signature` test is easily fixable, you can just change the line from

```
('next', 'next(iterator, default=None, /)', ['iterator', 'default'], ge, (3, 6)),
```
to

```
('next', 'next(iterator, default=None, /)', ['iterator', 'default'], lt, (3, 12)),
('next', 'next(iterator, default=None, /)', [], ge, (3, 12)),
```

I'm not 100% sure this is the correct solution, because I haven't tried it. But I feel like you get what I want to do here.

-----

The docstring test about next is a test that I would probably just skip if the environment doesn't match the currrent version (or InterpreterEnvironment). It's probably as simple as adding an `environment` fixture and then checking `environment.version_info == sys.version_info`. Do you understand what I think here?

jedi/api/environment.py
```diff
@@ -17,7 +17,7 @@
 
 _VersionInfo = namedtuple('VersionInfo', 'major minor micro')  # type: ignore[name-match]
 
-_SUPPORTED_PYTHONS = ['3.11', '3.10', '3.9', '3.8', '3.7', '3.6']
+_SUPPORTED_PYTHONS = ['3.12', '3.11', '3.10', '3.9', '3.8', '3.7', '3.6']
```
davidhalter, line 20: Just FYI I'm still thinking about allow users to use newer versions than 3.12, even if they are not officially supported yet. I feel like it's a bit strange, that CPython devs are not able to use a 3.13 Jedi at the moment, even though that is mostly the same as 3.12. But that's probably for another pull request.

PeterJCLaw:
Thanks, have adapted for the `next` signature docs changing in a60fdba1d460f98e15f74a3c83079c74451f4116.

jedi/api/environment.py
hugovk: Agreed.

For Pillow, we let people install for whatever version and platform they want.

But because of the C extensions, you're best of using a prebuilt wheel for Windows, as it's likely to fail for most people who don't have a compiler set up.

So for Windows + 3.13+ we print a warning instead: https://github.com/python-pillow/Pillow/blob/c556c5f2412658a80fceb8f8533d49ba70b319e5/setup.py#L42-L53

And there will be an alpha release next month for 3.13 ([PEP 719](https://peps.python.org/pep-0719/)), it's good to let people test early, it can help find bugs in CPython nice and early.

davidhalter:
Thanks a lot for taking care of this! I really appreciate this effort especially, because it involves a lot of annoying things.

davidhalter:
What do you think: Should I release a new version?

PeterJCLaw:
Yeah, a new version with this support would be good IMO. Can probably be 0.19.1 as there's nothing breaking?

davidhalter:
Released.

PR #1961, opened 2023-09-22
Include a link to the docs in the package metadata

PeterJCLaw:


davidhalter:
Thanks. Didn't even know about this.

PeterJCLaw:
Yeah, it's handy. I'm not completely sure how it works, but from what I can tell this key can contain arbitrary name: url pairs. Another one I thought about adding was a link to the changelog (though I realise that's embedded in the README too).

PR #1964, opened 2023-10-07
use `sys.executable()` standard function to obtain python executable

kloczek:
Use `sys.executable()` standard function instead custom code which assumes that python executable is `python`.


davidhalter:
The tests are failing unfortunately. If this works well, I'm happy to merge it, but not as long as the tests fail.

kloczek:
> The tests are failing unfortunately. If this works well, I'm happy to merge it, but not as long as the tests fail.

Looks like it fails in one env in custom code which is trying to figure out what is the sitelib path (which should be obtained from `sysconfig.get_path('purelib')`) 🤔

davidhalter:
Closing, because I don't think there is progress here (and I'm pretty sure that this might break in some cases).

Issue #1965, opened 2023-10-10
allow users to set cache_directory via env-var


woutervh:
jedi is mostly installed in a virtualenv, but uses $XDG's  ~/.cache/jedi ,  which is a good default / fallback.

Many python tools/libs allow the user to set the cache_dir via env-var:
BLACK_CACHE_DIR
MYPY_CACHE_DIR
RUFF_CACHE_DIR
....

please support an env-var JEDI_CACHE_DIR

davidhalter:
Why?

woutervh:
because libraries installed in a virtualenv should not write to hardcoded externals dirs like  ~/.cache/jedi.
XDG-directories are a very good default, but they should still remain default.

and setting XDG_CACHE_HOME to some project-specific dir only creates other issues.

woutervh:
and because I want to avoid monkeypatching :)

davidhalter:
But this does not explain when you would use this?

woutervh:
I never use jedi directly but subdependencies in many of my projects are.

The whole idea of virtualenvs is to provide a certain isolation from the rest of the system and the ability to run many versions in parallel. Writing to $HOME runs counter to that.

woutervh:
> this does not explain when you would use this?

I'm not understanding your question.
I never want a virtualenv-installed python library to write to $HOME.

So when would I use this JEDI_CACHE_DIR? In every project that depends on jedi.

davidhalter:
But how do you set this? Do you really set this variable by hand in every venv?

woutervh:
I set env-vars in a .env and pip install autoread-dotenv to automatically process the .env.

see  https://autoread-dotenv.readthedocs.io/   ( I wrote this package)

davidhalter:
Sorry this feels not important enough for me. I'm closing.

PR #1966, opened 2023-10-13
allow users to set cache_directory via env-var,  #1965

woutervh:
implements https://github.com/davidhalter/jedi/issues/1965

### DIFF
--- a/jedi/settings.py
+++ b/jedi/settings.py
@@ -86,7 +86,9 @@ cache_directory = os.path.expanduser(_cache_directory)
 The path where the cache is stored.
 
 The cache-directory can be explicitly set by an environment-variable
-``JEDI_CACHE_DIR``. If this env-var in not set, then
+``JEDI_CACHE_DIR``. 
+
+If this env-var is not set, then
 on Linux, the cache-directory defaults to ``~/.cache/jedi/``,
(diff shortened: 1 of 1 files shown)


jedi/settings.py
```diff
@@ -83,10 +85,17 @@
 """
 The path where the cache is stored.
 
-On Linux, this defaults to ``~/.cache/jedi/``, on OS X to
-``~/Library/Caches/Jedi/`` and on Windows to ``%LOCALAPPDATA%\\Jedi\\Jedi\\``.
+The cache-directory can be explicitly set by an environment-variable
+``JEDI_CACHE_DIR``.
+
+If this env-var is not set, then
```
davidhalter, line 91: Can you revert this sentence? I don't like the grammar.

woutervh: reverted.

davidhalter:
Now it's not documented anymore. Sorry I don't have the patience for this. I initially rejected this, so I'm going to close.

woutervh:
what does the word "revert" mean in your dictionary?

woutervh:
Python applications and certainly libraries should never write anything to hardcoded unconfigurable directories.

I don't understand your objection to such no brainer.

PR #1976, opened 2023-12-22
Implement added_conftest_path

segoon:
Add an argument for extra conftest files paths. It is used in specific environments where fixtures are collected through nonstandard way, e.g. in Yandex monorepository Arcadia. One may define a list of extra conftest.py files and use fixtures from there.

Also one can define pytest plugins in 'pytest_plugins' local variable and expect fixtures from the imported plugins to be accessible via jedi.


davidhalter:
Sorry, but this is something that we won't support (due to the ongoing maintenance). In general I'm avoiding such special cases. Thanks for the attempt, though.

PR #1977, opened 2023-12-28
[Typo] Fix some misalignment of docstrings

HairlessVillager:
Thanks for your work👍 I think this is a typo, so I create the pr directly.

> A string literal which appears as **the first expression** in a class, function or module. 
>
> *from [Glossary - Python Documentation](https://docs.python.org/3/glossary.html#term-docstring)*


davidhalter:
100%. Thanks!

PR #1983, opened 2024-02-07
properties with setters are now reported as 'property' for completion

eirannejad:
When complete is called on this script, current jedi reports `X` as a `'method'` instead of a `'property'`. This PR fixes that

```python
class Point3d:
    @property
    def X(self) -> float:
        return 0.0

    @X.setter
    def X(self, value: float):
        return None

p = Point3d()
p.
```

### DIFF
--- a/jedi/parser_utils.py
+++ b/jedi/parser_utils.py
@@ -339,10 +339,10 @@ def _function_is_x_method(decorator_checker):
 function_is_staticmethod = _function_is_x_method(lambda m: m == "staticmethod")
 function_is_classmethod = _function_is_x_method(lambda m: m == "classmethod")
 function_is_property = _function_is_x_method(
-    lambda m: m == "property" \
-                or m == "cached_property" \
-                # do not report __class__.__setter__ as a property
-                # eirannejad 2024-02-07:
-                # not sure if there are other cases but this passes the tests
-                or (m.endswith(".setter") and not m.startswith('__class__'))
+    lambda m: m == "property"
+            or m == "cached_property"
+            # do not report __class__.__setter__ as a property
+            # eirannejad 2024-02-07:
+            # not sure if there are other cases but this passes the tests
+            or (m.endswith(".setter") and not m.startswith('__class__'))
 )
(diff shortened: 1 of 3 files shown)


eirannejad:
@davidhalter I'm not sure why the integration tests are failing. Any ideas?

davidhalter:
Have a look at the failed test in `test/completion/context.py` at line 28 and I'm sure you will understand what failed (just read the comments) :)

davidhalter:
@eirannejad No sorry, that's not what I meant. My fault, sorry. I thought it was obvious that the comment `IMO this is not wanted` means that the tests are wrong and not your implementation. You actually fixed something that was previously just tested in its current (unfinished) state.

eirannejad:
@davidhalter Okay. I think I got that but to be honest I don't know the depth of what jedi does so I can write new tests for this change. I modified the PR so it does not report `property` for the `__class__` as it apparently has a `.setter` decorator. 

I could use a little guidance on how you'd want this to work ideally so I can make the necessary changes.

davidhalter:
OK, sorry. It seems I was still not clear enough. The original change you did was the right one.

However the current tests that were failing before are wrong:

```
    # This might or might not be what we wanted, currently properties are also
    # used like this. IMO this is not wanted ~dave.                           
    #? ['__class__']                                                          
    def __class__                                                             
    #? []                                                                     
    __class__                                                                 
```

This is what we currently have, but it's not what we want as I mentioned in the comment of the tests. Therefore I wanted you to change the `test/completion/context.py` file, remove my old comment (that says `[..] IMO this is not wanted [..]` and change the tests like this:

```
    #? []                                                          
    def __class__                                                             
    #? ['__class__']                                                                     
    __class__ 
```

This is what we would have wanted from the start with these tests, but without your change that was not possible. I guess I thought you would understand my comment and simply change the tests, but I can understand that this wasn't clear. Do you understand now?

eirannejad:
@davidhalter I think this is ready.

davidhalter:
Thanks a lot! @eirannejad

PR #1989, opened 2024-04-17
Fix #1988

### LINKED ISSUE #1988: Completion fails when variable name is identical to type name
The following is valid python code:

```python
class Test:
    str: str = 'abc'

print(Test.str)
```
Code completion fails though at the last dot below.

`Test.str.`

If you change the variable name "str" to anything else, code completion works.   The problem is caused when the variable name is identical to the type name.  This is an issue with certain stub files (e.g. [delphivcl](https://pypi.org/project/delphivcl/)) in which variable type annotations such as
`Bitmap: Bitmap`
are quite common.

pyscripter:
See #1988 

Test:

```python
from jedi import Script

source = """class Test:
    str: str

print(Test.str.)
"""

source2 = """class Picture:
    bitmap: bytes

class Image:
    Picture: Picture

img = Image()
pic = img.Picture.
"""

def main():
    script = Script(source)
    completions = script.complete(4, 15)
    assert completions[0].name == 'capitalize'

    script = Script(source2)
    completions = script.complete(8, 18)
    assert completions[0].name == 'bitmap'

if __name__ == '__main__':
    main()
```

### DIFF
--- a/jedi/inference/syntax_tree.py
+++ b/jedi/inference/syntax_tree.py
@@ -698,9 +698,11 @@ def tree_name_to_values(inference_state, context, tree_name):
                 ann_assign = expr_stmt.children[1]
                 if correct_scope and name.start_pos < ann_assign.start_pos:
                     found_annotation = True
-                    if ((expr_stmt.children[1].children[1].type == 'name') and
-                            (expr_stmt.children[1].children[1].value == tree_name.value) and
-                            context.parent_context):
+                    if (
+                        (expr_stmt.children[1].children[1].type == 'name')
+                        and (expr_stmt.children[1].children[1].value == tree_name.value)
+                        and context.parent_context
+                    ):
                         context = context.parent_context
                     value_set |= annotation.infer_annotation(
                         context, expr_stmt.children[1].children[1]
--- a/jedi/inference/syntax_tree.py
+++ b/jedi/inference/syntax_tree.py
@@ -695,8 +695,13 @@ def tree_name_to_values(inference_state, context, tree_name):
 
             if expr_stmt.type == "expr_stmt" and expr_stmt.children[1].type == "annassign":
                 correct_scope = parser_utils.get_parent_scope(name) == context.tree_node
-                if correct_scope:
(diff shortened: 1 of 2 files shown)


jedi/inference/syntax_tree.py
```diff
@@ -695,8 +695,15 @@ def tree_name_to_values(inference_state, context, tree_name):
 
             if expr_stmt.type == "expr_stmt" and expr_stmt.children[1].type == "annassign":
                 correct_scope = parser_utils.get_parent_scope(name) == context.tree_node
-                if correct_scope:
+                ann_assign = expr_stmt.children[1]
+                if correct_scope and name.start_pos < ann_assign.start_pos:
                     found_annotation = True
+                    if (
+                        (expr_stmt.children[1].children[1].type == 'name')
```
davidhalter, line 702: Can you replace `expr_stmt.children[1]` in all of these cases with `ann_assign`?

jedi/inference/syntax_tree.py
```diff
@@ -695,8 +695,15 @@ def tree_name_to_values(inference_state, context, tree_name):
 
             if expr_stmt.type == "expr_stmt" and expr_stmt.children[1].type == "annassign":
                 correct_scope = parser_utils.get_parent_scope(name) == context.tree_node
-                if correct_scope:
+                ann_assign = expr_stmt.children[1]
+                if correct_scope and name.start_pos < ann_assign.start_pos:
```
davidhalter, line 699: I just realized that you were probably correct with your solution and the start_pos comparison is not needed, so please remove that again. Sorry.

davidhalter (commented):
Can you please add a test to `test/completion/pep0484_basic.py`? You can problaby just take the code piece you provided and add a "test":

```class Test:
    str: str = 'abc'

#? ['upper']
Test.str.upp
```

Thanks!

jedi/inference/syntax_tree.py
pyscripter: Done

pyscripter:
Added test and addressed your feedback.

PR #2006, opened 2024-06-30
Enable linting of our test code, but not our test data files

PeterJCLaw:
This aims to make development of tests easier by adding more automated checks for minor errors. It also makes some slightly unusual code more explicit about what it's doing by requiring markers.

### DIFF
--- a/test/test_api/test_interpreter.py
+++ b/test/test_api/test_interpreter.py
@@ -26,7 +26,7 @@ def get_completion(source, namespace):
 
 
 def test_builtin_details():
-    import keyword
+    import keyword  # noqa: F401
 
     class EmptyClass:
         pass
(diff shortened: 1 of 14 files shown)


test/test_api/test_interpreter.py
```diff
@@ -532,7 +532,6 @@ def func(a, b, c):
 
 def test_type_var():
     """This was an issue before, see Github #1369"""
-    import typing
```
PeterJCLaw, line 535: I'm _assuming_ that the inline nature of this import is immaterial; I can't see anything in the origin of this test (https://github.com/davidhalter/jedi/commit/ab80646b860d89f0fea2c3aea262cd922c7413d7) which suggests that a global import (as this file now has) would change the behaviour, but would be good to get a second opinion.

davidhalter:
Thanks!

Issue #2010, opened 2024-07-02
Rhs in the form of "x" in "xxx" cannot infer the type


WutingjiaX:
I found in **_infer_comparison_part** ,There are the following code snippets：
```
    elif str_operator in ('in', 'not in'):
        return NO_VALUES
```
This will result in the inability to infer the type when cursor under the word **res**:
```
res = "f" in "foo"
```

Why not return the following code like the previous branch？：
```
  return ValueSet([
      _bool_to_value(inference_state, True),
      _bool_to_value(inference_state, False)
  ])
```

davidhalter:
Can you a bit more specific? What changes are you proposing?

WutingjiaX:
As I mentioned earlier, when I want to infer the type of the following code (the type of **res**), jedi cannot provide a result：
```
res = "f" in "foo"
```

the reason of that is in the function `def _infer_comparison_part(inference_state, context, left, operator, right)`, When encountering a branch of the **in** operator，it will return `NO_VALUES`

```
elif str_operator in ('in', 'not in'):
        return NO_VALUES
```

So why doesn't this branch return the following content：
```
  return ValueSet([
      _bool_to_value(inference_state, True),
      _bool_to_value(inference_state, False)
  ])
```
Like the returned content in the previous `COMPARISON_OPERATORS`branch:

```
    elif str_operator in COMPARISON_OPERATORS:
        if left.is_compiled() and right.is_compiled():
            # Possible, because the return is not an option. Just compare.
            result = left.execute_operation(right, str_operator)
            if result:
                return result
        else:
            # Omit some code

        return ValueSet([
            _bool_to_value(inference_state, True),
            _bool_to_value(inference_state, False)
        ])
```

After making the modifications, it can infer the correct content and all existing unit tests can still pass

PeterJCLaw:
@WutingjiaX thanks for reporting this, you're correct that Jedi currently doesn't infer a type here. I'd guess that the vast majority of uses of `x in y` expressions are within conditional statements rather than assignments, which is perhaps why this hasn't been an issue before.

Note that [`__contains__`](https://docs.python.org/3/reference/datamodel.html#object.__contains__) is something which types can optionally provide, so not all `x in y` tests are valid. The standard data model does indicate it should return `True` or `False` though, so I think assume that that's the return type (rather than e.g: looking up the real signature) could be a reasonable assumption. Given that many of the basic types do provide it, it may also be reasonable to assume that such a test is valid.

[...]

PR #2013, opened 2024-07-09
When inferring comparison operators, return a definite type instead of NO_VALUES for the in/not in operator

WutingjiaX:
As discussed in #2010

### DIFF
--- a/test/test_api/test_api.py
+++ b/test/test_api/test_api.py
@@ -403,3 +403,8 @@ def test_infer_after_parentheses(Script, code, column, expected):
         assert completions == []
     else:
         assert [c.name for c in completions] == [expected]
+
+
+def test_infer_in_operator(Script):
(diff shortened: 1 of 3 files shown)


test/test_api/test_api.py
```diff
@@ -403,3 +403,8 @@ def test_infer_after_parentheses(Script, code, column, expected):
         assert completions == []
     else:
         assert [c.name for c in completions] == [expected]
+
+
+def test_infer_in_operator(Script):
+    completions = Script("res = 'f' in 'foo'").infer(column=1)
+    assert completions[0].full_name == 'builtins.bool'
```
davidhalter, line 410: Can you please remove that test and add it to `test/completion/basic.py`? You can probably guess how that works in there.

Thanks a lot!

PR #2014, opened 2024-07-10
 When inferring comparison operators, return a definite type instead of NO_VALUES for the in/not in operator

WutingjiaX:
As discussed in #2010


davidhalter:
Thanks!

PR #2027, opened 2024-10-16
Filter duplicate imports when completing

WutingjiaX:
For the code snippet like:
```
from pydantic import Field, F
```
When the cursor is after the letter F，the completion includes item "Field", But this import actually already existed before

Pycharm also will not complete:
image

So I create a new pull request to filter duplicate imports.

However, there are still some details that have not been implemented(I don't think it's very important)，such as:
* filter duplicate imports in multiple imports statement
* import alias will also be filtered Generally, there is unlikely to be a situation where an alias is named exactly the same as a specific original name.

### DIFF
--- a/test/test_inference/test_imports.py
+++ b/test/test_inference/test_imports.py
@@ -307,6 +307,28 @@ def test_os_issues(Script):
     assert 'path' in import_names(s, column=len(s) - 3)
 
 
+def test_duplicated_import(Script):
+    def import_names(*args, **kwargs):
+        return [d.name for d in Script(*args).complete(**kwargs)]
+
+    s = 'import os, o'
+    assert 'os' not in import_names(s)
+    assert 'os' in import_names(s, column=len(s) - 3)
+
+    s = 'from os import path, p'
+    assert 'path' not in import_names(s)
+    assert 'path' in import_names(s, column=len(s) - 3)
+
+    s = 'import path as pp, p'
+    assert 'path' not in import_names(s)
+
+    s = 'from os import path as pp, p'
+    assert 'path' not in import_names(s)
+
+    s = 'from os import chdir as path, p'
+    assert 'path' in import_names(s)
+
+
 def test_path_issues(Script):
     """
     See pull request #684 for details.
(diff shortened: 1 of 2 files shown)


WutingjiaX:
The CI is faile in specific operating system version， But it seems that it was not introduced by my changes

davidhalter:
After a small fix the tests are fine again (https://github.com/davidhalter/jedi/actions/runs/11364219966). Would it be possible for you to rebase? I'd prefer to merge once the CI is green.

Thanks again, I think this is a useful addition and something that I found annoying as well!

test/test_inference/test_imports.py
```diff
@@ -307,6 +307,25 @@ def import_names(*args, **kwargs):
...
+
+    s = 'from os import path, p'
+    assert 'path' not in import_names(s)
+    assert 'path' in import_names(s, column=len(s) - 3)
+
+    s = 'import path as pp, p'
+    assert 'path' not in import_names(s)
+
+    s = 'from os import path as pp, p'
+    assert 'path' not in import_names(s)
```
davidhalter, line 326: Can you also add a test here for 
```
s = 'from os import chdir as path, p'
assert 'path' in import_names(s)
```

I feel like this is still wrong.

davidhalter (changes requested):

test/test_inference/test_imports.py
WutingjiaX: Yes, it will be wrong. I mentioned it above:
> import alias will also be filtered Generally, there is unlikely to be a situation where an alias is named exactly the same as a specific original name.

If you think it unacceptable，I can parse this syntax and remove it from the filter

davidhalter: Oh sorry, I didn't read that.

> If you think it unacceptable，I can parse this syntax and remove it from the filter

I think it is acceptable, but it is the kind of thing I always fix before merging. So if you don't do it, I have to :)

I think it's also a very easy fix (probably one simple if).

davidhalter: Also know that this piece of software is used by millions of people, so issues like this one will appear sooner or later.

WutingjiaX: Done～

davidhalter:
Thanks a lot for the quick fixes and in general for the PR, good stuff!

PR #2042, opened 2024-12-25
Fix inheritance diagram

### LINKED ISSUE #2041: Broken inheritance diagram in current documentation
While going through the documentation, I noticed that [this section](https://jedi.readthedocs.io/en/latest/docs/development.html#module-jedi.inference.base_value) of the online docs was broken. I was able to reproduce that locally, and found that I was missing the `dot` executable, because I installed `graphviz` with `pip` instead of my package manager it seems. After installing it with my package manager I did recover the graph.

So I'm thinking `dot` wasn't accessible on whatever machine was used to build the docs. Note that this is has been the case since version `0.19.2`, so the stable version of the documentation is also affected by this.

bluthej:
Closes #2041.

I think this should do the trick, as per [this section](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-apt-packages) of the readthedocs documentation.

I wasn't able to actually test this out myself because I couldn't find a way to test this locally, even in the RTD docs.
It looks like they make users rely on [Pull request previews](https://docs.readthedocs.io/en/stable/pull-requests.html#pull-request-previews) in order to test the config. Are those activated for this repo?


davidhalter:
Thanks a lot for looking into this. I have just enabled pull request previews. That's a really good suggestion. Can you try and force push this change, so we can test if this actually works?

bluthej:
Sure thing! I just did and then looked at the docs/readthedocs.org:jedi check, clicked on [Details](https://jedi--2042.org.readthedocs.build/en/2042/) which opened the preview and [the diagram is showing correctly](https://jedi--2042.org.readthedocs.build/en/2042/docs/development.html#module-jedi.inference.base_value) :)

Thanks for the kind words, I appreciate it!
I've been using `jedi` everyday for the past year and a half through `python-lsp-server`, so I'm very grateful that it exists and if I can contribute back even on a very small thing like that I'm very happy :)

davidhalter:
Thanks :)

PR #2051, opened 2025-03-04
perf: optimize transitions and path handling using sets 

allrob23:
## Description  
This PR improves performance by replacing list-based operations with sets in `allowed_transitions` and `safe_paths`.  
Using sets instead of lists reduces lookup time from **O(n) to O(1)** in most cases, leading to better efficiency.  

### Changes:  
- Replaced `+=` with `.update()` in `allowed_transitions` to leverage set operations.  
- Used `set()` for `safe_paths` instead of a list to optimize lookups.  

### Why is this needed?  
- **Faster lookups**: Checking membership in a set is **O(1)** on average, whereas lists are **O(n)**.  
- **Cleaner and more idiomatic Python**: Sets are the right data structure for unique unordered elements.  
- **Improved maintainability**: Ensures consistent use of efficient data structures.  

### Impact  
These changes should have **no functional impact** but will improve performance in cases where transitions or paths are frequently checked.

### DIFF
--- a/jedi/inference/imports.py
+++ b/jedi/inference/imports.py
@@ -480,7 +480,7 @@ def _load_builtin_module(inference_state, import_names=None, sys_path=None):
     if sys_path is None:
         sys_path = inference_state.get_sys_path()
     if not project._load_unsafe_extensions:
-        safe_paths = project._get_base_sys_path(inference_state)
+        safe_paths = set(project._get_base_sys_path(inference_state))
         sys_path = [p for p in sys_path if p in safe_paths]
 
     dotted_name = '.'.join(import_names)


jedi/inference/imports.py
```diff
@@ -480,7 +480,7 @@ def _load_builtin_module(inference_state, import_names=None, sys_path=None):
     if sys_path is None:
         sys_path = inference_state.get_sys_path()
     if not project._load_unsafe_extensions:
-        safe_paths = project._get_base_sys_path(inference_state)
+        safe_paths = set(project._get_base_sys_path(inference_state))
```
PeterJCLaw, line 483: Is this definitely correct/desirable? It seems that this variable is only used on the following line to construct another list. So the set conversion here appears to be pure overhead, while losing the ordering (not sure if the ordering is important or not).

allrob23: In the next line, we have if p in safe_paths inside a for loop. If safe_paths is a list, the time complexity will be O(n * m) due to the linear search in each iteration. However, if safe_paths is a set, membership checks are O(1), reducing the overall complexity to O(n).

PeterJCLaw: Oh, sorry, I misread which way around the loop was; somehow I read it as looping over `safe_paths` rather than over `sys_path`.

PeterJCLaw:
These changes seem fine, though I'm struggling locally to see much speedup. Is there a particular use-case you have which is substantially impacted here, or is this just from code inspection? I think the latter is fine, though I'm wondering in the former case it might be worth a test or something to show/check the impact?

allrob23:
This was flagged by a tool I’m developing, which performs a sort of code inspection to identify potential optimizations. However, it doesn’t report execution times, so I haven’t measured the actual impact.

PeterJCLaw:
Interesting. Something you might want to consider/might not be aware of (though if you're writing a tool maybe you already are) is that for small collections, a linear search is often faster, especially considering the time to create the `set`. I haven't extensively tested this on Python, though testing in other languages I've usually seen that N=4 or N=6 is about the threshold. (I thiiiiink I've also seen that tuples are faster in Python than lists, but obviously that only holds for essentially fixed collections)

Separately, I've fixed the docs issue -- if you wanted to rebase that should give a clean CI run.

davidhalter:
> I haven't extensively tested this on Python, though testing in other languages I've usually seen that N=4 or N=6 is about the threshold.

Ironically the transition case is very often going to be below N=4. And the sys path will probably be something like N=6 to N=20.

> (I thiiiiink I've also seen that tuples are faster in Python than lists, but obviously that only holds for essentially fixed collections)

Interesting, which Python version was it? I was always surprised that my measurements were slower with tuples, which always felt strange. But that was probably in Python 3.3-3.8 land, no idea how it is currently.

----------------------------

About the PR: Please remove the transitions part again, I'm pretty sure the transition case is slower with a set. This is because the set construction is almost certainly slower than 2*n lookups. The sys_path set is probably a small improvement and makes sense, as far as I can tell.

allrob23:
Thanks for the feedback from both of you! I changed the PR to just have `sys_path`

davidhalter:
Thanks!

PR #2054, opened 2025-03-12
limit columns to line_length

mstuehn:
(neo)vim provides a mode "virtualedit=all" where the cursor can be placed behind the last character of the line. Limit the column to line_lenght prevents raising a ValueError

### DIFF
--- a/jedi/api/helpers.py
+++ b/jedi/api/helpers.py
@@ -479,7 +479,7 @@ def validate_line_column(func):
         elif line_string.endswith('\n'):
             line_len -= 1
 
-        column = line_len if column is None else column
+        column = min(line_len, line_len if column is None else column)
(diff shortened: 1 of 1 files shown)


davidhalter:
This is unfortunately not fixing the issue, it's simply circumventing it. If I were to merge this, things like `not` will complete to `NotImplemented` instead of having all global completions. Goto will also act weird and so on. This is unfortunately not that simple. As I said before, this is unfortunately not a simple change.

Plus, this also needs tests.

mstuehn:
Ah, ok. The patch worked ok for me, but i have to admit, that i do not fully understand the implications. 
Sorry for the noise; unfortunately i do not have the capabilities to provide a proper solution.

Issue #2053, opened 2025-03-11
Cannot run `subprocess` with "non-standard" `sys.path`.


zuckerruebe:
Hi there

When using `pylsp` in `spyder` it crashes when setting up the `jedi` `environment` like so:

```powershell
Traceback (most recent call last):
[...]
  File "C:\Users\damian.birchler\Downloads\pytrnsys-gui-v0.10.27+process-gui_7984a6c-pytrnsys_fc95114-2025-03-10T08_09_22+00_00\site-packages\pylsp\workspace.py", line 549, in jedi_script
    environment = self.get_enviroment(environment_path, env_vars=env_vars)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\damian.birchler\Downloads\pytrnsys-gui-v0.10.27+process-gui_7984a6c-pytrnsys_fc95114-2025-03-10T08_09_22+00_00\site-packages\pylsp\workspace.py", line 581, in get_enviroment
    environment = jedi.api.environment.create_environment(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\damian.birchler\Downloads\pytrnsys-gui-v0.10.27+process-gui_7984a6c-pytrnsys_fc95114-2025-03-10T08_09_22+00_00\site-packages\jedi\api\environment.py", line 366, in create_environment
    return Environment(path, env_vars=env_vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\damian.birchler\Downloads\pytrnsys-gui-v0.10.27+process-gui_7984a6c-pytrnsys_fc95114-2025-03-10T08_09_22+00_00\site-packages\jedi\api\environment.py", line 68, in __init__
    self._get_subprocess()
  File "C:\Users\damian.birchler\Downloads\pytrnsys-gui-v0.10.27+process-gui_7984a6c-pytrnsys_fc95114-2025-03-10T08_09_22+00_00\site-packages\jedi\api\environment.py", line 79, in _get_subprocess
    raise InvalidPythonEnvironment(
[...]

PR #2055, opened 2025-03-13
Don't remove `sys.path[0]`.

zuckerruebe:
Fixes https://github.com/davidhalter/jedi/issues/2053.


zuckerruebe:
Hi @davidhalter 

Continuing our discussion in #2053, it seems just deleting the `del sys.path[0]` line passes the tests. Whether you want to go with the safer "only delete first entry if it's the empty string" only you can decide...

davidhalter:
Thanks. I feel like this is fine. Time will tell if there are going to be issues, but it definitely does not look necessary anymore in the general case.

Issue #2064, opened 2025-06-03
Python 3.14 compatibility


frenzymadness:
In Python 3.14, some implementation details of typing have been changed, for example: https://github.com/python/cpython/issues/105499 

This is causing failures of some of the tests:

```
=================================== FAILURES ===================================
________________ test_string_annotation[annotations8-result8-] _________________

annotations = {'return': 'typing.Union[str, int]'}, result = ['int', 'str']
code = ''

    @pytest.mark.parametrize(
        'annotations, result, code', [
            ({}, [], ''),
            (None, [], ''),
            ({'asdf': 'str'}, [], ''),
    
            ({'return': 'str'}, ['str'], ''),
            ({'return': 'None'}, ['NoneType'], ''),
            ({'return': 'str().upper'}, [], ''),
            ({'return': 'foo()'}, [], ''),
            ({'return': 'bar()'}, ['float'], ''),
    
            # typing is available via globals.
            ({'return': 'typing.Union[str, int]'}, ['int', 'str'], ''),
            ({'return': 'typing.Union["str", int]'},
             ['int', 'str'] if sys.version_info >= (3, 9) else ['int'], ''),
            ({'return': 'typing.Union["str", 1]'},
             ['str'] if sys.version_info >= (3, 11) else [], ''),
            ({'return': 'typing.Optional[str]'}, ['NoneType', 'str'], ''),
            ({'return': 'typing.Optional[str, int]'}, [], ''),  # Takes only one arg
            ({'return': 'typing.Any'},
             ['_AnyMeta'] if sys.version_info >= (3, 11) else [], ''),
    
            ({'return': 'typing.Tuple[int, str]'},
             ['Tuple' if sys.version_info[:2] == (3, 6) else 'tuple'], ''),
            ({'return': 'typing.Tuple[int, str]'}, ['int'], 'x()[0]'),
            ({'return': 'typing.Tuple[int, str]'}, ['str'], 'x()[1]'),
            ({'return': 'typing.Tuple[int, str]'}, [], 'x()[2]'),
    
            ({'return': 'typing.List'}, ['list'], 'list'),
            ({'return': 'typing.List[int]'}, ['list'], 'list'),
            ({'return': 'typing.List[int]'}, ['int'], 'x()[0]'),
            ({'return': 'typing.List[int, str]'}, [], 'x()[0]'),
    
            ({'return': 'typing.Iterator[int]'}, [], 'x()[0]'),
[...]

PR #2070, opened 2025-08-24
Add Python 3.14 into CI configuration

frenzymadness:
Related to #2064 

This adds Python 3.14 to CI without adding it to the supported versions yet.


davidhalter:
I'm not going to merge for now (until tests pass), but thanks anyway. I did not think it would be that easy.

I hope I find the time to fix the actual issues, but if somebody else wants to tackle this, please do.

PR #2104, opened 2026-08-02
Fix `TypeError` when `get_line_code` called on literals

clayote:
Calling this method on literals usually takes you to standard library code, tightly wrapping the underlying C. Not very useful most of the time, but a logical thing to ask for.

Jedi should have the ability.

The lack of it caused an error in emacs' anaconda-mode when attempting to inspect the name of a literal value.


davidhalter:
But the question here is: Why does this happen? This might just be a bug in a different place. Catching AttributeErrors like this is almost always a bad idea unless we actually know what's going on and then a comment should probably describe that.

davidhalter:
Also you are changing `name` here, which is an absolute no-go whenever you encounter python code. There should never be side effects from anywhere so random.

Issue #1888, opened 2022-10-19
Completions fail with KeyError in parser_cache if fast_parser is False 


HedgehogCode:
The autocompletion `jedi.Script#complete` fails with because the `parso.cache.parser_cache` is empty when setting `jedi.settings.fast_parser = False`.

Try this script to reproduce the issue:
```python
import jedi

jedi.settings.fast_parser = False

script = jedi.Script(
    code="im", # Just starting a new file. Trying to get 'import' completed
    path="", # Using "" for unsaved files as described in the docstring
)

# Complete at the end of the code
completions = script.complete() 
print(completions)
```

Traceback:
```
Traceback (most recent call last):
  File "/home/benjamin/misc/jedi/./jedi-18-1-bug.py", line 12, in <module>
    completions = script.complete() 
  File "/home/benjamin/misc/jedi/jedi/api/helpers.py", line 487, in wrapper
    return func(self, line, column, *args, **kwargs)
  File "/home/benjamin/misc/jedi/jedi/api/__init__.py", line 214, in complete
    return completion.complete()
  File "/home/benjamin/misc/jedi/jedi/api/completion.py", line 170, in complete
    cached_name, completion_names = self._complete_python(leaf)
  File "/home/benjamin/misc/jedi/jedi/api/completion.py", line 314, in _complete_python
    completion_names += self._complete_global_scope()
  File "/home/benjamin/misc/jedi/jedi/api/completion.py", line 379, in _complete_global_scope
    for filter in filters:
  File "/home/benjamin/misc/jedi/jedi/inference/context.py", line 486, in get_global_filters
    yield from context.get_filters(
  File "/home/benjamin/misc/jedi/jedi/inference/context.py", line 318, in get_filters
    next(filters, None)
  File "/home/benjamin/misc/jedi/jedi/inference/value/module.py", line 63, in get_filters
    ParserTreeFilter(
  File "/home/benjamin/misc/jedi/jedi/inference/filters.py", line 138, in __init__
    super().__init__(parent_context, node_context)
  File "/home/benjamin/misc/jedi/jedi/inference/filters.py", line 100, in __init__
    self._parso_cache_node = get_parso_cache_node(
  File "/home/benjamin/misc/jedi/jedi/parser_utils.py", line 287, in get_parso_cache_node
    return parser_cache[grammar._hashed][path]
KeyError: '715ad56c5f4f8395092c58b6b6f2deb4f906f81380929a836bd86ab253634875'
```

[...]

PR #2105, opened 2026-08-17
Fix: get_parso_cache_node() in jedi/parser_utils.py performs an unguarded...

M001N:
## Summary
Wrapped the get_parso_cache_node(...) call in jedi/inference/filters.py's _AbstractUsedNamesFilter.__init__ in a try/except KeyError, falling back to self._parso_cache_node = None, mirroring the existing fallback used for the path is None branch.

## Problem
davidhalter/jedi issue reference: davidhalter/jedi#1888

## Root Cause
get_parso_cache_node() in jedi/parser_utils.py performs an unguarded dict lookup into parso's internal parser_cache. That cache is only reliably populated for a path when diff_cache is enabled during parsing, but jedi passes diff_cache=settings.fast_parser (jedi/api/__init__.py, jedi/inference/imports.py). With fast_parser=False, the cache entry for an unsaved file's path may never exist, so the lookup raises KeyError. The same __init__ in jedi/inference/filters.py already treats a missing/unavailable cache node as an expected, handled case one branch above (path is None -> self._parso_cache_node = None), confirming the rest of the class tolerates a None cache node.

## Testing
PASS - all 3 tests passed (2 existing test_cache.py tests + new regression test); reproduced the KeyError crash pre-fix and confirmed it no longer occurs post-fix.

## Related Issue
davidhalter/jedi#1888


davidhalter:
**For the love of god, please disclose that this was done with AI!!!**


Hi, I'm sorry to say this, but I decided to not work at all with AI generated pull requests/content. There are multiple reasons for this:

1. I feel like at this point it's pretty subpar to what a good engineer can create. I have received many AI generated pull requests and they all have massive problems compared to what I received from actual people. People sometimes can't fix the problems they have without AI, but I regard this as a feature: If you don't understand the problem, I don't have to review your code.
2. I love people. And I would like to interact with all of you; learn from you; teach you. But if you are simply a bridge to an LLM, I'm just wasting my time. I cannot build any form of relationship.
3. LLMs are the antithesis to what I like about programming. I like the struggle of programming. Of writing code. I love how things come together after a good coding session. This is all missing. And while some form of AI might make that skill useless in the future, it's not useless now and I would really recommend anyone to avoid LLMs for generating complex code.
4. Even for written English I generally prefer non-LLM text, because it shows something about the person. I understand that some people are incredibly non-fluent, but that's fine. Just do your best and try to learn English. Otherwise we won't have the ability to talk if we ever meet, which would be a pitty.
5. LLMs are extremely good at generating code/text that looks reasonable. I use them a lot for brainstorming, they are incredible at putting out ideas. They are bad at facts. They are really bad at thinking. The positive thing here is that you have a brain and you can use it to think. Think hard first.
6. LLMs tend to generate lots of code that is then hard to review. Try to contribute small changes first.

Thank you anyway for trying to contribute to Open Source. I would really appreciate if you avoid the usage of LLMs in my projects. It is obviously fine to use LLMs as a search/brainstorming tool with my projects, just don't use it to interact with me.

PS: As a side note I want to mention that I'm probably not the only maintainer that has a difficult relationship with LLM content. So you probably should ask first before you use LLMs for other repositories.

